PR #39, opened 2020-08-11
Add check-spelling/check-spelling

jsoref:
Normally I offer fixes first and the action second, but my hand isn't up to that.

This adds the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

You don't want to add it right now, as it will scream:
https://github.com/jsoref/argocd-image-updater/commit/24460cfea548cd380296ee4413c620abcbc14996#commitcomment-41382721
but once you've fixed those items, you can merge this and it should help catch future things.

I'm happy to provide support for this tool (I need to make a 0.17-alpha release, I'm just collecting a bit more info and procrastinating on a feature).

You can use `pull_request:` instead of `push:` if you want the action to run in your repo in response to PRs.

I default to `push:` because it means I can iterate on a branch w/o making a PR.
If someone forks and doesn't enable actions (and I don't understand how this works), then they can bypass `push:`, also note that you can't fail a PR if you use `push:`, but I don't really recommend blocking on spelling, that's a bit of a high bar.


jsoref:
(You can slightly customize the message it provides, and over time I will improve it... the code it's suggesting assumes that the words are correct-ish but not in the dictionary it's using. Here, all the `words` it has identified are incorrectly spelled, and thus its advice is wrong...)

jannfis:
Thanks! Let's give it a try :)

PR #71, opened 2020-08-24
Only check semver & constraint if sorting by semver

mgoodness:
As discussed in Slack, image-updater checks that tags conform to SemVer even when using a regex pre-filter. This skips that check unless the sort mode is set to `VersionSortSemVer`.

Not sure if this is the best solution, but I am now able to deploy the most recent image (by timestamp) that matches my regexp.

Signed-off-by: Michael Goodness <michael.goodness@mlb.com>


jannfis (approved):
LGTM! Thank you a bunch!

Issue #138, opened 2021-01-08
Allow reuse of Argo CD repo credentials


mgoodness:
Per the [latest docs](https://github.com/argoproj-labs/argocd-image-updater/commit/4a7c10bfba8b3de2ace87c8c1798d87e82dd8afd#diff-98ff25374640ade0b3c23d02bbb3775e3fb759761c99286fa17e69d27f33695bR166) argocd-image-updater won't reuse Argo CD repo credentials for Git write-back. Is there a technical reason for the limitation, a security concern, or other? It would be great if users who understand the security implications could configure reuse and avoid adding another annotation to each Application definition.

jannfis:
The main reason right now is, that Argo CD API does not return credentials with a repository request, for security reasons. So, there's no way for Image Updater to automatically fetch the credentials. Also, the default SA of image updater has no access to read K8s resources in the `argocd` namespace by default.

I guess one of the upcoming steps of integration is to move the image updater into the `argocd` namespace, and also allow it to read the Argo CD resources in K8s. That way, we can also read and re-use credentials configured in Argo CD, directly from the K8s resources, and optionally have them override this with an annotation (e.g. if the user configured in Argo CD is read only).

jannfis:
I also think that in #35, @hawksight suggested to have a globally configurable secret holding Git credentials, so that not each Application would require a dedicated annotation, but repos would need a shared deploy key configured.

I'm pretty sure we won't change Argo CD API to also return secrets on repository data retrieval, and making argocd-image-updater deployment to be mandatory in argocd namespace is also more on the long term roadmap, would that be a trade-off you could live with?

jannfis:
Or maybe, it'd be possible to create a `Role` in the `argocd` namespace to allow reading of `argocd-cm` and the Git secrets, and bind it to `argocd-image-updater` SA. This would require some manual configuration as well.

Then, `argocd-image-updater` could use a rule like the following to figure out which secret to use:

1. If neither global secret nor application specific credential is given, read Git credentials from secret configured in Argo CD
2. If application specific credentials are given as annotation, use that
3. Fall back to globally configured secrets

I think this could work out.

mgoodness:
Using RBAC to grant image-updater `read` access to `argocd-cm` is exactly what I was thinking. I also like the idea of consolidating annotations.

alexmt:
I also like both proposals. Working on PR

alexmt:
[...]

PR #141, opened 2021-01-22
feat: Allow reuse of Argo CD repo credentials

alexmt:

Closes https://github.com/argoproj-labs/argocd-image-updater/issues/138

PR implements proposal described in https://github.com/argoproj-labs/argocd-image-updater/issues/138 . The default credentials source is "repocreds" since it requires no configuration.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -4,3 +4,5 @@ mypass
 otherapp
 wildcard
 wildcards
+credref
+repocreds
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -253,7 +253,7 @@ argocd-image-updater test nginx --allow-tags '^1.19.\d+(\-.*)*$' --update-strate
 			var kubeClient *kube.KubernetesClient
 			var err error
 			if kubeConfig != "" {
-				kubeClient, err = getKubeConfig(ctx, kubeConfig)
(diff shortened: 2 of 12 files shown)


jannfis (approved):
LGTM - awesome, thank you @alexmt !

diranged:
@jannfis @alexmt ,
  How would this play with https://github.com/argoproj/argo-cd/pull/5355?

alexmt:
Good point @diranged . We already had a discussion about re-using argocd repo server  to communicate with Git. The downside of that approach is that repo server would get API which is not used by Argo CD itself. 

Image updater already imports Argo CD as a library. So may be we just contribute new Git client methods back to Argo CD? In this case we will get github apps support and future features transparently. 

What do you think @jannfis ?

diranged:
I like currently how the image updater is just working along side argo and using native kubernetes calls for everything. If the argocd application controller is going to be responsible for being aware of the ".argocd-<app>.yaml" files anyways, perhaps thats where the responsibility should lie to push changes back into a repository based on the status of an Application resource?

The benefit is that _any_ Application customization or override can then be passed back into the Git repository as-code if the operators choose to do that... which would actually be a great win for our development environment where we want to enable developers to make changes to the Application objects directly, but are also afraid of configuration drift from what's in-code vs what is in-cluster.

Issue #140, opened 2021-01-21
Support managing Application CRD using K8S client


alexmt:
**Is your feature request related to a problem? Please describe.**

Currently, the image updater retrieves a list of applications using Argo CD API. That requires the creation of an Argo CD token and connection options (such as grpc-web, insecure etc).

We might simplify the initial configuration if the updater is running in the same namespace as Argo CD. In this case list of applications can be extracted using Kubernetes client.

**Describe the solution you'd like**

Add `--applications-source` flag that takes two parameters: `argocd`, `namespace`. If `namespace` is specified then the updater should read applications from the current namespace.

**Additional context**
I think this might improve adoption significantly since this way we support no configuration installation. The user just needs to run `kubectl apply -n argocd -f ...` (or add one more remote base to kustomization file) and the updater should work with public image registries out of the box.

It is still good to support connecting using argocd API. Docker registry might be not available from Argo CD namespace or updater cannot be installed into the same namespace for security reasons.

jannfis:
I like this idea. 

Ideally, we should also make the target namespace also configurable, with a sane default.

 So if `--applications-source` is set to namespace, we should get the target namespace to use via another flag, e.g. `--argocd-namespace` (similar to `--argocd-server-addr`), which defaults to `argocd`.

PR #144, opened 2021-01-22
feat: support managing Application CRD using K8S client (#140)

alexmt:

Closes https://github.com/argoproj-labs/argocd-image-updater/issues/140 

PR implements proposal from  https://github.com/argoproj-labs/argocd-image-updater/issues/140:

* By default image updater tries to manage Argo CD applications in the same namespace where it is running using K8S API 
* Namespace can be changed using `--argocd-namespace` flag
* Updater can be switch back to Argo CD API using `--applications-api argocd` flag (or `APPLICATIONS_API` env variable).

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -31,20 +31,24 @@ const defaultArgoCDServerAddr = "argocd-server.argocd"
 // Default path to registry configuration
 const defaultRegistriesConfPath = "/app/config/registries.conf"
 
+const applicationsAPIKindK8S = "kubernetes"
+const applicationsAPIKindArgoCD = "argocd"
+
 // ImageUpdaterConfig contains global configuration and required runtime data
 type ImageUpdaterConfig struct {
-	ClientOpts      argocd.ClientOptions
-	ArgocdNamespace string
-	DryRun          bool
-	CheckInterval   time.Duration
-	ArgoClient      argocd.ArgoCD
-	LogLevel        string
-	KubeClient      *kube.KubernetesClient
-	MaxConcurrency  int
-	HealthPort      int
-	MetricsPort     int
-	RegistriesConf  string
-	AppNamePatterns []string
+	ApplicationsAPIKind string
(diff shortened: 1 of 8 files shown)


cmd/main.go
```diff
@@ -71,7 +75,16 @@ func warmupImageCache(cfg *ImageUpdaterConfig) error {
 // Main loop for argocd-image-controller
 func runImageUpdater(cfg *ImageUpdaterConfig, warmUp bool) (argocd.ImageUpdaterResult, error) {
 	result := argocd.ImageUpdaterResult{}
-	argoClient, err := argocd.NewClient(&cfg.ClientOpts)
+	var err error
+	var argoClient argocd.ArgoCD
+	switch cfg.ApplicationsAPIKind {
+	case applicationsAPIKindK8S:
+		argoClient = argocd.NewK8SClient(cfg.KubeClient)
```
jannfis, line 82: Will this never return an error?

pkg/argocd/argocd.go
```diff
@@ -9,14 +9,60 @@ import (
...
+			return nil, err
+		}
+		return &updatedApp.Spec, nil
+	}
+
+}
+
+// NewAPIClient creates a new API client for ArgoCD and connects to the ArgoCD
+// API server.
+func NewK8SClient(kubeClient *kube.KubernetesClient) ArgoCD {
```
jannfis, line 62: To be consistent with `NewAPIClient`, I think we should have the return value specified as `(ArgoCD, error)`, even if we currently just return `nil` here.

jannfis (changes requested):
Thanks @alexmt! LGTM, with just one nitpicker comment.

pkg/argocd/argocd.go
alexmt: Agree, we might need to return error in a future. Applied your suggestion.

jannfis (approved):
Thanks! LGTM!

PR #148, opened 2021-01-29
fix: install missing git binary

alexmt:


jannfis (approved):
LGTM, I'll shut up the spell checker with the next PR.

I think as a follow up, we'll also need to configure Git user and e-mail before commit/push, and possibly have this configurable.

PR #167, opened 2021-03-21
#164 Send events on image change

### LINKED ISSUE #164: Create events for changes
We need to notify another systems when a version has changed and which is the new tag being used. 

I'd like to be able to monitor a kubernetes event containing at least:
- resource type ("deployment")
- resource name ("foo")
- tag used ("1.2.3")
- namespace ("foo")
- optionally, previous tag
- optionally, image name and/or repository.

This way we could add another piece to the pipeline to notify external services.

I've considered:
- to use [ArgoCD-notifications](https://github.com/argoproj-labs/argocd-notifications), but it only monitors Applications, so it won't know the tag being used.
- to use [kubernetes-event-exporter](https://github.com/opsgenie/kubernetes-event-exporter), configuring a label with the tag, but after running argocd-image-updater the tag disapears, and events might be duplicated due to replication.

**Additional context**
By creating an event, other systems can take it to perform other different actions: notifications (email, slack, pagerduty, ...), maintain a dashboard, monitoring (grafana with tags showing the tag currently used), traceability (logs), ...

magmax:
Sends an event on any image change (fixes #164)

I thought in re-using some argo-cd code, but finally I decided to copy and adapt it. It is the code related to send the event.

The namespace used is the one where argocd-image-updater is running in order to reduce the permissions required (a Role instead a ClusterRole).

I had to add some argo-cd dependencies to the kubernetes package, what is... ugly. Maybe you know a better way to do it.

Finally, understand this is my first golang contribution... I'm open to changes XD

NOTE: I didn't tested it under real fire. I have no docker repository accessible from my laptop.

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -32,13 +32,14 @@ type ImageUpdaterResult struct {
 }
 
 type UpdateConfiguration struct {
-	NewRegFN       registry.NewRegistryClient
-	ArgoClient     ArgoCD
-	KubeClient     *kube.KubernetesClient
-	UpdateApp      *ApplicationImages
-	DryRun         bool
-	GitCommitUser  string
-	GitCommitEmail string
+	NewRegFN          registry.NewRegistryClient
+	ArgoClient        ArgoCD
+	KubeClient        *kube.KubernetesClient
+	UpdateApp         *ApplicationImages
+	DryRun            bool
+	GitCommitUser     string
+	GitCommitEmail    string
+	DisableKubeEvents bool
 }
 
 type GitCredsSource func(app *v1alpha1.Application) (git.Creds, error)
@@ -206,7 +207,7 @@ func UpdateApplication(updateConf *UpdateConfiguration) ImageUpdaterResult {
 				message := fmt.Sprintf("Successfully updated image '%s' to '%s', but pending spec update (dry run=%v)", updateableImage.GetFullNameWithTag(), updateableImage.WithTag(latest).GetFullNameWithTag(), updateConf.DryRun)
 				imgCtx.Infof(message)
 				result.NumImagesUpdated += 1
-				if !updateConf.DryRun {
+				if !updateConf.DryRun && updateConf.KubeClient != nil && !updateConf.DisableKubeEvents {
 					annotations := map[string]string{
 						"image-old": updateableImage.GetFullNameWithTag(),
 						"image-new": updateableImage.WithTag(latest).GetFullNameWithTag(),
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -82,9 +82,9 @@ type helmOverride struct {
 }
 
 type change struct {
-    image     *image.ContainerImage
-	oldTag    *tag.ImageTag
-	newTag    *tag.ImageTag
+	image  *image.ContainerImage
+	oldTag *tag.ImageTag
+	newTag *tag.ImageTag
 }
 
 // UpdateApplication update all images of a single application. Will run in a goroutine.
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -203,8 +203,16 @@ func UpdateApplication(updateConf *UpdateConfiguration) ImageUpdaterResult {
 				result.NumErrors += 1
 				continue
 			} else {
-				imgCtx.Infof("Successfully updated image '%s' to '%s', but pending spec update (dry run=%v)", updateableImage.GetFullNameWithTag(), updateableImage.WithTag(latest).GetFullNameWithTag(), updateConf.DryRun)
+                message := fmt.Sprintf("Successfully updated image '%s' to '%s', but pending spec update (dry run=%v)", updateableImage.GetFullNameWithTag(), updateableImage.WithTag(latest).GetFullNameWithTag(), updateConf.DryRun)
+				imgCtx.Infof(message)
 				result.NumImagesUpdated += 1
+                if ! updateConf.DryRun {
+                    annotations := map[string]string{
+                            "image-old": updateableImage.GetFullNameWithTag(),
+                            "image-new": updateableImage.WithTag(latest).GetFullNameWithTag(),
+                        }
+                    updateConf.KubeClient.CreateApplicationEvent(&updateConf.UpdateApp.Application, "ImageUpdated", message, annotations)
+                }
 			}
 		} else {
 			// We need to explicitly set the up-to-date images in the spec too, so
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -203,16 +203,19 @@ func UpdateApplication(updateConf *UpdateConfiguration) ImageUpdaterResult {
 				result.NumErrors += 1
 				continue
 			} else {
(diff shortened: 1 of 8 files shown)


pkg/argocd/update.go
```diff
@@ -210,8 +210,19 @@ func UpdateApplication(updateConf *UpdateConfiguration) ImageUpdaterResult {
...
+				if !updateConf.DryRun {
+					annotations := map[string]string{
+						"image-old": updateableImage.GetFullNameWithTag(),
+						"image-new": updateableImage.WithTag(latest).GetFullNameWithTag(),
+					}
+					_, err = updateConf.KubeClient.CreateApplicationEvent(&updateConf.UpdateApp.Application, "ImageUpdated", message, annotations)
+					if err != nil {
+						imgCtx.Warnf("Event could not be sent: %v", err)
+					}
+				}
```
jannfis, line 225: I think this should go further down at https://github.com/argoproj-labs/argocd-image-updater/blob/af590d33b782c339f4bdeb06a7fb79a313703e19/pkg/argocd/update.go#L245, after changes have been committed successfully to Argo CD and/or Git.

Also, because it is possible that we do not have a Kubernetes client, we should explicitly test for `updateConf.KubeClient != nil` before trying to use it.

jannfis (changes requested):
Hey, awesome @magmax! Pretty great for your first lines of Go, actually! I only have a few comments (see inline) and also the following:

* We need to give the ServiceAccount for the Image Updater appropriate Kubernetes RBAC permissions to create Events. The `Role` is part of the manifests and can be found [here](https://github.com/argoproj-labs/argocd-image-updater/blob/master/manifests/base/rbac/argocd-image-updater-role.yaml). Can you please provide proper authorization there?
* I feel that emitting events should be enabled by default, but optionally can be disabled. It would be great to have a command line switch to disable it (e.g. `--disable-kube-events`). The main command can be found [here](https://github.com/argoproj-labs/argocd-image-updater/blob/af590d33b782c339f4bdeb06a7fb79a313703e19/cmd/main.go#L379) and the flags are defined [here](https://github.com/argoproj-labs/argocd-image-updater/blob/af590d33b782c339f4bdeb06a7fb79a313703e19/cmd/main.go#L524). We provide default values to most flags that are override-able using environment variables (e.g. using `env.GetBoolVal()`), and then map those to keys in the `argocd-image-updater-config` ConfigMap, as shown [here](https://github.com/argoproj-labs/argocd-image-updater/blob/af590d33b782c339f4bdeb06a7fb79a313703e19/manifests/base/deployment/argocd-image-updater-deployment.yaml#L24). I think this one should be settable by `IMAGE_UPDATER_KUBE_EVENTS` with a ConfigMap key of `kube.events`.

pkg/argocd/update.go
magmax: Hmmm.... you are right. But I'd need to store the list of updated images in order to use that information in the message. Correct?

jannfis: Yes, we'd probably need to store it in some kind of temporary data structure to access it later on.

Also, sorry, I forgot to mention that you'll need to run `make manifests` (requires `kustomize` installed) after changing the role in the kustomize source manifests and also push those changes to Git, because we'll provide a pre-rendered `install.yaml` and the CI checks whether it differs from the manifests generated during CI run.

magmax:
I did the change.
My brains's python section says that I should have created an object for that.
The C section says it won't work due to memory weaks and misunderstanding.
But the Golang section says that everything is fine XD

I didn't tested it under real fire because I have no repository accessible from this computer and it would take me a while to do it.

pkg/argocd/update.go
```diff
@@ -243,6 +253,19 @@ func UpdateApplication(updateConf *UpdateConfiguration) ImageUpdaterResult {
 				result.NumImagesUpdated = 0
 			} else {
 				logCtx.Infof("Successfully updated the live application spec")
+				result.NumImagesUpdated += 1
+				if !updateConf.DisableKubeEvents && updateConf.KubeClient != nil {
+					annotations := map[string]string{}
+					for _, c := range changeList {
+						annotations[fmt.Sprintf("%s/oldtag", c.imageName)] = c.oldTag.TagName
+						annotations[fmt.Sprintf("%s/newtag", c.imageName)] = c.newTag.TagName
```
jannfis, line 261: On a second thought, I'm not sure whether this will work reliably. The name parts of annotations are rather strict, and must follow a rule:

```
regex used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') with an optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName')
```

Also, the name must be max 63 characters long. I think we must either make sure that the image name is properly normalized, look for another name for the annotation (and put the image name in its value) or think about a completely another solution (e.g. putting the details into the event's `Message`).

Is there a specific reason that you use an annotation for storing the updated images?

magmax: Not really. I want just to "pass parameters" between the argocd-image-updater and a listener through events.

magmax: New approach. Now annotations have less randomness and a fixed size around 40 characters

jannfis: Awesome! I tested it successfully, the event now looks like:

```
Name:             guestbook-multi.166fdad08646fd49
Namespace:        argocd
Labels:           <none>
Annotations:      argocd-image-updater.image-0/image-name: heptio-images/ks-guestbook-demo
                  argocd-image-updater.image-0/new-tag: 0.2
                  argocd-image-updater.image-0/old-tag: 0.1
                  argocd-image-updater.image-1/image-name: coredns/coredns
                  argocd-image-updater.image-1/new-tag: 1.8.3
                  argocd-image-updater.image-1/old-tag: 1.8.0
API Version:      v1               
Count:            1   
...
Involved Object:                
  API Version:       argoproj.io/v1alpha1
  Kind:              Application
  Name:              guestbook-multi                                                                     
  Namespace:         argocd                                                                                                                                                                                        
...
Message:             Successfully updated application 'guestbook-multi'
...
Source:
  Component:  ArgocdImageUpdater
Type:         Normal
```

I'm not sure whether it's intended to use the image name without the registry prefix (the Image Updater's API is possibly not good documented in that regard), but I'd suggest to change

```go
changeList = append(changeList, change{updateableImage.ImageName, updateableImage.ImageTag, containerImageNew.ImageTag})
```
to
```go
changeList = append(changeList, change{updateableImage.GetFullNameWithoutTag(), updateableImage.ImageTag, containerImageNew.ImageTag})
```

at the place where you store the image information to be propagated to the event. 

And I also guess this would be the last change required before merge :tada:

magmax: I'm thinking in using annotations to compose a message on slack. So allowing the user to decide might  be a really cool idea :)

So, I finally decided to add both.

jannfis: I'm fine with that solution as well :+1:

jannfis (approved):
LGTM!

Thanks for this contribution @magmax - much appreciated!

Also, you said it's your first steps into Golang, but I have to say great job on the code and thanks for constructively working through the review with me!

magmax:
Thank you very much for your comments and support.

axot:
If `argocd` is running in a different namespace as `involvedObject.namespace`, the event can't be sent.

> "Event could not be sent: Event \"xxxx\" is invalid: involvedObject.namespace: Invalid value: \"yyyy\": does not match event.namespace" application=zzzz

Issue #187, opened 2021-04-09
feature request: compatibility with knative (or other scale-to-zero deployment models)


NickCao:
**Is your feature request related to a problem? Please describe.**
When trying to update images referenced in knative services, I found lines complaining "Image foo seems not to be live in this application, skipping" in debug log. Only when the given service happens to be active when the update kicks in, it will be considered live. And I found [`GetImagesFromApplication`](https://github.com/argoproj-labs/argocd-image-updater/blob/f1cc8e4542b0531f66fc5546f7467833b1d08ebe/pkg/argocd/argocd.go#L411-L421) to be the primary cause, as application.status.summary.images only contains images that are currently running.

**Describe the solution you'd like**
Add an annotation `argocd-image-updater.argoproj.io/<image_alias>.no-check`, so that the specified image is always updated whether it's live or not.

**Describe alternatives you've considered**
Or we can rework the `GetImagesFromApplication` logic so that it relies on spec rather than status, but that would require too much effort.

**Additional context**
A quick demo of the situation.
![Screenshot from 2021-04-09 11-44-45](https://user-images.githubusercontent.com/15247171/114125414-00bc9580-9929-11eb-9329-023521cc272b.png)

jannfis:
Hi @NickCao, this is indeed an interesting scenario, thank you for bringing it up.

I think an additional annotation, such as you proposed, would be possible. I think it should not be `argocd-image-updater.argoproj.io/<image_alias>.no-check` however, but maybe rather something along `argocd-image-updater.argoproj.io/<image_alias>.force-update`.

As you correctly pointed out, the second option will be difficult to implement on the Image Updater side, because we're piggy-backing on Argo CD here. One of the major goals for Image Updater is to have only very limited permissions on the cluster, and perform all more privileged operations as well as information gathering through Argo CD. So basically this would require a change to Argo CD, to list the images found in the *desired* manifests instead of in the *live* manifests. But certainly, might be something to consider while going forward.

iamnoah:
I need this feature so I'm working on it now 😀

NickCao:
Tested #189 personally for several days, and it worked like a charm, guess we can safely close this now.

Issue #181, opened 2021-03-31
Kustomize aliases without newTag defined result in a nil pointer dereference and application crash


amalagaura:
**Describe the bug**
We wish to have a generic kustomization.yaml without an image tag. We wish to use the argocd image updater to set the image tags. We use a generic image name called `image`. We first tried without the image alias in kustomization.yaml. In that case the image updater fails to update or find the image alias. Apparently Kustomize needs to do a first pass to change the image alias to the final image. So then we add the image `newName`. But without a `newTag` the application crashes. 

**To Reproduce**
A kustomization.yaml with the contents:

```
images:
- name: image
   newName: some-registry.com/some-image
```

with Argo Application with annotations:

```
argocd-image-updater.argoproj.io/image-list: image=some-registry.com/some-image
argocd-image-updater.argoproj.io/image.kustomize.image-name: some-registry.com/some-image
argocd-image-updater.argoproj.io/image.update-strategy: latest
```

This creates a
```
 panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x196e453]
goroutine 22 [running]:
github.com/argoproj-labs/argocd-image-updater/pkg/tag.(*ImageTag).IsDigest(...)
        /src/argocd-image-updater/pkg/tag/tag.go:96
github.com/argoproj-labs/argocd-image-updater/pkg/tag.(*ImageTag).Equals(...)
        /src/argocd-image-updater/pkg/tag/tag.go:102
github.com/argoproj-labs/argocd-image-updater/pkg/argocd.UpdateApplication(0xc000c73b78, 0xc000c45310, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /src/argocd-image-updater/pkg/argocd/update.go:252 +0xc43
main.runImageUpdater.func1(0xc000193e50, 0xc0001a57a0, 0xc000c35ff0, 0xc000c45310, 0xc000050ae0, 0x1, 0xc000c35ff4, 0xc00071fd40, 0x14, 0xc0007253b0, ...)
        /src/argocd-image-updater/cmd/main.go:160 +0x23c
created by main.runImageUpdater
        /src/argocd-image-updater/cmd/main.go:147 +0xa3d
```
 
**Expected behavior**
Image updater should update the image to latest

**Additional context**
[...]

PR #189, opened 2021-04-11
allow image updates to be forced if container not present #181

iamnoah:
Supports #187. I've tested against my deployment of a PodTemplate it works wonderfully :)

### DIFF
--- a/docs/configuration/images.md
+++ b/docs/configuration/images.md
@@ -61,6 +61,21 @@ of the [Semver library](https://github.com/Masterminds/semver) we're using.
     [filtering tags](#filtering-tags)
     below.
 
+
+### Forcing Image Updates
+
+By default, Image Updater will only update an image that is actually used in your Application
+(i.e., is it exported in the Status of your ArgoCD Application.)
+
+To support custom resources and things like PodTemplates that don't actually create a container, 
+you may force an update:
+
+```yaml
+argocd-image-updater.argoproj.io/image-list: myalias=some/image
+argocd-image-updater.argoproj.io/myalias.force-update: "true"
+```
+
+
 ## Assigning aliases to images
 
 It's possible (and sometimes necessary) to assign an alias name to any given
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -172,14 +172,10 @@ func FilterApplicationsForUpdate(apps []v1alpha1.Application, patterns []string)
 			continue
 		} else {
 			log.Tracef("processing app '%s' of type '%v'", app.GetName(), app.Status.SourceType)
-			imageList := make(image.ContainerImageList, 0)
-			for _, imageName := range parseImageList(updateImage) {
-				allowed := image.NewFromIdentifier(strings.TrimSpace(imageName))
-				imageList = append(imageList, allowed)
-			}
+			imageList := parseImageList(updateImage)
 			appImages := ApplicationImages{}
 			appImages.Application = app
-			appImages.Images = imageList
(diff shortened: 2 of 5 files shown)


pkg/argocd/argocd.go
```diff
@@ -187,6 +187,10 @@ func FilterApplicationsForUpdate(apps []v1alpha1.Application, patterns []string)
 	return appsForUpdate, nil
 }
 
+func parseImageList(updateImage string) []string {
+	return strings.Split(updateImage, ",")
+}
```
jannfis, line 192: I like that this becomes externalized in its own method. Maybe it would make sense to also pull in the white-space trimming into this method, what do you think? Also, I think it could be worth to spend this method its own unit test :)

pkg/argocd/argocd.go
```diff
@@ -417,6 +421,18 @@ func GetImagesFromApplication(app *v1alpha1.Application) image.ContainerImageLis
 		images = append(images, image)
 	}
 
+	annotations := app.Annotations
+	if updateImage, ok := annotations[common.ImageUpdaterAnnotation]; ok {
```
jannfis, line 425: Could you please add some explaining commentary to this block? It's easy to understand in the context of this pull request, but I think it might be worth to add some inline documentation explaining why we're doing this so people without the context will understand later on.

pkg/argocd/argocd_test.go
```diff
@@ -56,6 +56,26 @@ func Test_GetImagesFromApplication(t *testing.T) {
 		imageList := GetImagesFromApplication(application)
 		assert.Empty(t, imageList)
 	})
+
+	t.Run("Get list of images from application that has force-update", func(t *testing.T) {
```
jannfis, line 60: Tests are always appreciated, thank you! :+1:

docs/configuration/images.md
```diff
@@ -61,6 +61,21 @@ of the [Semver library](https://github.com/Masterminds/semver) we're using.
     [filtering tags](#filtering-tags)
     below.
 
+
+### Forcing Image Updates
```
jannfis, line 65: +1 for documentation update :)

jannfis (commented):
Thank you @iamnoah for taking the time and effort to submit this new feature!

Your PR looks awesome, and I have no formal request for change. Just some items up for discussion - please see below.

pkg/argocd/argocd.go
iamnoah: Once I pulled in the white space, I noticed the next step was to convert to a ContainerImage so I pulled that in as well. I just focused tests on the string parsing since NewFromIdentifier seems well covered.

pkg/argocd/argocd_test.go
```diff
@@ -702,3 +722,10 @@ func TestKubernetesClient_UpdateSpec_Conflict(t *testing.T) {
 
 	assert.Equal(t, "https://github.com/argoproj/argocd-example-apps", spec.Source.RepoURL)
 }
+
+func Test_parseImageList(t *testing.T) {
+	assert.Equal(t, []string{"foo", "bar"}, parseImageList(" foo, bar ").Originals())
+	// should whitespace inside the spec be preserved?
```
jannfis, line 728: I think whitespace stripping should be done here, but let's do that in another PR: https://github.com/argoproj-labs/argocd-image-updater/blob/f1cc8e4542b0531f66fc5546f7467833b1d08ebe/pkg/image/image.go#L133

jannfis (approved):
LGTM, thank you @iamnoah !

PR #192, opened 2021-04-14
fix issue with force-update and semver

iamnoah:
I realize the test might be a little strange, but I assumed my issue had something to do with customize being used to change the image name so I started there and finally narrowed it down to the version constraint being used as a version.


pkg/argocd/argocd.go
```diff
@@ -429,6 +431,16 @@ func GetImagesFromApplication(app *v1alpha1.Application) image.ContainerImageLis
...
+					// code consuming this list assumes the tag is valid, but for a force update, we may only have a constraint which isn't a valid tag
+					// if using semver, try to parse the tag and remove it if it fails
+					if strategy, set := annotations[fmt.Sprintf(common.UpdateStrategyAnnotation, img.ImageAlias)]; (!set ||
+						image.ParseUpdateStrategy(strategy) == image.VersionSortSemVer) && img.ImageTag != nil {
+						_, err := semver.NewVersion(img.ImageTag.TagName)
+						if err != nil {
+							log.Debugf("Dropping tag from force-update of image %s because %v", img, err)
+							img.ImageTag = &tag.ImageTag{}
+						}
+					}
```
jannfis, line 443: I believe we can simply set the tag to `nil` from what is specified in `image-list`, without having to perform checks for the update strategy or whether the tag can be correctly parsed as a semver.

If an image tag is `nil` in one of the images we check for, we explicitly create an empty tag [here](https://github.com/argoproj-labs/argocd-image-updater/blob/93d76f91a8242421edf6300cabf86c9dac1ac028/pkg/argocd/update.go#L160) and handle that case appropriately [here](https://github.com/argoproj-labs/argocd-image-updater/blob/93d76f91a8242421edf6300cabf86c9dac1ac028/pkg/image/version.go#L86)

So I think we can safely remove all the logic here, and just set `img.ImageTag = nil` here.

jannfis (commented):
Hey @iamnoah - thanks for this bug fix, and also for including a unit test for it!

I think we can simplify the fix, please see below. WDYT?

jannfis:
Regarding the failing golangci-lint check, it will check for properly formatted import blocks. The logic it follows is to check

* whether the imports are correctly grouped (groups are defined by an empty line within the imports)
* whether each group is sorted alphabetically

Import groups are Golang standard library, own package, and external packages, so basically:

```
import (
   // standard lib
   "fmt"
   "time"

  // own package imports
  "github.com/argoproj-labs/argocd-image-updater/pkg/image

  // external packages
  "github.com/Masterminds/semver"
)

pkg/argocd/argocd.go
iamnoah: Makes sense! I'll update shortly

jannfis (approved):
LGTM, thank you @iamnoah !

PR #194, opened 2021-04-14
fix : Images not updated if registry or repository is different with same version

iamnoah:
Given this config:
```
imageList: foo=gcr.io/bar:>=1.0.0
foo.kustomize.image-name=bar
```

If my application is running `bar:1.0.0` I'd expect an update to `gcr.io/bar:1.0.0`, however that was not happening because only tag was considered in the difference. To fix that, I had to add an extra "KustomizeImage" to also be considered for matching purposes, and add registry and name to the update check if this kustomize image is present.

I've added an additional test to show that once updated there should be no more updates (no endless loop of updates, etc.)

This is builds off of #192

### DIFF
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -189,7 +189,7 @@ func parseImageList(annotations map[string]string) *image.ContainerImageList {
 		splits := strings.Split(updateImage, ",")
 		for _, s := range splits {
 			img := image.NewFromIdentifier(strings.TrimSpace(s))
-			if kustomizeImage, ok := annotations[fmt.Sprintf(common.KustomizeApplicationNameAnnotation, img.ImageAlias)]; ok {
+			if kustomizeImage := img.GetParameterKustomizeImageName(annotations); kustomizeImage != "" {
 				img.KustomizeImage = image.NewFromIdentifier(kustomizeImage)
 			}
 			results = append(results, img)
@@ -432,11 +432,9 @@ func GetImagesFromApplication(app *v1alpha1.Application) image.ContainerImageLis
 	// Check the image list for images with a force-update annotation, and add them if they are not already present.
 	annotations := app.Annotations
 	for _, img := range *parseImageList(annotations) {
-		img.ImageTag = nil // the tag from the image list will be a version constraint, which isn't a valid tag
-		if forceStr, force := annotations[fmt.Sprintf(common.ForceUpdateOptionAnnotation, img.ImageAlias)]; force && strings.ToLower(forceStr) == "true" {
-			if images.ContainsImage(img, false) == nil {
-				images = append(images, img)
-			}
+		if img.HasForceUpdateOptionAnnotation(annotations) {
+			img.ImageTag = nil // the tag from the image list will be a version constraint, which isn't a valid tag
+			images = append(images, img)
 		}
 	}
 
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -167,12 +167,12 @@ func FilterApplicationsForUpdate(apps []v1alpha1.Application, patterns []string)
 
 		// Check whether application has our annotation set
 		annotations := app.GetAnnotations()
-		if updateImage, ok := annotations[common.ImageUpdaterAnnotation]; !ok {
+		if _, ok := annotations[common.ImageUpdaterAnnotation]; !ok {
 			log.Tracef("skipping app '%s' of type '%s' because required annotation is missing", app.GetName(), app.Status.SourceType)
 			continue
 		} else {
 			log.Tracef("processing app '%s' of type '%v'", app.GetName(), app.Status.SourceType)
-			imageList := parseImageList(updateImage)
+			imageList := parseImageList(annotations)
 			appImages := ApplicationImages{}
 			appImages.Application = app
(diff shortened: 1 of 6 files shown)


jannfis (commented):
Thanks for your PR @iamnoah and sorry, but I'm having a hard time understanding the reasoning behind it.

The `<alias>.kustomize.image-name` is currently used to define which Kustomize image to override. With this change, the value of that annotation would be included to be checked for new versions, despite we want to check for updates to the overriden image? Why would we also want to check for the original image here?

Again sorry for not understanding it, but could you maybe elaborate a little on the use case?

iamnoah:
> could you maybe elaborate a little on the use case?

Sure. The issue is having a existing deployment like this:

```yaml
spec:
  template:
   spec:
    containers:
      - image: bar:latest
```

If I wasn't using image updater, I would add a kustomization like this:

```yaml
kind: Kustomization
images:
- name: bar
  newName: gcr.io/bar
  newTag: 0.1.0
```

That would update the image to `gcr.io/bar:0.1.0`.

But `.argocd-source.yaml` will override anything I put in `images` with this:

```yaml
images:
- bar:0.1.0
```

When what I wanted was:

```yaml
images:
- bar=gcr.io/bar:0.1.0
```

So I am forced to tell image updater about the re-mapping via `kustomize-name`.


Why does this happen? Because when looking at a deployed app, image updater will check `applicationImages.ContainsImage(applicationImage, false)`. There is no match because `gcr.io/bar` isn't `bar`. 

So we need a different `ContainerImage` to check.  What if we just look for `bar` instead, since that is what `kustomize-name` says to do?
- If the application doesn't have an override yet, we will find `bar` and override it with `gcr.io/bar` as long as the tag is different. If they are the same tag, nothing happens.
- If the image *is* overriden then looking for `bar` won't work because now the application contains only `gcr.io/bar`!

The problem is that the current state of application images can't tell you anything about what was overriden, and if we do fix the match, but have the same tags, no update happens. So we need two fixes:
- Match either `kustomize-name` or the override name to handle both the case where the image has not been update and the one where it has.
- If a `kustomize-name` is present, check if the desired registry is correct.

pkg/argocd/argocd.go
```diff
@@ -183,11 +183,17 @@ func FilterApplicationsForUpdate(apps []v1alpha1.Application, patterns []string)
 	return appsForUpdate, nil
 }
 
-func parseImageList(updateImage string) *image.ContainerImageList {
-	splits := strings.Split(updateImage, ",")
-	results := make(image.ContainerImageList, len(splits))
-	for i, s := range splits {
-		results[i] = image.NewFromIdentifier(strings.TrimSpace(s))
+func parseImageList(annotations map[string]string) *image.ContainerImageList {
```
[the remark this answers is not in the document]
iamnoah: I am fine with that, but I think I have demonstrated here that for non-trivial kustomize use cases, more config than just the image list is necessary, so some complex object or multiple parameters will have to be passed in.

pkg/argocd/argocd.go
```diff
@@ -183,11 +183,17 @@ func FilterApplicationsForUpdate(apps []v1alpha1.Application, patterns []string)
...
+	results := make(image.ContainerImageList, 0)
+	if updateImage, ok := annotations[common.ImageUpdaterAnnotation]; ok {
+		splits := strings.Split(updateImage, ",")
+		for _, s := range splits {
+			img := image.NewFromIdentifier(strings.TrimSpace(s))
+			if kustomizeImage, ok := annotations[fmt.Sprintf(common.KustomizeApplicationNameAnnotation, img.ImageAlias)]; ok {
```
jannfis, line 192: Can you please use `image.GetParameterKustomizeImageName()` instead of going through annotations manually, please?

pkg/argocd/argocd.go
```diff
@@ -425,13 +431,11 @@ func GetImagesFromApplication(app *v1alpha1.Application) image.ContainerImageLis
...
-	if updateImage, ok := annotations[common.ImageUpdaterAnnotation]; ok {
-		for _, img := range *parseImageList(updateImage) {
-			img.ImageTag = nil // the tag from the image list will be a version constraint, which isn't a valid tag
-			if forceStr, force := annotations[fmt.Sprintf(common.ForceUpdateOptionAnnotation, img.ImageAlias)]; force && strings.ToLower(forceStr) == "true" {
-				if images.ContainsImage(img, false) == nil {
-					images = append(images, img)
-				}
+	for _, img := range *parseImageList(annotations) {
+		img.ImageTag = nil // the tag from the image list will be a version constraint, which isn't a valid tag
+		if forceStr, force := annotations[fmt.Sprintf(common.ForceUpdateOptionAnnotation, img.ImageAlias)]; force && strings.ToLower(forceStr) == "true" {
```
jannfis, line 436: We do not have an `image.HasForceUpdateOptionAnnotation()` method yet, but maybe we should introduce one for this purpose?

jannfis (changes requested):
Thanks for your thorough explanation @iamnoah and sorry that it took me a little longer to review, it was a busy week so far.

I have a couple of change requests, please see below.

jannfis (approved):
LGTM now!

Thank you for your contribution and your patience, @iamnoah!

PR #200, opened 2021-04-30
feat: allow write-back to actual kustomization files

### LINKED ISSUE #199: Write Back to kustomization.yml

I like my overrides checked in to git. While `.argocd-source.yml` is nice, in practice, I find that I want to use the overriden images for other purposes too, e.g., setting up a local Minikube environment. It sucks that I can't get updated images when using the same base with local kustomize. The end result is that my local environment has to be manually and painstakingly updated while my dev environment is always up to date.

Update `kustomization.yml` instead. Can be done easily with `kustomize edit set image`. I think it would just look like one more write back method, and be implemented as such.

Other options I considered:

 * Run [a script](https://gist.github.com/iamnoah/0e04d162373ec5df43e02e3fff210083) in my CI that parses the file and runs the right `kustomize edit set image` commands. That's yet another commit, and increases the odds of causing some kind of weird CI+CD feedback loop.
 * A kustomize plugin - painful for end users to consume.

I'll probably do this myself in the next few days and submit a PR, but thought I should put this in to see if anyone else has thoughts on it.

iamnoah:
fixes  #199

I don't love the new type and annotations, this feels like a sub-option of the git method, but I don't have a great idea for better options.


jannfis:
Hey, how about this. We do have `write-back-method` annotation for specifying the method of writing back (e.g. `argocd` and `git` currently), we could have an additional `write-back-target` annotation to specify the target of the write-back.

That new annotation could have values `override[:<app-source.yaml>]` (default, uses the `.app-source.yaml`) or `kustomize[:<kustomization.yaml>]`, each with an optional parameter (after a colon) to specify which file to actually override - e.g. some people don't use `kustomization.yaml` but `Kustomization.yaml` or even plain `Kustomization`. And maybe it can also be used to specify some path within the repository to use.

Oh, and never mind about implementing the `override` option with this PR, we can introduce it along with parameters later on. We should just make sure that the default behavior stays the same after this change (e.g. writes an app source override if `write-back-target` is not given).

jannfis:
Regarding the spell checking of the docs, please add the "unknown" words it complains about (if they are not typos after all) to https://github.com/argoproj-labs/argocd-image-updater/blob/master/.github/actions/spelling/allow.txt

jannfis:
Hey @iamnoah, can you please take a look at the complaints of the linter (as described [here](https://github.com/argoproj-labs/argocd-image-updater/pull/192#issuecomment-819390130))

jannfis:
LGTM, thanks @iamnoah

PR #201, opened 2021-05-02
fix document

ShotaKitazawa:
fix document of https://argocd-image-updater.readthedocs.io/en/stable/configuration/applications


jannfis (approved):
Good catch, thanks a lot @ShotaKitazawa !

LGTM.

PR #212, opened 2021-06-02
feat(manifests): add ssh volume mounts and volume

spoukke:
The default behavior of image updater is to use argocd credentials. If it is setup with ssh, this volume mounts and volume are needed in order for image updater to be able to commit to the repository

### DIFF
--- a/manifests/base/deployment/argocd-image-updater-deployment.yaml
+++ b/manifests/base/deployment/argocd-image-updater-deployment.yaml
@@ -98,16 +98,21 @@ spec:
           initialDelaySeconds: 3
           periodSeconds: 30
         volumeMounts:
-        - name: image-updater-conf
-          mountPath: /app/config
+        - mountPath: /app/config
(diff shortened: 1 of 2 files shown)


jannfis (commented):
Thanks for this PR @spoukke and sorry for the late review.

I feel this mount should be marked as `optional`, because there will be scenarios where this CM does not exist (e.g. user installs the image updater outside `argocd` namespace). Can you please change it so it's an optional mount? Thanks.

spoukke:
> Thanks for this PR @spoukke and sorry for the late review.
> 
> I feel this mount should be marked as `optional`, because there will be scenarios where this CM does not exist (e.g. user installs the image updater outside `argocd` namespace). Can you please change it so it's an optional mount? Thanks.

Hi ! No worries. The change is done, let me know if there is anything else

jannfis (approved):
LGTM, thank you @spoukke !

PR #213, opened 2021-06-11
fix(ext/git): Branch method not call runCmd correctly

huhudev-git:
fix if use `Branch()`, it will type `git` command twice

- current

```sh
git git checkout sourceBranch
git git branch targetBranch
```

- fixed

```sh
git checkout sourceBranch
git branch targetBranch
```


jannfis (approved):
LGTM, thank you @huhudev-git !

PR #241, opened 2021-08-20
Update docs: kustomize.image_name -> kustomize.image-name

robertoriv:
The available annotations table shows `<image_alias>.kustomize.image-name`, but the examples use `<image_alias>.kustomize.image_name`.

`<image_alias>.kustomize.image-name` would be the correct annotation.


jannfis (approved):
Thank you @robertoriv!

LGTM.

PR #251, opened 2021-09-03
fix: modify SetKustomizeImage logic when use alias_name

nyuta01:
In the previous implementation, there was a problem that argocd-image-updater would create a new image when using newName of kustomize.
This is due to the fact that the current kustomize image information retrieved from the argucd API does not include the value of alias_name.
(`[<alias_name>=]<image_path>[:<version_constraint>]`)

So, in this PR, even if the current image information doesn't contain alias_name, we can update the kustomize image information appropriately

### manifest

application.yaml

```yaml
kind: Application
metadata:
  name: adservice
  annotations:
    argocd-image-updater.argoproj.io/image-list: adservice=nyuta01/adservice:^1.0.0
    argocd-image-updater.argoproj.io/write-back-target: kustomization
    argocd-image-updater.argoproj.io/adservice.kustomize.image-name: adservice
spec:
  project: default
  source:
    repoURL: git@github.com:nyuta01/xxx.git
    targetRevision: main
    path: xxx/adservice
...
```

xxx/adservice/kustomization.yaml

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

resources:
- deployment.yaml

images:
- name: adservice
  newName: nyuta01/adservice
  newTag: v1.2.0
```

xxx/adservice/deployment.yaml

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: adservice
spec:
  selector:
    matchLabels:
      app: adservice
  template:
    metadata:
      labels:
        app: adservice
    spec:
      containers:
      - name: server
        image: adservice
...
```

### result

before
```yaml
images:
  - name: adservice
    newName: nyuta01/adservice
    newTag: v1.2.0
+  - name: nyuta01/adservice
+    newTag: v1.2.1

```

after 
```yaml
images:
  - name: adservice
    newName: nyuta01/adservice
-    newTag: v1.2.0
+    newTag: v1.2.1
```

### DIFF
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -414,6 +414,17 @@ func SetKustomizeImage(app *v1alpha1.Application, newImage *image.ContainerImage
 		app.Spec.Source.Kustomize = &v1alpha1.ApplicationSourceKustomize{}
 	}
 
+	for i, kImg := range app.Spec.Source.Kustomize.Images {
+		curr := image.NewFromIdentifier(string(kImg))
+		override := image.NewFromIdentifier(ksImageParam)
+
+		if curr.ImageName == override.ImageName {
+			curr.ImageAlias = override.ImageAlias
+		}
+
+		app.Spec.Source.Kustomize.Images[i] = v1alpha1.KustomizeImage(curr.String())
+	}
+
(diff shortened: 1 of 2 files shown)


jannfis (approved):
LGTM! Thank you, @nyuta01

PR #264, opened 2021-09-27
chore: Test successful update two images

zmx:
ADD a test case for Update Two Images, FIX `result.NumImagesUpdated`


jannfis (approved):
LGTM, thank you @zmx !

PR #283, opened 2021-10-26
chore: Update and refactor Git client

jannfis:
This updates the Git client to the version from Argo CD v2.1.5

Should pull in a couple of enhancements, and also the possibility to use GitHub app authentication for repository interaction.

Signed-off-by: jannfis <jann@mistrust.net>


timothyjlaurent:
I don't think this works. The `GitHubAppCreds` are never returned from this function https://github.com/jannfis/argocd-image-updater/blob/972500fc657f5980915f07317f963b4b091e4ff7/pkg/argocd/gitcreds.go#L49

timothyjlaurent:
Likely related to #613

PR #284, opened 2021-10-28
WIP: add Webhook to reduce no. of registry scans required

tuananh:
this is just a poc. I'm looking for feedbacks. I'm not sure this is the correct/prefer way to implement.

- create  a webhook endpoint for each registry
- upon receiving event from Webhook, only compare the current tag and the incoming tag from Webhook. I'm not sure this would work for all cases.
- with this, we can reduce the `CheckInterval` to reduce no. of registry scan required.
- Only nexus is implemented for now.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -244,7 +244,7 @@ func newRunCommand() *cobra.Command {
 	runCmd.Flags().StringVar(&kubeConfig, "kubeconfig", "", "full path to kubernetes client configuration, i.e. ~/.kube/config")
 	runCmd.Flags().IntVar(&cfg.HealthPort, "health-port", 8080, "port to start the health server on, 0 to disable")
 	runCmd.Flags().IntVar(&cfg.MetricsPort, "metrics-port", 8081, "port to start the metrics server on, 0 to disable")
-	runCmd.Flags().IntVar(&cfg.RegistryWebhookPort, "registry-webhook-port", 8082, "port to start registry webhook server on, 0 to disable")
+	runCmd.Flags().IntVar(&cfg.RegistryWebhookPort, "registry-webhook-port", 0, "port to start registry webhook server on, 0 to disable")
 	runCmd.Flags().BoolVar(&once, "once", false, "run only once, same as specifying --interval=0 and --health-port=0")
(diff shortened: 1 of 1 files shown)


jannfis (commented):
This is an awesome PR :rocket: , thanks a lot @tuananh! And sorry for coming back so late to it.

I'm willing to merge the code early on, so we can progress on its development without you having to constantly rebase it to changes in the master branch. 

However, I have two concerns:

* The web hook feature should be marked experimental and not enabled by default unless we are sure it's stable (and safe) enough to enable it by default. So the default value for `--registry-webhook-port` parameter should be set to `0` so people would have to explicitly enable it. When we had a bit of testing (and possibly refactoring, and bug fixes) we can then enable it by default.

* The way this was implemented is that it trusts the input from the registry webhook, e.g. uses the tag in the hook's payload. This is sensitive, and we should therefore require webhook secrets and not have it optional. Otherwise, anybody could send arbitrary tags and trigger updates to possibly unwanted images.

WDYT?

tuananh:
@jannfis thanks. i'm ok with merging it early on and have it disable. 

i change the default webhook port to 0. 

as for webhook secret. i implemented it for sonatype nexus but dockerhub doesn't support secret in webhook (at least in free version i'm using) so that's why i didn't enforce it.

vr:
@jannfis @tuananh any news since 6 month ago ? 🙏

manelpb:
any updates on this? it would be nice if we could get it merged

tkaesserfm:
Please, any update on this? :pray:

tran4774:
Does it stop updating?? This feature is valuable

hugosxm:
Any update on this one please ?

mubarak-j:
@tuananh is no longer using this project and is abandoning this PR. A new PR to add webhook functionality is needed.

wangtaotao0524:
hi guys, any update on this one? it is an amazing feature.

oubaydos:
any updates?

Mark24Slides:
Same feat, a month ago:
https://github.com/argoproj-labs/argocd-image-updater/pull/1159

chengfang:
See #1159 

Thanks for providing the poc and discussion here!

Issue #287, opened 2021-11-01
Ability to specify filename for helm based git write back target


iangard:
We would like to use argocd image updater to automatically update the development branch of our application config repo and write back the version to git.  This works well.  We would then like to promote to test by merging this file into our test branch (which does not use image updater) and eventually prod branch which also does not use image updater.  The issue is, each of our applications is named differently, myapp-dev, myapp-test, myapp-prod and the file written to git is ".argocd-source-<appname>.yaml".  I would like to make image updater use the ".argocd-source.yaml" instead so it would be honored in all environments like it is documented in Argo docs, https://argo-cd.readthedocs.io/en/stable/user-guide/parameters/#store-overrides-in-git

The solution I am thinking is to make this similar to the kustomize configuration so you could specify:
argocd-image-updater.argoproj.io/write-back-target: "helm:/.argocd-source.yaml"

I think the above gives the most flexibility in the case that Argo starts accepting different filenames in future releases.  You could also implement it as:
argocd-image-updater.argoproj.io/write-back-target: "helm:exclude-appname" 
or something similar to denote that the templated name of .argocd-source-<appname>.yaml should not be used which aligns with ArgoCD's implementation of either having app name or not.

iangard:
After writing this up and discussing with a colleague, we came up with a work-around until this can be implemented.  You can create a symlink from the environment file that you use argocd image updater on to the generic, non app specific filename.  We just tested this out and it works as expected.

`ln -s .argocd-source-myapp-dev.yaml .argocd-source.yaml `
Commit symlink file

PR #289, opened 2021-11-03
add the ability to specify filename for helm based git write back target

rafedramzi:
based on feature request: https://github.com/argoproj-labs/argocd-image-updater/issues/287


iangard (approved):
This is exactly what I had in mind when I opened the issue.  I am not set up to run the tests but the logic looks sound and it meets the requirement

iangard (approved):
This is what I was looking for.  THank you.

jannfis:
Hey, apologies for coming so late to this one! I was just thinking, that this is not specific to Helm, because it really is writing the app source YAML, just to a different path. 

So, I think the `write-back-target` parameter name should not be `helm` (and not exclusively associated to applications of type Helm at all), but probably should be named `app-source` and work for both, Kustomize and Helm (and later on, possibly other) types of applications.

albeldas:
Hi,

any update on this feature request? We really need it.

Thanks.

ixxeL2097:
any update on this ? seems like it would be very nice to merge the feature but still no review

Reyalsorik:
This is what we are looking for!

An update would be greatly appreceiated, thanks! 👍

aqeelat:
I think this was achieved in #551 636

Issue #290, opened 2021-11-03
Inconsistent flag to specify registries config


janpieper:
**Describe the bug**

The flag to specify the registries config file is inconsistent.

**To Reproduce**

```
$ argocd-image-updater run --help | grep registries-conf
      --registries-conf-path string      path to registries configuration file (default "/app/config/registries.conf")
```

vs.

```
$ argocd-image-updater test --help | grep registries-conf
      --registries-conf string      path to registries configuration
```

**Expected behavior**

Have `--registries-conf` **OR** `--registries-conf-path` for both commands.

**Additional context**

_None_

**Version**

0.11.0

**Logs**

_None_

jannfis:
Good catch. I think for better backwards compatibility, we should reuse what's defined for the `run` command, so `--registries-conf-path`.

janpieper-gcx:
Fixed by #295

janpieper-gcx:
@jannfis Hmmm, I cannot close this issue 🤔

PR #295, opened 2021-11-08
Have consistent flag to specify registries conf path

janpieper:
This PR renames `--registries-conf` of the `test`-command to `--registries-conf-path` like we already use for the `run`-command.

The issue has been described in #290.

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -15,17 +15,17 @@ import (
 
 func newTestCommand() *cobra.Command {
 	var (
-		semverConstraint  string
-		strategy          string
-		registriesConf    string
-		logLevel          string
-		allowTags         string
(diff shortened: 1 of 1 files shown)


janpieper:
Is it okay to fully remove `--registries-conf`? This is a **breaking change**, but this project hasn't reached v1.0 yet :thinking:

@jannfis what do you think?

jannfis:
Hey @janpieper, thanks for this PR. I think the `test` command is not (supposed to be) used in any scripts, so the breaking change is totally fine from my side. But thanks for pointing it out. We'll make sure it's mentioned in the next release notes.

jannfis (approved):
LGTM, thanks @janpieper !

PR #298, opened 2021-11-10
docs: Update quoted true, false in configmap

manivannan-g:
The non-quoted true, false variable assignment in the `argocd-image-updater-config` ConfigMap is causing failures. Quoted them now.


jannfis (approved):
LGTM, thank you @manivannan-g for improving the documentation!

PR #300, opened 2021-11-17
feat: use HTTP_PROXY for registry endpoint

derdanne:
use HTTP_PROXY in transport object for registry endpoint if defined in environment

### DIFF
--- a/pkg/registry/endpoints.go
+++ b/pkg/registry/endpoints.go
@@ -212,7 +212,10 @@ func (ep *RegistryEndpoint) GetTransport() *http.Transport {
 	if ep.Insecure {
 		tlsC.InsecureSkipVerify = true
 	}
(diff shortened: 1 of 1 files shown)


jannfis (changes requested):
Thanks a lot for your contribution, @derdanne !

I think it would be easier to use `http.ProxyFromEnvironment` method for setting the `Transport`'s proxy configuration, instead of rolling up our own logic for getting the environment. This would also have the benefit that `https_proxy` and `no_proxy` would be considered as well.

Reference can be found at: https://pkg.go.dev/net/http#ProxyFromEnvironment

derdanne:
Hey thanks for the feedback. I have switched to the http.ProxyFromEnvironment method.

jannfis (approved):
LGTM! Thank you for this contribution, @derdanne !

Issue #198, opened 2021-04-26
Git commit to non existing branch


vsychov:
Hello, 

**Is your feature request related to a problem? Please describe.**
If argo cd application created for non-existing project branch e.g. `my-super-feature-branch` and it's not exists yet in git, I expect that argocd-image-updater can create it for me if docker image for that branch in registry exists (something found according to `image-list` and `allow-tags`).

**Describe the solution you'd like**
Add new annotations:
- `git-base-branch` - git branch, that will be used to create branch, specified in `git-branch`
- `application-type` - specify application type (e.g. Helm or Kustomize), because if branch not exists yet, argocd will not return this information

What do you think?

nick-homex:
@jannfis This can be closed now as #304 is merged.

PR #304, opened 2021-11-25
feat: add ability to specify a different write and base branch

nick-homex:
This add new functionality by adding a write-branch to `argocd-image-updater.argoproj.io/git-branch` using a separator, which will be used to push back changes to the target repository. The branch from the Application's spec or specified with the existing `git-branch` annotation will be used as the base branch for checkout.

Further, this annotation can use Golang templating to allow for dynamic branch names based off of the changes.

This could integrate with various CI/CD (such as GitHub Actions) to automatically create a pull request.

This varies slightly from the suggestion in #198, though sill resolves the issue. Rather than adding a new annotation to specify the base branch, we're specifying a write branch. This decision was made for ease of implementation, as rather than having to add a new annotation and change an existing one, a user will only need to add a new annotation.

### DIFF
--- a/pkg/argocd/git.go
+++ b/pkg/argocd/git.go
@@ -2,7 +2,7 @@ package argocd
 
 import (
 	"bytes"
-	"crypto/sha1"
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 	"io/ioutil"
@@ -82,11 +82,11 @@ func TemplateBranchName(branchName string, changeList []ChangeEntry) string {
 
 	type branchNameTemplate struct {
 		Images []imageChange
-		SHA1   string
+		SHA256   string
 	}
 
 	// Let's add a unique hash to the template
-	hasher := sha1.New()
+	hasher := sha256.New()
 
 	// We need to transform the change list into something more viable for the
 	// writer of a template.
@@ -95,6 +95,7 @@ func TemplateBranchName(branchName string, changeList []ChangeEntry) string {
 		changes = append(changes, imageChange{c.Image.ImageName, c.Image.ImageAlias, c.OldTag.String(), c.NewTag.String()})
 		id := fmt.Sprintf("%v-%v-%v,", c.Image.ImageName, c.OldTag.String(), c.NewTag.String())
 		_, hasherErr := hasher.Write([]byte(id))
+		log.Infof("writing to hasher %v", id)
 		if hasherErr != nil {
 			log.Errorf("could not write image string to hasher: %v", hasherErr)
 			return ""
@@ -103,7 +104,7 @@ func TemplateBranchName(branchName string, changeList []ChangeEntry) string {
 
 	tplData := branchNameTemplate{
 		Images: changes,
-		SHA1:   hex.EncodeToString(hasher.Sum(nil)),
+		SHA256: hex.EncodeToString(hasher.Sum(nil)),
 	}
 
 	err2 := tpl.Execute(&cmBuf, tplData)
@@ -115,7 +116,9 @@ func TemplateBranchName(branchName string, changeList []ChangeEntry) string {
 	toReturn := cmBuf.String()
 
 	if len(toReturn) > 255 {
-		return toReturn[:255]
+		trunc := toReturn[:255]
+		log.Warnf("write-branch name %v exceeded 255 characters and was truncated to %v", toReturn, trunc)
+		return trunc
 	} else {
 		return toReturn
 	}
--- a/pkg/argocd/git.go
+++ b/pkg/argocd/git.go
@@ -82,7 +82,7 @@ func TemplateBranchName(branchName string, changeList []ChangeEntry) string {
 
 	type branchNameTemplate struct {
 		Images []imageChange
-		SHA256   string
+		SHA256 string
 	}
 
 	// Let's add a unique hash to the template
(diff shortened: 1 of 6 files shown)


jannfis:
Thanks @nick-homex and apologies for coming back so late to this PR! It indeed looks useful. I only have minor concerns: Usually, Git branch names have some limit for their length, depending on the Git implementation (I think, GitHub for example maxes out at 255 characters).

Should we take this into consideration when creating the branch name? Maybe use the template you implement as the input for a simple hash algorithm such as SHA256, and use its fixed length hex representation as the name of the branch to create? This way, branch name would stay unique for each updated images combination possible, while not running risk of exceeding the branch name's length restriction.

nick-homex:
> Thanks @nick-homex and apologies for coming back so late to this PR! It indeed looks useful. I only have minor concerns: Usually, Git branch names have some limit for their length, depending on the Git implementation (I think, GitHub for example maxes out at 255 characters).
> 
> Should we take this into consideration when creating the branch name? Maybe use the template you implement as the input for a simple hash algorithm such as SHA256, and use its fixed length hex representation as the name of the branch to create? This way, branch name would stay unique for each updated images combination possible, while not running risk of exceeding the branch name's length restriction.

Hey! Sorry for the late response, was on holidays. A simpler solution might just be to truncate the branch name if it is too long. The chance of collisions is possible, though rare if the user includes the version numbers in their template. Perhaps add the hashed value as a top-level template variable that they could use instead? That way we leave it up to the user if they want to use a guaranteed unique branch name, or if they would prefer one that reflects the images changed (but might get truncated).

nick-homex:
@jannfis I've added a SHA1 hash as a template option and truncated the length to 255 if it exceeds that, let me know if you have any other issues!

pkg/argocd/git.go
```diff
@@ -58,11 +60,72 @@ func TemplateCommitMessage(tpl *template.Template, appName string, changeList []
...
+		NewTag string
+	}
+
+	type branchNameTemplate struct {
+		Images []imageChange
+		SHA1   string
+	}
+
+	// Let's add a unique hash to the template
+	hasher := sha1.New()
```
jannfis, line 89: Can you please use `sha256` instead of `sha1`? I know the implications for this particular use-case are marginal, but SHA1 is going to go away rather soon and should not be used anymore.

pkg/argocd/git.go
```diff
@@ -58,11 +60,72 @@ func TemplateCommitMessage(tpl *template.Template, appName string, changeList []
...
+	err2 := tpl.Execute(&cmBuf, tplData)
+	if err2 != nil {
+		log.Errorf("could not execute template for Git branch name: %v", err2)
+		return ""
+	}
+
+	toReturn := cmBuf.String()
+
+	if len(toReturn) > 255 {
+		return toReturn[:255]
```
jannfis, line 118: I think we should issue a warning to the log in case the branch name is truncated.

jannfis (changes requested):
Thanks for the update, @nick-homex !

I have two comments that need to be addressed, please check below.

Also, I was thinking about the following: Instead of introducing a new annotation for the write-back branch, how about making this configuration part of the `git-branch` annotation, e.g.:

```
argocd-image-updater.argoproj.io/alias.git-branch: <checkout branch>:<write-branch>
```
AFAIK, the colon `:` is not a valid character for Git branch names, so could be used as a separator. Behavior would be as follows:

1. Use checkout branch from Argo CD application, and no write back branch: Annotation not specified or empty value
2. Use specific checkout branch without write-back-branch: `somebranch"
3. Use specific checkout branch with write-back-branch: `somebranch:otherbranch`
4. Use checkout branch from Argo CD application with write-back-branch: `:otherbranch`

I'm not quite sure from the usability point of view here. But we're trying to refrain from new annotations when possible, just to prevent something like annotation pollution in the Application CR.

I'm happy to merge this PR without above mentioned thing changed, however. Just some food for thought, and discussion.

pkg/argocd/update.go
```diff
@@ -449,6 +450,10 @@ func parseGitConfig(app *v1alpha1.Application, kubeClient *kube.KubernetesClient
 	if ok {
 		wbc.GitBranch = strings.TrimSpace(branch)
 	}
+	writebranch, ok := app.Annotations[common.GitWriteBranchAnnotation]
+	if ok {
+		wbc.GitWriteBranch = strings.TrimSpace(writebranch)
+	}
```
jannfis, line 456: Also, instead of accessing annotations directly, can you please add an appropriate method to get the value from the app to [here](https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/image/options.go)

nick-homex: The file you linked is for per-image configuration? It looks like all the app-level config (like all the existing code for `WriteBackConfig` is already accessed in this way.

nick-homex:
@jannfis Done the requested changes, good idea on using one annotation and the separator.

pkg/argocd/update.go
jannfis: Yeah, you're right - my fault, sorry. 

I should refactor the code to encapsulate the global options into getter methods as well some day :) Never mind then - and with encoding the write back branch in the same annotation as the checkout branch, this change is also not necessary anymore.

jannfis (approved):
LGTM!

Thanks for this awesome contribution, @nick-homex!

PR #307, opened 2021-11-29
feat: Enable auth plugins for Kubernetes client

joyrex2001:
This PR enables the default auth plugins in order to obtain credentials from cloud providers when required.


jannfis (approved):
LGTM, thanks @joyrex2001 !

PR #309, opened 2021-12-02
feat: Assume app type when write-back-type is set to kustomize

iamnoah:
Recently started using tooling based on kustomize that is configured in Argo as a plugin. It stopped getting updates, so this is a workaround to continue to get kustomize updates.


jannfis:
Thanks, @iamnoah! I personally think that we should not have an override for that, but rather don't care about the application type when the `write-back-type` is set to `kustomize`

WDYT?

iamnoah:
> don't care about the application type when the `write-back-type` is set to `kustomize`

That makes sense to me, I'll adjust the logic when I can.

jannfis (approved):
LGTM, thanks @iamnoah !

Issue #210, opened 2021-05-25
Invalid reference format for digest update strategy


LS80:
I tried out the digest update strategy for a Helm application (introduced in https://github.com/argoproj-labs/argocd-image-updater/pull/176) and found that the default behaviour will produce an invalid reference format `some/image:sha256:<somelonghashstring>`. This should be `some/image@sha256:<somelonghashstring>`, so it appears that the `sha256:` is being considered part of the "tag". If using the `helm.image-spec` annotation the format is valid.

Once solution would be to append `@sha256` to the image name instead of making it part of the tag? Or at least document that the `helm.image-spec` annotation is required?

jannfis:
Hey @LS80, good point there. 

However, since it's the Helm template which constructs the final image name from the parameters given, are there any best practices documentation for Helm how to properly handle SHAs instead of tags, i.e. how to pass them to the template?

Because in the end, it's up to the Helm template whether to render input variables `image.name=foo` and `image.tag=sha256:somehash` as `foo@sha256:somehash` or as `foo:sha256:somehash`.

LS80:
I'm unaware of any Helm best practices for that, but perhaps the simplest fix is to always include the tag name as well as the digest. So set `image.tag=latest@sha256:somehash`.

Of course it might not be `latest`, so if we set

```
argocd-image-updater.argoproj.io/image-list: someimage=some/image:sometag
argocd-image-updater.argoproj.io/someimage.update-strategy: digest
```

Then it should set `image.tag=sometag@sha256:somehash`.

I think that would be compatible with existing Helm templates that use a tag parameter, so there should be no need for an additional Helm parameter for the digest. Do you think that would work?

P.S. This is enough reason to use Kustomize instead of Helm in my opinion 😄

jannfis:
Hey @LS80, sorry for coming back so late to this. I agree that probably the format you are refering (`sometag@sha256:somehash`) is the way to go forward for compatibility. I will check how that works out and if it does, publish a patch release shortly.

imdbere:
Are there any updates on this ?

mehrdadpfg:
same issue here
looking forward for updates

ElArtista:
Similar to #259

dwi-mfbcs:
Given the fix mentioned above, does it address the issue? What's the ETA for including the fix in a release?

jannfis:
[...]

PR #317, opened 2021-12-12
fix: Invalid image reference format when using digest strategy with helm charts

### LINKED ISSUE #259: Digest is not working (Digest, Tag, Helm)
**Describe the bug**
argocd-image-updater: v0.10.3
argocd: v2.1.2
gitrepo: helm

Strategy
```
argocd-image-updater.argoproj.io/write-back-method: argocd
argocd-image-updater.argoproj.io/image-list: myimagealias=gcr.io/project-name/images-path/my-image-name:my-tag-name
argocd-image-updater.argoproj.io/myimagealias.update-strategy: digest

```

I'm using the digest strategy for image updating and from argocd-image-updater end all seems to work fine, and I'm getting the successfully updated message. Yet ArgoCD doesn't seem to update the image.

```
time="2021-09-16T19:47:35Z" level=info msg="Successfully updated the live application spec" application=dev-environment
```

When I check the Argo CD, I can see the following has been added to parameters, but it does't trigger an update.
```
image.name
  gcr.io/project-name/images-path/my-image-name
image.tag
 sha256:xxxxxxx
```

My approach / requirement is to have  argocd as the write-back method and only the team members updating the deployment repo with image details and repo tags per each environment.

1. Can anyone tell me what I might be doing wrong here?
2. Any help on fixing this issue is appreciated

ElArtista:
Fixes #259 #210

When using digest update strategy, the helm image.tag parameter now composes a full tag of both the tag and the digest in the form:
```
sometag@sha256:<somelonghashstring>
```
This avoids invalid references in almost all helm charts that construct the image ref as `image.repository:image.tag`, that so far resulted to the invalid references of the form:
```
some/image:sha256:<somelonghashstring>
```

### DIFF
--- a/pkg/image/image.go
+++ b/pkg/image/image.go
@@ -105,8 +105,8 @@ func (img *ContainerImage) GetFullNameWithTag() string {
 	return str
 }
 
-// GetFullTag returns tag name along with any tag digest set for the image
-func (img *ContainerImage) GetFullTag() string {
+// GetTagWithDigest returns tag name along with any tag digest set for the image
+func (img *ContainerImage) GetTagWithDigest() string {
 	str := ""
 	if img.ImageTag != nil {
 		if img.ImageTag.TagName != "" {
--- a/pkg/image/image.go
+++ b/pkg/image/image.go
@@ -105,6 +105,24 @@ func (img *ContainerImage) GetFullNameWithTag() string {
 	return str
 }
 
+// GetFullTag returns tag name along with any tag digest set for the image
+func (img *ContainerImage) GetFullTag() string {
+	str := ""
+	if img.ImageTag != nil {
+		if img.ImageTag.TagName != "" {
+			str += img.ImageTag.TagName
+		}
+		if img.ImageTag.TagDigest != "" {
+			if str == "" {
+				str += "latest"
+			}
+			str += "@"
+			str += img.ImageTag.TagDigest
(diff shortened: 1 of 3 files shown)


pkg/image/image.go
```diff
@@ -105,6 +105,24 @@ func (img *ContainerImage) GetFullNameWithTag() string {
 	return str
 }
 
+// GetFullTag returns tag name along with any tag digest set for the image
+func (img *ContainerImage) GetFullTag() string {
```
jannfis, line 109: Can we please rename this to something more precise, e.g. `GetTagWithDigest()`?

jannfis (changes requested):
Thanks a lot, @ElArtista - this is useful!

I only have one cosmetic concern, please see below.

Also, can you please provide unit tests for the method?

pkg/image/image.go
ElArtista: Will do immediately!

ElArtista: Done.

ElArtista:
Added unit tests, I believe this is now ready!

jannfis (approved):
LGTM. Thanks a lot @ElArtista !

dwi-mfbcs:
Waiting for this fix to be released. Do we have an ETA for a release that will include this fix? Thanks.

PR #325, opened 2021-12-22
docs: Fixed ConfigMap name

martin-marko:
The docs mentioned `argocd-image-updater-cm` while the correct name is `argocd-image-updater-config`


jannfis (approved):
LGTM! Thanks @martin-marko

PR #337, opened 2022-01-09
fix: Do not retrieve metadata for all tags with digest strategy

### LINKED ISSUE #313: hitting api requests limit in gcr
**Describe the bug**

we have about 600-700 tags in gcr registry for all the images

updater makes about 5000 api requests to gcr registry each time

running with default interval of 2 min we are hitting the limit of api requests

we are using tags like "prod" to deploy to prod "dev" for dev

+
**Expected behavior**

only make one request per sha, also start from the end and if the needed tag is found , stop searching

may be some other solution?

**Version**
v0.10.3

**Logs**

time="2021-12-07T11:43:59Z" level=error msg="error fetching metadata for dev-collection/data-utils-server:0.46: Get \"https://gcr.io/v2/ZZZ/YYY/blobs/sha256:97a804dd72026a88ca41869d3d998033ce8c1073af2c621616e15d059a4fb98a\": http: non-successful response (status=429 body=\"Quota Exceeded.\")"

jannfis:
When using digest update strategy, the Image Updater would have fetched metadata for all tags returned by the registry. For digest strategy, caching is also disabled and this leads to too many requests to the registry.

Now, only metadata for the tag specified in the version constraint is fetched. All other tags are skipped.

Fixes #313

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -51,10 +51,18 @@ func (endpoint *RegistryEndpoint) GetTags(img *image.ContainerImage, regClient R
 
 	tags := []string{}
 
-	// Loop through tags, removing those we do not want
-	if vc.MatchFunc != nil || len(vc.IgnoreList) > 0 {
+	// For digest strategy, we do require a version constraint
+	if vc.SortMode == image.VersionSortDigest {
+		if vc.Constraint == "" {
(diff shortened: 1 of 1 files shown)


diranged:
@jannfis Thank you for fixing this - it decreased our API calls by many of orders of magnitude, and dramatically increases the responsiveness of the image updater for our developers!

Issue #314, opened 2021-12-10
Metric for registry errors


janpieper-gcx:
**Is your feature request related to a problem? Please describe.**

The documentation lists `argocd_image_updater_registry_errors_total` and this metrics is used in the example Grafana dashboard, but it is not implemented.

**Describe the solution you'd like**

Implement the metric.

**Describe alternatives you've considered**

Maybe it was implemented in the past and removed for some reason, but couldn't find any information about that.

**Additional context**

https://github.com/argoproj-labs/argocd-image-updater/blob/da58411eea9f98811f55cc10b92c150385e2d9ae/docs/install/start.md#L309

https://github.com/argoproj-labs/argocd-image-updater/blob/da58411eea9f98811f55cc10b92c150385e2d9ae/config/example-grafana-dashboard.json#L518

pacoguzman:
Working on this as my `first issue`

pacoguzman:
I've just found that metric is already implemented

https://github.com/argoproj-labs/argocd-image-updater/blob/7ed1f675ca2373dd8f0442e1a398e3fc9a227c55/pkg/metrics/metrics.go#L158-L161

https://github.com/argoproj-labs/argocd-image-updater/blob/7ed1f675ca2373dd8f0442e1a398e3fc9a227c55/cmd/run.go#L322

is the confusion due to the naming?

pacoguzman:
Opps sorry my fault the definition for that metric is here

https://github.com/argoproj-labs/argocd-image-updater/blob/7ed1f675ca2373dd8f0442e1a398e3fc9a227c55/pkg/metrics/metrics.go#L85-L88

And that metric is not what you pointed out.

So forget about this I'll keep looking

jannfis:
Closed by #362

PR #362, opened 2022-01-22
fix: Use existing registry requests failed metric

pacoguzman:
This PR fixes the issue raised at https://github.com/argoproj-labs/argocd-image-updater/issues/314

The documentation and the grafana dashboard example are pointing to a not implemented metric instead point to the correct one `argocd_image_updater_registry_requests_failed_total`


jannfis (approved):
LGTM! Good catch, thank you!

PR #364, opened 2022-01-22
fix(manifests): Set deployment strategy to recreate to prevent concurrent replicas

### LINKED ISSUE #359: Concurrent image-updaters running during rollout
**Describe the bug**

It is expressed in the documentation to not run more than one image-updater (per application, at least), but I notice when doing a rollout (eg: for a config or secret change) the old pod stays up until the new one is up. I wonder if that can cause problems under certain conditions?

image

**To Reproduce**

`kubectl -n argocd rollout restart deployment argocd-image-updater`

**Expected behavior**

Given the guidance in the documentation about concurrent instances, I would expect no more than one replica running at a given time, and in the case of a failed deployment, rollback is done.

**Additional context**

https://github.com/argoproj-labs/argocd-image-updater/blob/7ed1f675ca2373dd8f0442e1a398e3fc9a227c55/manifests/base/deployment/argocd-image-updater-deployment.yaml#L1-L8

**Version**

```
argocd-image-updater: v0.11.3+f62b7d7
  BuildDate: 2022-01-19T22:16:42Z
  GitCommit: f62b7d7841a42030fe4c75de1951c46b0e7fca2e
  GoVersion: go1.16.13
  GoCompiler: gc
  Platform: linux/amd64
```

**Logs**

N/A

NickLarsenNZ:
Set `strategy: Recreate` to stop rolling deployments with concurrent replicas

- Fixes #359

---

Looks like the workflow uses Kustomize 2.0.2. Newer versions replace `bases` with `resources` (see  b432ebe), which seems to [cause the `make manifests` check to fail](https://github.com/argoproj-labs/argocd-image-updater/runs/4909466408?check_suite_focus=true).

I used the following which failed, so I reverted the change it made.

```
{Version:kustomize/v4.4.1 GitCommit:b2d65ddc98e09187a8e38adc27c30bab078c1dbf BuildDate:2021-11-11T23:27:14Z GoOs:darwin GoArch:amd64}
```

### DIFF
--- a/manifests/base/deployment/argocd-image-updater-deployment.yaml
+++ b/manifests/base/deployment/argocd-image-updater-deployment.yaml
@@ -10,6 +10,8 @@ spec:
   selector:
    matchLabels:
     app.kubernetes.io/name: argocd-image-updater
+  strategy:
+    type: Recreate
   template:
     metadata:
       labels:
--- a/manifests/install.yaml
+++ b/manifests/install.yaml
@@ -85,6 +85,8 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/name: argocd-image-updater
+  strategy:
+    type: Recreate
   template:
     metadata:
       labels:


jannfis (approved):
LGTM. Thanks @NickLarsenNZ !

Issue #258, opened 2021-09-16
Support match applications options beside name


mubarak-j:
**Is your feature request related to a problem? Please describe.**
Currently image-updater seems to come with option to filter or match applications by name e.g `--match-application-name staging-*`. It's common to use k8s labels to filter applications which is currently not supported

**Describe the solution you'd like**
It would be great if image-updater can match applications by labels

**Describe alternatives you've considered**
We considered using `ignore-tags` option but it's not suitable in this use case. 

**Additional context**

jannfis:
Oh, this issue must have been slipped through, sorry. 

I agree, this is a great idea to filter applications by labels and would be nice to have.

jaideepr97:
Hey @jannfis could you assign this to me?
Thanks

jannfis:
Thanks @jaideepr97 - let me know any questions you may have during implementation :+1:

jannfis:
Implemented by #368

PR #368, opened 2022-01-24
feat: support filtering apps by label in CLI

jaideepr97:
This PR addresses issue #258 
It adds a new CLI parameter `--match-application-label` which accepts a string label input to match applications against
It currently only accepts a single label to filter against, and does not support pattern matching on label values, but those enhancements can be made in future PRs 

steps to test out PR:
- create annotated apps containing 1 or more labels 
- set up argocd-image-updater account & token as per documented steps
- run argocd-image-updater binary with additional  parameter `--match-application-label` and supply desired label

application should be updated with new image after sync

### DIFF
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -145,10 +145,34 @@ func nameMatchesPattern(name string, patterns []string) bool {
 	return false
 }
 
+// Match app labels against provided filter label
+func matchAppLabels(appName string, appLabels map[string]string, filterLabel string) bool {
+
+	if filterLabel == "" {
+		return true
+	}
+
+	filterLabelMap, err := parseLabel(filterLabel)
+	if err != nil {
+		log.Errorf("Unable match app labels against %s: %s", filterLabel, err)
+		return false
+	}
+
+	for filterLabelKey, filterLabelValue := range filterLabelMap {
+		log.Tracef("Matching application name %s against label %s", appName, filterLabel)
+		if appLabelValue, ok := appLabels[filterLabelKey]; ok {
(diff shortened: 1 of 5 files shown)


pkg/argocd/argocd.go
```diff
@@ -198,6 +228,20 @@ func parseImageList(annotations map[string]string) *image.ContainerImageList {
 	return &results
 }
 
+func parseLabel(inputLabel string) (map[string]string, error) {
+	var selectedLabels map[string]string
+	var labelFieldDelimiter = "="
```
jannfis, line 233: Nit: `labelFieldDelimeter` should be a const instead of a var.

pkg/argocd/argocd.go
```diff
@@ -145,10 +145,34 @@ func nameMatchesPattern(name string, patterns []string) bool {
...
+// Match app labels against provided filter label
+func matchAppLabels(appName string, appLabels map[string]string, filterLabel string) bool {
+
+	if filterLabel == "" {
+		return true
+	}
+
+	filterLabelMap, err := parseLabel(filterLabel)
+	if err != nil {
+		log.Tracef("Unable match app labels against %s: %s", filterLabel, err)
```
jannfis, line 157: Should this maybe be logged out as error instead?

jannfis (commented):
Thanks @jaideepr97 - I have just two nits, please see below.

jaideepr97:
Thanks @jannfis 
addressed your comments

jannfis (approved):
LGTM! Thanks @jaideepr97 !

PR #373, opened 2022-02-06
feat: Add configMap to allow define parameters for SSH client

zimbres:
Since ssh-rsa has been deprecated since OpenSSH 8.2 but some git server are still using it like Azure DevOps, its necessary to bring back this feature.

Besisdes, since for internal networks, some people use not have too much concern about deprecated algorithms, this PR can give then hability to change the ssh client and make it compatible with their systems.

configMap example:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-image-updater-ssh-config
  namespace: argocd
data:
  config: |
    Host *
          PubkeyAcceptedAlgorithms +ssh-rsa
          HostkeyAlgorithms +ssh-rsa
```

This will be loaded in /app/.ssh/config as default user home path in image.

### DIFF
--- a/config/example-config-ssh-config.yaml
+++ b/config/example-config-ssh-config.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: argocd-image-updater-ssh-config
+  namespace: argocd
+data:
+  config: |
+    Host *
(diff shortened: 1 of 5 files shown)


jannfis (approved):
LGTM

Thank you @zimbres !

PR #384, opened 2022-02-20
Checkout after creating a new branch

katainaka0503:
I'm a big fan of the new feature below. It will allow us more flexible GitOps operations.

https://github.com/argoproj-labs/argocd-image-updater/pull/304

However, I found a bug after trying the feature by myself on my local environment.

It seems after creating a new branch, `Argo CD` doesn't checkout the newly created branch. In the end, the new branch is pushed without any changes committed.

So, I've moved checkout process to after branch creation.

### DIFF
--- a/pkg/argocd/update_test.go
+++ b/pkg/argocd/update_test.go
@@ -1786,7 +1786,7 @@ func Test_CommitUpdates(t *testing.T) {
 		defer cleanup()
 		gitMock.On("Checkout", mock.Anything).Run(func(args mock.Arguments) {
 			args.Assert(t, "mydefaultbranch")
-		}).Return(nil)
+		}).Return(nil).Once()
 		gitMock.On("Add", mock.Anything).Return(nil)
 		gitMock.On("Branch", mock.Anything, mock.Anything).Return(nil)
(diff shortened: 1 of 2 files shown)


pkg/argocd/update_test.go
```diff
@@ -1786,7 +1786,7 @@ func Test_CommitUpdates(t *testing.T) {
 		defer cleanup()
 		gitMock.On("Checkout", mock.Anything).Run(func(args mock.Arguments) {
 			args.Assert(t, "mydefaultbranch")
-		}).Return(nil)
+		}).Return(nil).Once()
```
jannfis, line 1789: Can you please explain why the `.Once()` is necessary?

jannfis (commented):
Thanks @katainaka0503! I just have a question, please see below.

@nick-homex since you initially contributed the write-back branch feature, can you please have a look at this as well?

pkg/argocd/update_test.go
katainaka0503: Thank you for your comment.

This is because in the current code, after checking out the source branch, argo cd checks out push branch .

But, technically, this is not necessary.
I'll fix it.

katainaka0503: Fixed it!

jannfis (approved):
LGTM! Thank you, @katainaka0503 !

PR #407, opened 2022-03-23
Support using regex capture groups to clean up the semver of a tag

djeebus:
This allows argocd-image-updater to support finding the latest tag even when they don't _really_ follow semver exactly. Some good examples of this:
- [plexinc/pms-docker](https://hub.docker.com/r/plexinc/pms-docker/) example: 1.25.7.5604-980a13e02 (semver requires the final character to be a `+`, not `-`)
- [fireflyiii/core](https://hub.docker.com/r/fireflyiii/core/) example: version-5.6.16 (leading chars should be `v`, not `version-`)

This PR would let us set the `regexp` to be (in pms-docker's case) `^(\d+\.\d+\.\d+)\.\d+-.*$'`, or just `(\d+\.\d+\.\d+)`. The code checks for groups, and if none are found, assumes that the full tag is the semver (same as current behavior). If a capture group is found, however, then it parses the capture group as a semver, rather than the full thing.

Please let me know if there's anything I can do to help this along, or if I'm out-of-scope for this project, and I'll figure something out. This would help me keep a lot of my homelab images up to date though. Thanks!!


pkg/tag/tag.go
```diff
@@ -14,9 +14,11 @@ import (
 // ImageTag is a representation of an image tag with metadata
 // Use NewImageTag to to initialize a new object.
 type ImageTag struct {
-	TagName   string
-	TagDate   *time.Time
-	TagDigest string
+	TagName     string
+	TagNamePart string
```
jannfis, line 18: I think this field would need some additional comment on what it actually represents, it is not clear to me.

pkg/registry/registry.go
```diff
@@ -88,7 +110,7 @@ func (endpoint *RegistryEndpoint) GetTags(img *image.ContainerImage, regClient R
 			} else if endpoint.TagListSort == TagListSortLatestLast {
 				ts = i
 			}
-			imgTag := tag.NewImageTag(tagStr, time.Unix(int64(ts), 0), "")
+			imgTag := tag.NewImageTagWithTagPart(tagStr.original, tagStr.found, time.Unix(int64(ts), 0), "")
```
jannfis, line 113: Hm. I think `tagStr` should be renamed to `tagInfo` or something like that in the loop to better reflect the fact that it's not a string anymore. Using `tagStr` was a bad decision in the first place, but now it just becomes confusing :)

pkg/tag/tag.go
```diff
@@ -47,13 +49,34 @@ func (il SortableImageTagList) Swap(i, j int) {
 
 // NewImageTag initializes an ImageTag object and returns it
 func NewImageTag(tagName string, tagDate time.Time, tagDigest string) *ImageTag {
+	return NewImageTagWithTagPart(tagName, tagName, tagDate, tagDigest)
+}
+
+// NewImageTagWithTagPart initializes an ImageTag object and returns it
```
jannfis, line 55: Please be more specific what the difference to `NewImageTag` is. Also I was wondering if this should be rather something more along the lines of the following, instead of a new initializer:

```go
func (t *ImageTag) WithTagNamePart(tagNamePart string) *ImageTag {
  t.TagNamePart = tagNamePart
  return t
}

tag := NewImageTag(...).WithTagNamePart(...)

```
WDYT?

jannfis (commented):
Thanks @djeebus !

This indeed looks useful. However, I'm not so sure whether it's a good idea to reuse `allowed-tags` for transforming the tag name to a version though, even if it may seem reasonable and elegant currently. But it would not allow us to use multiple regexps in the future for `allow-tags`, or introduce a simpler matcher (e.g. a glob) to be combined with `allowed-tags` in a semver update strategy.

I think we may want to introduce a new annotation, e.g. `semver-transformer` that will be used for this transformation.

I'm happy to discuss the pros and cons of both before you make a change, tho.

jannfis:
The `allow-tags` is usually not required in a `semver` context, because the allowed versions are specified using semver constraint in the `image-list` annotation. In the `semver` strategy, we ignore any tag that doesn't parse into a semver anyway.

For example, if you want to stay within 1.2 minor version, you'd configure

```
argocd-image-updater.argoproj.io/image-list: fireflyiii=some/where/fireflyiii:1.2.x
```

I would suggest the following for this change:

* Early in the tag processing, apply any transformation from the transformer annotation
* Use existing semver constraint checking against the extracted tag from the transformer if applicable, otherwise use the original tag value

djeebus:
@jannfis sorry, that was a bad example, just wanted to make sure the names made sense. I've submitted a new PR (#413), as the new approach was different enough from this that I wanted to keep the diff clean. Feel free to close this one!

PR #413, opened 2022-04-06
Add a semver-transformer annotation to assist in converting sloppy tags to semvers

djeebus:
This allows argocd-image-updater to support finding the latest tag even when they don't _really_ follow semver exactly. Some good examples of this:
- [plexinc/pms-docker](https://hub.docker.com/r/plexinc/pms-docker/) example: `1.25.7.5604-980a13e02` (semver requires at most 3 numeric sections, i.e. 1.2.3, not 1.2.3.4)
- [fireflyiii/core](https://hub.docker.com/r/fireflyiii/core/) example: `version-5.6.16` (leading chars should be `v`, not `version-`)

This PR would let us set the `regexp` to be (in pms-docker's case) `^(\d+\.\d+\.\d+)\.\d+-.*$'`, or just `(\d+\.\d+\.\d+)`. The code checks for groups, and if none are found, assumes that the full tag is the semver (same as current behavior). If a capture group is found, however, then it parses the capture group as a semver, rather than the full thing.

A few other small things added: 
- the `test` command now supports `--semver-transformer`
- if `test` can't parse `--semver-transformer` or `--allow-tags`, it exits with an error message
- if `update` can't parse `--semver-transformer` or `--allow-tags`, it skips the image with an error message, making it more obvious that you've configured something poorly
- `MatchFuncRegexp` now creates a closure, removing the need for a separate `MatchArgs interface{}` on the `VersionConstraint` struct
- removed `ConstraintMatchMode and its associated consts in `pkg/image/version.go`, as they were completely unused.
- split up some of the matching logic in `pkg/registry/registry.go` so it is more clear why an image is skipped when tracing.

let 
(This supercedes #407)


djeebus:
sorry, looks like I missed a code path. let me work on this a bit and see if I can slim it down some

djeebus:
Two things have stopped me from getting this done :( 

1. found out that there was a major code path missing - I need to use the `semver-transformer` in order to transform the _current_ image in order to compare it against _new_ images. That proved to be a little tedious, I'm headed into code that I haven't seen yet, so it's gonna take some more time to sort that out.
2. Life :( 

Once I get #2 sorted out, I'll get back to #1. Thanks for understanding!

RuiSMagalhaes:
any news on this @djeebus ?

HuseyinCoinmerce:
Would also be nice to have this feature

djeebus:
Sorry, had twins and lost my spare time. If anyone wants to pick this up, I'd be happy to offer what little help I can!

PR #428, opened 2022-04-29
Support GPG-signed commits, fixes #427

### LINKED ISSUE #427: Git write-back method and gpg-signed commits
We use `git` write-back method, but our repository has branch protection rules, which require all commits to `main` branch to be signed with a gpg key. It looks like there is partial support for passing gpg key id to the `git commit` command [here](https://github.com/argoproj-labs/argocd-image-updater/blob/master/ext/git/writer.go#L16-L17), but I can not find a way to configure both `CommitOptions.SigningKey` and `CommitOptions.SignOff` options.

Support configuring these setting via `--git-commit-signing-key` and `--git-commit-sign-off`cli args and `GIT_COMMIT_SIGNING_KEY` and `GIT_COMMIT_SIGN_OFF` env vars [here](https://github.com/argoproj-labs/argocd-image-updater/blob/89daab684fcdac4c4d3a0683fe4764a9c1abd0b4/cmd/run.go#L223-L225) 

Disable `require signed commits` requirement in `main` branch protection rules.

vgrigoruk:
Fixes #427

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -44,6 +44,8 @@ type ImageUpdaterConfig struct {
 	GitCommitUser       string
 	GitCommitMail       string
 	GitCommitMessage    *template.Template
+	GitCommitSigningKey string
+	GitCommitSignOff    bool
 	DisableKubeEvents   bool
 }
 
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -222,6 +222,8 @@ func newRunCommand() *cobra.Command {
 	runCmd.Flags().BoolVar(&warmUpCache, "warmup-cache", true, "whether to perform a cache warm-up on startup")
 	runCmd.Flags().StringVar(&cfg.GitCommitUser, "git-commit-user", env.GetStringVal("GIT_COMMIT_USER", "argocd-image-updater"), "Username to use for Git commits")
 	runCmd.Flags().StringVar(&cfg.GitCommitMail, "git-commit-email", env.GetStringVal("GIT_COMMIT_EMAIL", "noreply@argoproj.io"), "E-Mail address to use for Git commits")
+	runCmd.Flags().StringVar(&cfg.GitCommitSigningKey, "git-commit-signing-key", env.GetStringVal("GIT_COMMIT_SIGNING_KEY", ""), "GnuPG key ID used to sign the commits")
+	runCmd.Flags().BoolVar(&cfg.GitCommitSignOff, "git-commit-sign-off", env.GetBoolVal("GIT_COMMIT_SIGN_OFF", false), "Whether to sign-off git commits")
 	runCmd.Flags().StringVar(&commitMessagePath, "git-commit-message-path", defaultCommitTemplatePath, "Path to a template to use for Git commit messages")
 	runCmd.Flags().BoolVar(&cfg.DisableKubeEvents, "disable-kube-events", env.GetBoolVal("IMAGE_UPDATER_KUBE_EVENTS", false), "Disable kubernetes events")
 
--- a/manifests/base/deployment/argocd-image-updater-deployment.yaml
+++ b/manifests/base/deployment/argocd-image-updater-deployment.yaml
@@ -78,6 +78,18 @@ spec:
               name: argocd-image-updater-config
               key: git.email
               optional: true
+        - name: GIT_COMMIT_SIGNING_KEY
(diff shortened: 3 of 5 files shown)


jannfis:
Sorry for being so late on this PR!

Reading through the code, I have the feeling that some part of the code may be missing? I see configuration around setting the signing key and sign-off functions, but I don't see any code changes to the Git client to actually perform the signing or sign-off.

vgrigoruk:
Hey @jannfis . Thank you for looking into this. 
As far as I can see, the code which adds necessary flags to `git commit` CLI command is already there (but was previously unused):
- https://github.com/argoproj-labs/argocd-image-updater/blob/a03f31915d71ae39f1c9ac5d7d4d1501d11d158b/ext/git/writer.go#L34-L39

jannfis:
> Hey @jannfis . Thank you for looking into this. As far as I can see, the code which adds necessary flags to `git commit` CLI command is already there (but was previously unused):
> 
>     * https://github.com/argoproj-labs/argocd-image-updater/blob/a03f31915d71ae39f1c9ac5d7d4d1501d11d158b/ext/git/writer.go#L34-L39

Oops. Thanks, then it was probably dead code all the time. Would you mind writing documentation for this? In order to sign commits with PGP, I assume some private key ring needs to be mounted to the pod. Could you include step-by-step instructions for people on how to do this?

Thanks!

ForbiddenEra:
> > Hey @jannfis . Thank you for looking into this. As far as I can see, the code which adds necessary flags to `git commit` CLI command is already there (but was previously unused):
> > ```
> > * https://github.com/argoproj-labs/argocd-image-updater/blob/a03f31915d71ae39f1c9ac5d7d4d1501d11d158b/ext/git/writer.go#L34-L39
> > ```
> 
> Oops. Thanks, then it was probably dead code all the time. Would you mind writing documentation for this? In order to sign commits with PGP, I assume some private key ring needs to be mounted to the pod. Could you include step-by-step instructions for people on how to do this?
> 
> Thanks!

Have gpg+gnupg installed, and then something like:

```
    mkdir -p ~/.gnupg && chmod 600 ~/.gnupg
    echo "pinentry-mode loopback" >> ~/.gnupg/gpg.conf
    echo "allow-loopback-pinentry" >> ~/.gnupg/gpg-agent.conf
    {
      echo \#\!/usr/bin/env bash
      echo "/usr/bin/gpg --batch --no-tty --pinentry-mode loopback --passphrase $GPG_PASS \"\$@\""
    } > ~/.gnupg/.gpgsign
    chmod 500 ~/.gnupg/.gpgsign
    gpg -q --batch --passphrase $GPG_PASS --import $GPG_PRIVATE_KEY_FILE
```
and also
```
    git config --global gpg.program ~/.gnupg/.gpgsign
```

Would be nice to get this in since it's almost 2y old now?!

vgrigoruk:
We ended up changing our setup to use `argocd` writeback method (in order to support manual rollback scenarios in ArgoCD, which turned to be problematic in case of using `git` write-back method). So this feature is not required for us anymore and I don't have a capacity to push this forward.

ForbiddenEra:
TBH I ended up just having a CI pipeline update the kustomization in my repo because of all the issues with Argo with overrides with branches and/or submodules - it just doesn't work (the override makes 'changes' in the repo, then when Argo subsequently tries to pull new submodules or switch branches, git says you have uncommitted changes and now you're screwed)

Kind of annoying in the end that I spent hours of frustration dealing with the aforementioned bug, not having GPG-ability and everything that I got the same result with like 3 lines of pipeline code :/

Not to diss on the project but, I feel like there's a lot of cases where the image updater would be used and perhaps should be part of Argo itself not an add-on. Smaller teams aren't always making helm charts for each internal service/app letalone specifically versioning them letalone wanting to bump a version everytime they commit something and want to review it deployed, all of those cases (and many more I'm sure) don't end up updating the manifests (especially when the image is targeting 'latest' or 'staging' tag or something) and thus Argo doesn't sync, obviously this is known since image updater exists but again I think this should be a 1st class thing, not an alpha add-on that seems to be collecting a bit of dust already :/

dlactin:
Hey @vgrigoruk I've added support for SSH signing keys and built upon the work you've done here. 

Do you mind signing off on this commit so I can rebase and add it to my PR?

PR #442, opened 2022-06-04
fix: Doc error --docker-server

hawkesn:
Small fix, the flag `--docker-registry` is not a valid flag but `--docker-server` is.
Referring to the official Kubernetes documentation here: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-by-providing-credentials-on-the-command-line


jannfis (approved):
Thanks and apologies for the late review!

LGTM.

PR #452, opened 2022-06-22
[docs] Use 'argocd' namespace in Installation docs

joaomlneto:
We start by applying the installation manifests in namespace "argocd".
However, later we restart the deployment in a separate "argocd-image-updater" namespace.


joaomlneto:
Nevermind; didn't realize all of that was part of the *Method 2* section.

PR #471, opened 2022-08-18
docs: Added warning about existing tag in semver

kostis-codefresh:
Signed-off-by: Kostis Kapelonis <kostis@codefresh.io>

When changing from latest to semver strategy, users must be careful and make sure that the existing
tag is already following semver

Otherwise trying to change from tag "rc1" to "1.0" or something similar doesn't work


jannfis (approved):
LGTM! Thanks and sorry for the late reaction

Issue #475, opened 2022-08-24
Write back git is slow


yuyuvn:
**Is your feature request related to a problem? Please describe.**
We using argocd image updater to sync our cluster, which will need to sync more than 100 namespace at a time.

Because of this lock, it will fetch and commit one by one and our concurrency won't have effect for speed this progress.
https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/argocd/update.go#L489

Each time argocd image updater run git fetch, it will take about 20~30s, some time it take 1 minute. 😇 

-> it took years to finish one check in our environment.

**Describe the solution you'd like**
The lock name should consider to include not just `repoURL` but `targetRevision`

**Describe alternatives you've considered**
Try git fetch with `--depth 1` may also help decrease fetch time for long history repo.
https://github.com/argoproj-labs/argocd-image-updater/blob/a03f31915d71ae39f1c9ac5d7d4d1501d11d158b/ext/git/client.go#L330

mubarak-j:
[This lock](https://github.com/argoproj-labs/argocd-image-updater/blob/49825ec4df76b3d2571515525ac71b83efd30acc/pkg/argocd/update.go#L489) is potentially a show-stopper for mono-repos because all applications share the same `app.Spec.Source.RepoURL`, image updater is locked into processing one application at a time. In my use case, with ~500 apps, image-updater takes 13 mins to cycle through all apps for git writeback operation. Related [slack thread](https://cloud-native.slack.com/archives/C0296T47CHY/p1680124993249549).

I ended up sharding the image-updater controller using `--match-application-label`, but even with sharding the git writeback cycle keeps increasing as new apps are added.

I suggest including `appName` in the lock name and if Git push fails e.g Git remote ref is ahead because another app in the same mono-repo was updated seconds/mins earlier, then simply handle the error by `'git pull ...') before pushing again`. I believe this will speed up the process significantly for mono-repo which I believe is a common use case with ArgoCD.

PR #485, opened 2022-10-19
Use --depth 1 when fetch

yuyuvn:
Issue: https://github.com/argoproj-labs/argocd-image-updater/issues/475

Using depth 1 when fetching will help speed up when fetching a very large history repo

### DIFF
--- a/ext/git/client.go
+++ b/ext/git/client.go
@@ -327,9 +327,9 @@ func (m *nativeGitClient) Fetch(revision string) error {
 
 	var err error
 	if revision != "" {
-		err = m.runCredentialedCmd("git", "fetch", "origin", revision, "--tags", "--force")
+		err = m.runCredentialedCmd("git", "fetch", "origin", revision, "--tags", "--force", "--depth", "1")
 	} else {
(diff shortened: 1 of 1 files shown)


jannfis:
Thank you for your PR, @yuyuvn, and apologies for the late reply.

Unfortunately, this PR modifies a component that we copy verbatim from Argo CD and then enrich with our own functions. We do not touch any of the existing functions, though. That means merging this PR would result in some potential regressions in the future.

I think the correct way would be to leverage the Argo CD repo server for any Git related operations. We are in the process of merging the Image Updater into Argo CD, and then we could make such a modification aligned with how Argo CD envisions it, too.

chengfang:
@yuyuvn thanks for the enhancement patch. Shallow fetch is now implemented in image-updater. For details see https://github.com/argoproj-labs/argocd-image-updater/blob/62b07d5771ce4bedd2c6f9dc98db5bd11f0b4848/ext/git/writer.go#L164 

https://github.com/argoproj-labs/argocd-image-updater/blob/62b07d5771ce4bedd2c6f9dc98db5bd11f0b4848/ext/git/client.go#L68

PR #507, opened 2022-12-22
Examples for semver and digest update strategy

dmouse:
Proposal examples to use semver and digest update strategy


mindovermiles262:
Already found this helpful. Thank you!

jannfis (approved):
LGTM, thank you for those! And apologies for coming back so late to this PR.

PR #520, opened 2023-01-16
docs(readme): add link for proposal to integrate with argo-cd

morey-tech:
Provide visibility to the proposal to integrate (graduate?) the project in the `argoproj` org and out of the lab.


jannfis (approved):
Thanks!

PR #534, opened 2023-02-16
Name strategy regexp in example need to be correctd 

iamraj007:
Name strategy among Update strategies, have its example have a wrong regexp. Need correction to match format mentioned in last line `YYYY-MM-DD`.  Current one is this format YYYY-MMDD (which is not what, doc wish to show)


jannfis (approved):
Great catch, thank you for improving the docs!

LGTM.

PR #538, opened 2023-02-28
Adds a new update strategy: manual

mvazquezc:
When running argocd-image-updater as part of a pipeline, it will be handy to be able to manually specify the image tag that you want to configure on your files.

Example:

Let's say we have a pipeline that builds our app, the resulting container image will get a tag that will match the commit id that generated that pipeline execution. Something like `example-image:7b7d2504`.

Other pipelines/people may be working with that image and generating different tags, so using something like `latest` strategy won't work.

I want the pipeline to update my Argo APP so as soon as the image is tagged and pushed the app gets deployed to my env. In the past, we used this tool[1]. 

In the pipeline, the last step will be updating the files in git with the new tag generated.

[1] https://github.com/gitops-tools/image-updater

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -45,6 +45,7 @@ type ImageUpdaterConfig struct {
 	GitCommitMail       string
 	GitCommitMessage    *template.Template
 	DisableKubeEvents   bool
+	ManualTagValue      string
 }
 
 // newRootCommand implements the root command of argocd-image-updater
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -33,6 +33,7 @@ func newRunCommand() *cobra.Command {
 	var warmUpCache bool = true
 	var commitMessagePath string
 	var commitMessageTpl string
+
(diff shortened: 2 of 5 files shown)


mvazquezc:
Hello, could someone review it?

mvazquezc:
> Apologies please for coming back so late to this PR.
> 
> I was wondering the use case for manual update, because you specify a tag on the command line. Do you use Image Updater from within your CI pipeline? If yes, what is the benefit of running Image Updater in the CI vs. just comitting to the GitOps repository where your manifests are?

@jannfis we could commit to our GitOps repository directly as you mentioned, but since image-updater already supports working with Git this just adds a new option to the tool. We may have pipelines that benefit from `latest` or `semver` strategies, if we could have this `manual` we will be using the same tooling/process for all of them rather than ones using image-updater and others using kustomize/git commands.

mvazquezc:
@jannfis what are the next steps to get this in? Thanks!

PR #541, opened 2023-03-20
Documentation: ArgoCD apps in any namespace incompatibility

seanmorton:
For those using [Applications in any namespace](https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace), the docs should mention that Image Updater is not compatible with any `Application` resources installed in namespaces other than `argocd`.

### DIFF
--- a/docs/configuration/applications.md
+++ b/docs/configuration/applications.md
@@ -22,6 +22,11 @@ only resources of kind `application.argoproj.io`, that is, your Argo CD
 `Application` resources. Annotations on other kinds of resources will have no
 effect and will not be considered.
 
+!!!warning
+    Argo CD Image Updater is not currently compatible with the
+    [Argo CD Applications in any namespace](https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace/)
+    feature. All `Application` resources must be declared in Argo CD control plane's namespace (which is usually argocd.)
+
 As explained earlier, your Argo CD applications must be of either `Kustomize`
 or `Helm` type. Other types of applications will be ignored.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -8,14 +8,18 @@ that are managed by
     Argo CD Image Updater is under active development.
 
     You are welcome to test it out on non-critical environments, and of
-    course to 
+    course to
     [contribute](./contributing/start.md) by many means.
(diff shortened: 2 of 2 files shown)


ElanHasson:
That is unfortunate. This is exactly what I'm doing and I just installed and deployed it :D

ElanHasson:
This is a dupe of #601 and #612 cover this.

JStacy86 (approved):
This should absolutely be added. It completely got me.

seanmorton:
Looks like most recent effort to add support for this was closed: https://github.com/argoproj-labs/argocd-image-updater/pull/672

Could we merge this warning so folks are at least aware? My change also adds a brief warning about this to the documentation root.

docs/configuration/applications.md
```diff
@@ -22,6 +22,12 @@ only resources of kind `application.argoproj.io`, that is, your Argo CD
 `Application` resources. Annotations on other kinds of resources will have no
 effect and will not be considered.
 
+!!!warning
+    Argo CD Image Updater is not currently compatible with the
+    [Argo CD Applications in any namespace](https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace/)
+    beta feature. All `Application` resources must be declared in the `argocd`
+    namespace.
```
chengfang, line 29: We should probably remove "beta" as this feature is no longer beta in Argo CD.  How about changing the ns to "Argo CD control plane's namespace (which is usually argocd)"?

chengfang (approved):

chengfang:
Thanks @seanmorton @ElanHasson @JStacy86 for the fix and review!

mcanevet:
Hasn't #687 added the support of Applications in any Namespace?

chengfang:
@mcanevet thanks for pointing that out. @seanmorton @ElanHasson @JStacy86 could you please verify if the current master branch works for you when using image-updater for apps in any namespace?

wd:
I can confirm the image-updater can work with apps in any namespace if you use the `APPLICATIONS_API` approach.
https://github.com/argoproj-labs/argocd-image-updater/issues/601#issuecomment-2249819819

ElanHasson:
Thanks folks!

Issue #562, opened 2023-05-05
multiple image in one applicatio, updater keeps looping the second image version between default and desired value


mattdecorniquet:
**Describe the bug**
We have an issue where we use 2 images in a single helm application, configured image-updater to monitor and update both those images and this doesn't work reliably.
Sometimes both images are updated, sometimes only the first in the list, and sometimes it updates both, then deletes the values for the second image in the .argocd-source file, then re-adds it, then redeletes it etc etc.

This is very similar and perhaps identical to #202, for which the fix still seems to be present in the code base.
It's also very similar to #365 and #170.

Our annotations look like this:

```  template:
    metadata:
      name: '{{`{{env}}`}}'
      annotations:
        argocd.argoproj.io/manifest-generate-paths: .:../../helm/ciam-app
        argocd-image-updater.argoproj.io/image-list: '{{`{{env}}`}}=gcr.io/xxxx/app, {{`{{env}}`}}-liquibase=gcr.io/xxxx/liquibase'
        argocd-image-updater.argoproj.io/{{`{{env}}`}}.helm.image-name: image.name
        argocd-image-updater.argoproj.io/{{`{{env}}`}}.helm.image-tag: image.tag
        argocd-image-updater.argoproj.io/{{`{{env}}`}}-liquibase.helm.image-name: imageLiquibase.name
        argocd-image-updater.argoproj.io/{{`{{env}}`}}-liquibase.helm.image-tag: imageLiquibase.tag
        argocd-image-updater.argoproj.io/git-branch: main
        argocd-image-updater.argoproj.io/write-back-method: git:secret:argocd/argocd-repo-creds-ssh-creds
        argocd-image-updater.argoproj.io/{{`{{env}}`}}.update-strategy: latest
        argocd-image-updater.argoproj.io/{{`{{env}}`}}.allow-tags: '{{`{{regexp}}`}}'
        argocd-image-updater.argoproj.io/{{`{{env}}`}}-liquibase.update-strategy: latest
        argocd-image-updater.argoproj.io/{{`{{env}}`}}-liquibase.allow-tags: '{{`{{regexp}}`}}'
```

Our helm chart specifies this:

```image:
    name: gcr.io/xxxx/app
    pullPolicy: IfNotPresent
    # Overrides the image tag whose default is the chart appVersion.
    tag: v1.0.1
imageLiquibase:
    name: gcr.io/xxxx/liquibase
    pullPolicy: IfNotPresent
[...]

PR #573, opened 2023-06-09
feat: Respect original parameter overrides with git write-back

ks-yim:
Modification:

Make image-updater respect the original parameter overrides written in `.argocd-source-<appName>.yaml`.

Motivation:

Currently, the original overridden parameters stored in `.argocd-source-<appName>.yaml`
are not respected when image-updater writes back its change to git.

Imagine that a user has originally defined `.argocd-source-<appName>.yaml` like below
```yaml
helm:
  parameters:
  - name: some.prop
    value: some-value
    forcestring: true
```
...and image-updater wants to write back the additional overrides below
- `foo.image.repository: foo`
- `foo.image.tag: 2.0.0`

The user may want the edited file to be like...
```yaml
helm:
  parameters:
  - name: some.prop
    value: some-value
    focestring: true
  - name: foo.image.repository
    value: foo
    forcestring: true
  - name: foo.image.tag
    value: 2.0.0
    forcestring: true
```
... but what's actually committed to git is the following and it deletes the original user-defined parameter overrides.
```yaml
helm:
  parameters:
  - name: foo.image.repository
    value: foo
    forcestring: true
  - name: foo.image.tag
    value: 2.0.0
    forcestring: true
```

And this image-updater behavior is a suspected cause of #562 and the likes
as image-updater ignores the tags for an image that it failed to fetch when
there are multiple updatable images defined in an `Application` and generate 
the parameter overrides without the ignored image. 
If only a part of image tags were successfully fetched from the image registry
due to unstable network, image-updater may write a broken parameter override
file to git.

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -3,6 +3,7 @@ package argocd
 import (
 	"context"
 	"fmt"
+	"golang.org/x/exp/slices"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -375,7 +376,7 @@ func setAppImage(app *v1alpha1.Application, img *image.ContainerImage) error {
 
 // marshalParamsOverride marshals the parameter overrides of a given application
 // into YAML bytes
-func marshalParamsOverride(app *v1alpha1.Application) ([]byte, error) {
+func marshalParamsOverride(app *v1alpha1.Application, originalData []byte) ([]byte, error) {
 	var override []byte
 	var err error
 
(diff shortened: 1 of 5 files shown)


pkg/argocd/update.go
```diff
@@ -385,21 +386,46 @@ func marshalParamsOverride(app *v1alpha1.Application) ([]byte, error) {
...
 			Helm: helmParameters{
 				Parameters: app.Spec.Source.Helm.Parameters,
 			},
 		}
+
+		if len(originalData) == 0 {
+			override, err = yaml.Marshal(newParams)
+			break
+		}
+		err := yaml.Unmarshal(originalData, &params)
```
jannfis, line 423:
```suggestion
		err = yaml.Unmarshal(originalData, &params)
```

jannfis (changes requested):
Nice one, thank you!

The linter complained, and it turns out your code shadows `err` in one place. Please fix it, otherwise LGTM :)

jannfis:
Oh, and please make sure the DCO passes by signing off your commits

ks-yim:
@jannfis
Thanks for taking your time to review this!

The issue described in the PR description has actually caused a disruption to our production(insignificant, though) workload so it would be very nice if this PR is shipped into the next release.

Resolved the variable shadowing issue and signed-off all my commits included in the PR.
I tried to make it pass `Spell checking` and `Run codegen` by merging the latest `master` but had no luck. 😅

jannfis:
Thanks for looking into it, @ks-yim!

I have just merged #590 , so if you rebase your PR, the codegen step should pass.

I'm aiming to release v0.13.0 after https://github.com/argoproj-labs/argocd-image-updater/pull/578 has been merged, however, it needs some attention from the submitter.

jannfis (approved):
LGTM. Going to merge it without rebase required, because the codegen error is negligible.

PR #576, opened 2023-06-21
Slice is a ref type, no need to use pointer of it here

suzaku:


jannfis:
Is that true? I always assumed that slices are two-fold, the slice's header is passed by value, its elements by reference. So once you start manipulating the slice itself (and not its elements) in a function, you need to pass the slice by reference (i.e. a pointer).

suzaku:
> Is that true? I always assumed that slices are two-fold, the slice's header is passed by value, its elements by reference. So once you start manipulating the slice itself (and not its elements) in a function, you need to pass the slice by reference (i.e. a pointer).

Yes, we'll need to use a pointer if we intend to update the slice itself, is that the case here with `Images`?

jannfis:
> is that the case here with Images

You tell me. But apparently, the unit tests are failing with this change.

PR #581, opened 2023-06-29
Documentation Error: Incorrect Usage of <image_name> in Annotations

geekloper:
I have noticed an error in [the documentation](https://argocd-image-updater.readthedocs.io/en/stable/configuration/images/#update-strategies) related to annotations. The documentation suggests using <image_name> instead of the correct placeholder <image_alias> when working with annotations ( Update strategies, Filtering tags, Ignoring certain tags, Image platforms & Specifying pull secrets ) . This mistake can potentially misguide users and cause confusion.

### DIFF
--- a/docs/configuration/images.md
+++ b/docs/configuration/images.md
@@ -126,7 +126,7 @@ You can define the update strategy for each image independently by setting the
 following annotation to an appropriate value:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


jannfis (approved):
LGTM, thanks!

PR #584, opened 2023-07-01
Fix: git zombie processes left behind after periodic image updater runs

dmavis:
As mentioned in https://github.com/argoproj-labs/argocd-image-updater/pull/583, during troubleshooting and testing for that issue, I noticed that zombie git processes were building up in the running container after each periodic image updater run. The Argo CD project had a similar issue: https://github.com/argoproj/argo-cd/issues/3611 and my fix uses the same solution

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ FROM alpine:latest
 
 RUN apk update && \
     apk upgrade && \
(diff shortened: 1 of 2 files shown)


jannfis (approved):
Great catch, thanks!

LGTM.

PR #600, opened 2023-08-03
fix: tighten securityContext to comply with restricted PSS

tsawada:
makes argocd-image-updater compatible with restricted Pod Security Standard


tsawada:
We've been running 0.12.2 with this securityContext for a few weeks and things seem to be okay. All other components in ArgoCD have the same securityContext since https://github.com/argoproj/argo-cd/pull/9765 , and I think it is preferable for argocd-image-updater to align with them.

@jannfis please let me know if I can do anything to help this getting merged. Thanks in advance!

Issue #371, opened 2022-02-01
Git commit fails when auto-sync is disabled in ArgoCD


janpieper:
**Describe the bug**

We set up an alert in Prometheus based on the `argocd_image_updater_images_errors_total` metric to be informed in case there is a problem with any image update. For some applications, especially for prod environments, we disable auto-sync in ArgoCD.

So in case someone pushes a new image to the docker registry, the ArgoCD Image Updater updates the `kustomization.yaml` to use the newly pushed image. That's fine. But if your application is configured to not auto-sync these changes, the ArgoCD Image Updater will try to update the `kustomization.yaml` over and over again. This fails, because it cannot commit the change to the `kustomization.yaml` because git does not detect any changes.

This problem causes the `argocd_image_updater_images_errors_total` to increase which then causes an alert in our alerting system although it is a false-positive in this case.

**To Reproduce**

1. Disable auto-sync in ArgoCD for an application that uses the ArgoCD Image Updater
2. Push a new image for that application to the docker registry
3. Do **NOT** sync the application manually!
4. Wait until the ArgoCD Image Updater runs the next time

**Expected behavior**

In theory, the ArgoCD Image Update should be fine with not updating the `kustomization.yaml`, but **only** in case the change has already been committed before.

**Additional context**

_None_

**Version**

0.11.0

**Logs**

```
time="2022-01-31T15:54:57Z" level=info msg="Starting image update cycle, considering 12 annotated application(s) for update"
time="2022-01-31T15:54:57Z" level=info msg="Setting new image to example.azurecr.io/git-example-frontend:v3.2.0" alias=frontend application=example-frontend image_name=git-example-frontend image_tag=v3.1.0 registry=example.azurecr.io
time="2022-01-31T15:54:57Z" level=info msg="Successfully updated image 'example.azurecr.io/git-example-frontend:v3.1.0' to 'example.azurecr.io/git-example-frontend:v3.2.0', but pending spec update (dry run=false)" alias=frontend application=example-frontend image_name=git-example-frontend image_tag=v3.1.0 registry=example.azurecr.io
time="2022-01-31T15:54:57Z" level=info msg="Committing 1 parameter update(s) for application example-frontend" application=example-frontend
time="2022-01-31T15:54:57Z" level=info msg="Starting configmap/secret informers"
time="2022-01-31T15:54:57Z" level=info msg="Configmap/secret informer synced"
[...]

PR #607, opened 2023-08-16
fix: Kustomize skip commit if nothing changed

Fs02:
Related: https://github.com/argoproj-labs/argocd-image-updater/issues/371

- If a change to manifest is not synced, image updater will keep trying to update the image and try to produce a commit without any changes, this is resulting in error log and metrics which is noisy
- This PR will skip commit for kustomization file when nothing is changed. this case can happen when the manifest in git is not synced

### DIFF
--- a/pkg/argocd/git.go
+++ b/pkg/argocd/git.go
@@ -329,40 +329,51 @@ func writeKustomization(app *v1alpha1.Application, wbc *WriteBackConfig, gitC gi
 		return err, false
 	}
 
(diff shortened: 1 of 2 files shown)


mfmsky (approved):

lukaszsurfer:
👋 It looks like a year has passed and the problem is still active, the merge is welcome 😄

chengfang:
@lukaszsurfer thanks for bringing that up. Will take a look.

chengfang:
I resolved the conflict, and can you all please take another look?

chengfang (approved):

PR #617, opened 2023-08-30
fix: support ocischema.DeserializedImageIndex in registry client

jessebye:
Copies the logic from `manifestlist.DeserializedManifestList` to also handle `ocischema.DeserializedImageIndex` (which ECR returns for certain types of images).


jessebye:
@jannfis if you have a moment to review, this resolves an issue with using ECR and ARM64 images.

jannfis:
Thank you @jessebye , appreciate that you are tackling this. I'm not sure if I'm good with all that duplicated code. 

Would you think it's possible to come up with code that's not as redundant?

jessebye:
@jannfis I was not happy with that either. I think it could be refactored a bit to eliminate the duplication. Also I think there is a bug, it was updating one parameter but not pinning the digest correctly - seems there may be another place I need to update for it to work. I'll take another look and let you know when it's ready for review!

jessebye:
botched the commits, going to start a fresh branch + PR for this

PR #651, opened 2023-12-08
fix: filter images by image-list annotation

### LINKED ISSUE #650: [BUG] Image should be filtered by image-list annotation
Argocd image updater crashes with the following message. 
I think the problem lies in the fact that the images discovered are not filtered by the annotation `argocd-image-updater.argoproj.io/image-list` .
In my case, I have an image injected by operator istio at runtime that should not be observed by argocd image updater.

Error log :

> time="2023-12-08T11:22:53Z" level=error msg="Could not update application spec: could not find an image-name annotation for image istio/proxyv2" application=loans-staging

Application annotations :

> metadata:
>   annotations:
>     argocd-image-updater.argoproj.io/image-list: >-
>       alias=registry.corp/application/image:latest

Application status :

> status:
>   summary:
>     images:
>       - 'docker.io/istio/proxyv2:1.19.3'
>       - 'registry.corp/application/image:latest'

The code lines that should be updated :

https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/argocd/argocd.go#L488:491

ziouf:
Try to fix #650

@jannfis


pkg/argocd/argocd.go
```diff
@@ -481,19 +481,32 @@ func SetKustomizeImage(app *v1alpha1.Application, newImage *image.ContainerImage
...
+	for _, i := range *list {
+		if i.ImageName == img.ImageName {
+			return true
+		}
+	}
+	return false
+}
+
 // GetImagesFromApplication returns the list of known images for the given application
 func GetImagesFromApplication(app *v1alpha1.Application) image.ContainerImageList {
```
pasha-codefresh, line 495: I am not really like such approach, because you change total sense of this function. I think it would be better to add validation [here](https://github.com/ziouf/argocd-image-updater/blob/14bdc8115a450d87812e4404952fa20431065057/pkg/argocd/update.go#L422). Either you need rename function and change comment

pasha-codefresh (changes requested):

PR #666, opened 2024-01-09
Preserve field order when updating kustomization.yaml file

### LINKED ISSUE #502: Retain or sort YAML ordering of images component when using Git write-back method

When the Image Updater updates kustomiztion.yaml file it does not sort the `image` component for new entries (nor retain existing formatting).

```
- name: ...
  digest: ...
```

vs (say via the use of `kustomize`)
```
  - digest: ...
    name: ...
```

I would like the ordering to be sorted, or to retain the existing formatting.

**Additional context**
This is useful for us because we perform `git diff` comparisons (golden-tests) as part of our CI on our rendered manifests. Because argocd-image-updater changes the ordering, we frequently hit failed pipelines.

jtdoepke:
This commit causes argocd-image-updater to preserve the order of fields in the `kustomization.yaml` file when updating the image(s). Doing so will minimize git diffs in the case that other tools also edit the `kustomization.yaml` file.

Fixes #502

### DIFF
--- a/pkg/argocd/git_test.go
+++ b/pkg/argocd/git_test.go
@@ -1,6 +1,7 @@
 package argocd
 
 import (
+	"os"
 	"testing"
 	"text/template"
 	"time"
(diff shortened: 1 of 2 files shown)


jtdoepke:
The `build_image` job failed, but I'm pretty sure that's not due to my changes.

pkg/argocd/git_test.go
```diff
@@ -215,3 +216,74 @@ images:
...
+
+	filter, err := imagesFilter(v1alpha1.KustomizeImages{"foo@sha23456"})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	tests := []struct {
+		name        string
+		content     []byte
+		wantContent []byte
```
chengfang, line 245: can we keep `content` and `wantContent` as string to avoid multiple conversions?

chengfang:
@jtdoepke can you also fix the DCO failure with sign-off?

chengfang (approved):

chengfang:
@jtdoepke thanks!

PR #671, opened 2024-01-22
dockerfile house keeping

mqxter:
resolving this error 
The system-wide python installation should be maintained using the system
92.69     package manager (apk) only.
92.69
92.69     If the package in question is not packaged already (and hence installable via
92.69     "apk add py3-somepackage"), please consider installing it inside a virtual
92.69     environment, e.g.:

Screenshot 2024-01-22 at 15 16 27

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,9 +14,7 @@ FROM alpine:latest
 
 RUN apk update && \
     apk upgrade && \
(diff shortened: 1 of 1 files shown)


chengfang:
Thanks for reporting the issue and the fix. It will be addressed in #700, as there are already some comments there. Nonetheless, your contribution is much appreciated!

Issue #601, opened 2023-08-05
Image Updater does not work with "App in any namespace" Argo configuration


renchap:
**Describe the bug**
[App in any namespace](https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace/) is a set of Argo feature allowing you to have `Application` resources in other namespaces than `argocd`.

This setup is not supported by Image Updater

**To Reproduce**
- Configure an Argo app living in another namespace according to the docs above
- Setup Image Updater annotations on it
- Run Image Updater and see that it does not pick the application

**Expected behavior**
Image Updater detects the Application in other namespaces than `argocd`

**Additional context**

As the Application was not detected using the `kubernetes` application API, I tried switching to `applications_api: argocd`.

I ensured that Argo could update the apps by updating the RBAC role as described in the official App in Any namespace doc.

In this mode, it correctly detects the Argo Application, updates it (see above), but then it fails to update the parameter in Argo, because it does not pass the namespace of the app, and Argo uses `argocd` by default when no namespace is specified (see logs below).

**Version**
0.12.2

**Logs**

Those logs are after configuring `applications_api: argocd`

`argocd-image-updater` logs:

```
time="2023-08-05T08:44:46Z" level=debug msg="found 1 from 1 tags eligible for consideration" image="rg.fr-par.scw.cloud/my-project/test-image:production"
time="2023-08-05T08:44:46Z" level=info msg="Setting new image to rg.fr-par.scw.cloud/my-project/test-image@sha256:d468cd177dcc4e0bedb94492b7e171c9478ad10486bb40f021caa02f9a2ab660" alias=test-image application=test-image image
_name=my-project/test-image image_tag=dummy registry=rg.fr-par.scw.cloud
time="2023-08-05T08:44:46Z" level=debug msg="target parameters: image-spec= image-name=image.name, image-tag=image.tag" application=test-image image=rg.fr-par.scw.cloud/my-project/test-image
time="2023-08-05T08:44:46Z" level=info msg="Successfully updated image 'rg.fr-par.scw.cloud/my-project/test-image@dummy' to 'rg.fr-par.scw.cloud/my-project/test-image@sha256:d468cd177dcc4e0bedb94492b7e171c9478ad10486bb40f021caa02f
9a2ab660', but pending spec update (dry run=false)" alias=test-image application=test-image image_name=my-project/test-image image_tag=dummy registry=rg.fr-par.scw.cloud
[...]

PR #672, opened 2024-01-26
Namespace support

rossigee:
Add --namespaced flag, that when true will only monitor the configured namespace for applications. If it is false, then monitor all namespaces for applications.

By default this is set to true to preserve the current behavior.

fixes https://github.com/argoproj-labs/argocd-image-updater/issues/601

(Picking this up from @iandelahorne - https://github.com/argoproj-labs/argocd-image-updater/pull/612#issuecomment-1910674396)


ElanHasson:
Thanks @rossigee  for picking this up.

strangnet:
Does this also solve the issue of wrong namespace when creating the ApplicationEvent?

I face the issue where image-updater throws an error in kube/kubernetes.go on line 128:

`Event "xxxx" is invalid: involvedObject.namespace: Invalid value "xxxx": does not match event.namespace`

I was thinking of testing to change the namespace used in the command:

```go
result, err := client.Clientset.CoreV1().Events(app.ObjectMeta.Namespace).Create(client.Context, &event, metav1.CreateOptions{})
```

to use the namespace of the found app instead of the `client.Namespace`.

fad3t:
Hello, any update on this? Thx!

ElanHasson:
I need this-- I am happy to work with someone to get this handled.

ElanHasson:
Hi @jannfis, any chance you could take a look and provide some feedback here?

rossigee:
Sorry guys, I'm going to have to abandon this PR.

The reason is that ArgoCD image updater doesn't appear to fit my use case, as the majority of the image tags I need it to update appear on Application manifests that wrap Helm releases.

Here's an example...

```yaml
kind: Application
metadata:
  name: myapp
  namespace: mynamespace
spec:
  project: myapp
  destination:
    server: "https://kubernetes.default.svc"
    namespace: mynamespace
  source:
    chart: myapp
    repoURL: registry.mydomain.lan/charts
    targetRevision: 1.2.3
    helm:
      releaseName: myapp
      valuesObject:
        replicaCount: 1
        image:
          repository: registry.mydomain.lan/mynamespace/myapp
          tag: "1.0-123-ce3e866"
```

As the `repoURL` is an OCI chart repository and not a git repo (as ArgoCD seems to assume!) it fails at the first step, as it can't `git clone` it !

As far as I can see there is no other way to manage the above manifest, so this tool is not going to work for me. I may have to try '[kobold](https://github.com/bluebrown/kobold)' instead, but even that [needs patching](https://github.com/bluebrown/kobold/issues/19) due to it not handling the split `repository` and `tag` style commonly found in Helm charts, as per the above example.

PR #586, opened 2023-07-04
feat: Allow ACR authentication using Azure CLI

xescab:
Install azure-cli in Docker image in order to use the `az acr login` command.

Can be used with Azure Managed Identities with the following script:

```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-image-updater-config
  namespace: argocd
data:
  log.level: debug
  registries.conf: |
    registries:
    - name: acrexample
      api_url: https://acrexample.azurecr.io/
      prefix: acrexample.azurecr.io
      ping: yes
      insecure: no
      credentials: ext:/app/scripts/acr-login.sh
      credsexpire: 10h
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-image-updater-config-acr
  namespace: argocd
data:
  acr-login.sh: |
    #!/bin/sh
    LOGIN=$(az login --identity)
    REGISTRY="acrexample"
    TOKEN=$(az acr login --name $REGISTRY --expose-token --output tsv --query accessToken)
    echo "00000000-0000-0000-0000-000000000000:$TOKEN"
```

Closes #550 and #473

Issue #550, opened 2023-04-03
Cannot pull images from Azure Container Registry


matteotumiati:
**Describe the bug**
I cannot pull images that are stored in Azure Container Registry.

**To Reproduce**
1. Install ArgoCD from scratch
2. Install argocd-image-updater (`kubectl apply -n argocd -f https://raw.githubusercontent.com/argoproj-labs/argocd-image-updater/stable/manifests/install.yaml`)
3. Create a Kubernetes secret with credentials for the ACR (access using admin):

```bash
kubectl create secret docker-registry acr \
    --namespace argocd \
    --docker-server=<container-registry-name>.azurecr.io \
    --docker-username=<container-registry-name> \
    --docker-password=<password>
```

4. Edit the ConfigMap argocd-image-updater-config (`kubectl edit configmap argocd-image-updater-config -n argocd`) to include the section for custom registries:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"v1","kind":"ConfigMap","metadata":{"annotations":{},"labels":{"app.kubernetes.io/name":"argocd-image-updater-config","app.kubernetes.io/part-of":"argocd-image-updater"},"name":"argocd-image-updater-config","namespace":"argocd"}}
  creationTimestamp: "2023-04-03T08:12:02Z"
  labels:
    app.kubernetes.io/name: argocd-image-updater-config
    app.kubernetes.io/part-of: argocd-image-updater
  name: argocd-image-updater-config
  namespace: argocd
  resourceVersion: "13173"
  uid: 99218d75-12cf-41cf-b9e4-198fce05d60a
data:
  registries.conf: |
    registries:
    - name: 'azure'
      prefix: <container-registry-name>.azurecr.io
      api_url: https://<container-registry-name>.azurecr.io/
      credentials: pullsecret:argocd/acr
      default: true
```

It is defined as `pullsecret`, because the secret created in step 3 has a field `.dockerconfigjson` in the `.data` section.

5. Create an empty Helm chart and an application to deploy a simple container from the container registry in Azure. It will fail with `Init:ImagePullBackOff`.

**Expected behavior**
The image can be pulled successfully.

**Additional context**
N/A

**Version**
Latest

**Logs**
N/A

joeferreirapremera:
have you tried to repo this with a kubelet identity that has Acrpull against the ACR?

[...]

Issue #473, opened 2022-08-22
Unable to authenticate to Azure ACR using repository tokens


gilles-gosuin:
**Describe the bug**

argocd-image-updater fails to retrieve tags when registry authentication is configured to use Azure ACR tokens.

**To Reproduce**

**Version 1: Using the token name and password as credentials**

That was my initial attempt:

1. Create an ACR repository (Premium SKU is required for tokens) 
2. Create a scope map, a token and generate token password(s) (https://docs.microsoft.com/en-us/azure/container-registry/container-registry-repository-scoped-permissions)
3. Configure argocd-image-updater to use the token name and password as credentials for the registry

When I realized it wouldn't work, I attempted it with Service Principal credentials (client id / secret) and it worked fine.

**Version 2: Using a credential script to generate an OAuth token using the Docker registry API**

Since I could successfully do a simple `docker login` with the token credentials, I assumed that argocd-image-updater did not implement the OAuth flow correctly, I embarked on trying to understand how the Docker Registry API works and I ended up creating the following credentials script (redacted elements have been replaced by meaningful variable names): 

```shell
#!/bin/sh
set -e
scope="repository:${repository}:pull,metadata_read"
token_credentials=$(echo -n "${token_name}:${token_password}" | base64 -w0)
response=$(wget --header "Authorization: Basic ${token_credentials}" -O - -q "https://${registry_hostname}/oauth2/token?service=${registry_hostname}&scope=${scope}")
token=$(echo -n "${response}" | sed -rn 's/\{\"access_token\"\:\"(.*)\"\}/\1/p')
echo "00000000-0000-0000-0000-000000000000:${token}"
```

**Expected behavior**
Both versions presented above should work fine, as they follow the Docker Registry API and reproducing them manually works fine as well.

**Additional context**

Logging in to docker using the token credentials directly works fine:

```
docker login ${registry_hostname} -u ${token_name} -p ${token_password}
```

So does listing the tags using a token generated by the registry's token endpoint (using enssentially the same script as the one shown above):

```
scope="repository:${repository}:pull,metadata_read"
token_credentials=$(echo -n "${token_name}:${token_password}" | base64 -w0)
response=$(wget --header "Authorization: Basic ${token_credentials}" -O - -q "https://${registry_hostname}/oauth2/token?service=${registry_hostname}&scope=${scope}")
token=$(echo -n "${response}" | sed -rn 's/\{\"access_token\"\:\"(.*)\"\}/\1/p')
[...]

PR #676, opened 2024-02-17
docs: add instructions to authenticate to Azure Container Registry with workload identity

etiennetremel:
Add instructions to authenticate to Azure Container Registry with workload identity.

Closes #586, #550 and #473


Pionerd:
I just tested this, but the credentials are not picked up (`argocd-image-updater test <private_image>` fails). How can I know the script is actually ran? I can see the script present, the `registries.conf` is OK, the `/var/run/secrets/azure/tokens/azure-identity-token` is set, what am I missing?

vepetkov:
> I just tested this, but the credentials are not picked up (`argocd-image-updater test <private_image>` fails). How can I know the script is actually ran? I can see the script present, the `registries.conf` is OK, the `/var/run/secrets/azure/tokens/azure-identity-token` is set, what am I missing?
> 
> Edit: Three observations:
> 
> * I think it makes sense to document that the `ConfigMap` should be mounted in such a way that the script becomes executable (e.g. `defaultMode: 493`).
> * The documentation should be clear in the `ACR_NAME` is expected to be including `azurecr.io`.
> * Although the script gives a good token when run manually (I can use it with `docker login` and `docker pull`), it still does not work when used in `argocd-image-updater test`. Please see: [Cannot pull images from Azure Container Registry #550 (comment)](https://github.com/argoproj-labs/argocd-image-updater/issues/550#issuecomment-2047839433)

I was able to test it successfully but I in addition to the 2 things you mentioned above about the `defaultMode` and the `ACR_NAME`, I also had to add the shebang ` #!/bin/sh` on top of the auth.sh file in the configmap. 
Without it the script would run manually but fail when the image updater tries to run it with the error `Could not set registry endpoint credentials: error executing /app/auth/auth.sh: fork/exec /app/auth/auth.sh: exec format error`.

To test it within the pod, I had to explicitly specify the path to the `registries.conf` file as it did not pick it up (
`argocd-image-updater --registries-conf-path /app/config/registries.conf test <private_image>`). 
However, it all works fine when deployed and the workload identity is used for get the available tags from ACR.

etiennetremel:
Thanks for testing out @vepetkov and @Pionerd, I was using the official helm chart which does include some of these settings by itself.

jribmartins (approved):

avo-sepp:
This works if you only have a single ACR, it relies on the ACR being name being hard coded into a single Environment variable $ACR_NAME. This is not a viable solution for the project as a whole. As it's pretty common to have more that one Container Registry at an Org, usually with the same vendor. 

We need in-house support for Workload Identities or Azure Container Registry push secrets.

cveld:
@etiennetremel where can I find the specific code that is responsible reacting upon the seemingly magic prefix `ext`? Maybe we can easily extend the logic and include commandline parameters to the script? In that case we can leverage one general token script with the container registry as the parameter. 

Although if a code change is required we possibly can better include the azure workload identity logic.

mturzynski-xelerit:
Worked for me. Thank you so much!

- Using single ACR
- Docs were clear enough

chengfang (approved):

jwhy89:
I used helm instead and made this PR to the helm chart instead of having to patch all those resources:
https://github.com/argoproj/argo-helm/pull/3319

You can also use env to look through multiple registries with the auth script. But it gets a little tricky.

PR #678, opened 2024-02-22
docs: Add ability to toggle dark-mode

kazysgurskas:


jannfis:
Nice! Thanks for this, and I'm really sorry for the long delay in reviewing this.

Would you be able to fix the DCO check by signing off your commit? I'd appreciate that a lot, and then I will merge this PR.

Fixing the DCO in this PR should be as simple as running

```
git commit --amend -s
```

in your local PR branch and then force-pushing.

PR #663, opened 2024-01-08
fix: git write back Helm value files

burnb:
Related: #650 
Fixes for #636 
Based on PR #651

- [does not consider image aliases](https://github.com/argoproj-labs/argocd-image-updater/pull/636#discussion_r1415418188)
- [does not support nested param names](https://github.com/argoproj-labs/argocd-image-updater/pull/636#issuecomment-1822711396)
- image name and tag settled in helm value file only for the last image in the annotation image list

PR #680, opened 2024-02-29
Create directory automatically

jnovick:
* When writing to a new file, create the directory for the file if it does not already exist
* This is useful if writing `helmvalues` to a new directory that is solely used for updating the image version. The ArgoCD Application can add a second source and reference the file with the `ignoreMissingValueFiles` flag turned on as an additional values file
* If the directory already exists, this command does nothing.

This will be most beneficial once https://github.com/argoproj-labs/argocd-image-updater/pull/663 is also merged.

### DIFF
--- a/pkg/argocd/git.go
+++ b/pkg/argocd/git.go
@@ -291,6 +291,12 @@ func writeOverrides(app *v1alpha1.Application, wbc *WriteBackConfig, gitC git.Cl
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


jnovick:
@jannfis Any hints on how to resolve the CI failures? They do not appear to be related to my actual PR. Thanks for any guidance that you can provide.

chengfang (approved):

PR #686, opened 2024-03-11
fix: filter images by image-list annotation

kizzie:
Signed-off-by: Cyril MARIN <marin.cyril@gmail.com>
Signed-off-by: Aleksandr Petrov <burnb83@gmail.com>


kizzie:
ignore this.... always fun when you're playing with forks

PR #687, opened 2024-03-12
fix: Pass AppNamespace to UpdateSpec api call

mwijngaard:
Adds support for applications in other namespace than default argocd namespace (when using the argocd api).


mwijngaard:
Any chance this could get merged?

chengfang (approved):
LGTM

PR #699, opened 2024-04-13
Updates to Dockerfile

aburan28:
Update the Dockerfile to install awscli using apk instead of pip3. This is because the build currently fails when trying to install using pip3


chengfang:
@aburan28 is this PR superseded by #700 ? If so, can you close this one? Thanks.

PR #700, opened 2024-04-13
fix dockerfile to pass builds

aburan28:


bebosudo:
I was also hitting the following error when running the `pip install --upgrade pip` step:
```
error: externally-managed-environment

× This environment is externally managed
╰─> 
    The system-wide python installation should be maintained using the system
    package manager (apk) only.
    
    If the package in question is not packaged already (and hence installable via
    "apk add py3-somepackage"), please consider installing it inside a virtual
    environment, e.g.:
    
    python3 -m venv /path/to/venv
    . /path/to/venv/bin/activate
    pip install mypackage
    
    To exit the virtual environment, run:
    
    deactivate
    
    The virtual environment is not deleted, and can be re-entered by re-sourcing
    the activate file.
    
    To automatically manage virtual environments, consider using pipx (from the
    pipx package).

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

Installing the `aws-cli` package from `apk` instead of using `pip` solved my problem too.
Oddly, the container size increased from 237MB to 271MB after doing that though, probably because the `apk` package pulls some more dependency in compared to `pip`.

chengfang:
@aburan28 why the version upgrade `github.com/miracl/conflate`? Is it required fixing the `error: externally-managed-environment` problem?

aburan28:
@chengfang reverted the change for miracl conflate. likely related to a unintended previous change when attemping updating go to 1.22

Ileriayo:
Just curious. 
The image currently has v1 of the `awscli` baked in, does this PR update it to v2?

image

chengfang:
Just fyi, the lint CI failures (not related to this PR) will need to be fixed (see issue #706 ) to meet the merge requirements. I'll look into it.

aburan28:
@chengfang updated

jannfis (approved):
LGTM, thanks!

Issue #706, opened 2024-05-03
golangci-lint context loading failed: no go files to analyze


chengfang:
golangci-lint failed in CI run with errors: "context loading failed: no go files to analyze"

Need to fix configurations of golangci-lint:
* `.github/workflows/ci-tests.yaml`
  * update golangci-lint action from v3 to v5
  * update the used golangci-lint version from 1.52.2 to 1.57.2
  * add an explicit `setup-go` step before running golangci-lint
* `.golangci.yml`
  * update the used golangci-lint version from 1.52.2 to 1.57.2
  * reconfig the deprecated `skip-files` and `skip-dirs`
  * remove `service` field

PR #707, opened 2024-05-03
Fix: golangci lint

0luwaj0x1:
This PR fixes [Issue #706](https://github.com/argoproj-labs/argocd-image-updater/issues/706)

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -59,10 +59,17 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-      - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v3
+      - name: Setup Golang
+        uses: actions/setup-go@v4
         with:
-          version: v1.52.2
+          go-version: '1.20'
+          cache: false
+        env:
+          GO111MODULE: off
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v5
+        with:
+          version: v1.57.2
           args: --timeout 5m
   test:
     name: Ensure unit tests are passing
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,9 +1,5 @@
 run:
   timeout: 2m
-  skip-files:
-    - ".*\\.pb\\.go"
-  skip-dirs:
-    - vendor/
 linters:
   enable:
     - vet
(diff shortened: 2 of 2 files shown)


.golangci.yml
```diff
@@ -15,5 +11,7 @@ linters:
 linters-settings:
   goimports:
     local-prefixes: github.com/argoproj-labs/argocd-image-updater
-service:
-  golangci-lint-version: 1.52.2
+issues:
+  exclude:
+    - ".*\\.pb\\.go"
+    - "vendor/.*"
```
chengfang, line 17: I think we should list `exclude-files` and `exclude-dirs` separately for clarity. In the above the vendor exclusion doesn't seem to include all subdirs.

```yaml
exclude-files:
  - ".*\\.pb\\.go"
exclude-dirs:
  - vendor/
```

.github/workflows/ci-tests.yaml
```diff
@@ -59,10 +59,12 @@ jobs:
...
         uses: actions/checkout@v3
+      - name: Setup Golang
+        uses: actions/setup-go@v3
       - name: Run golangci-lint
-        uses: golangci/golangci-lint-action@v3
+        uses: golangci/golangci-lint-action@v5
         with:
-          version: v1.52.2
+          version: v1.57.2
           args: --timeout 5m
```
chengfang, line 68: We're still seeing lint failures, though with a slightly better error message. Wondering if https://github.com/golangci/golangci-lint/issues/395#issuecomment-1690510780 could help?

Dockerfile
```diff
@@ -14,7 +14,9 @@ FROM alpine:latest
 
 RUN apk update && \
     apk upgrade && \
-    apk add ca-certificates git openssh-client python3 py3-pip tini && \
+    apk add --no-cache ca-certificates git openssh-client python3 py3-pip tini && \
+    python3 -m venv /venv && \
+    . /venv/bin/activate && \
```
chengfang, line 19: #700 already has a fix for the above problem in Dockerfile. I think we should treat this as a separate issue to be addressed by #700. If you agree, can you please undo this change in Dockerfile, and keep this PR for fixing lint? @0luwaj0x1

chengfang:
@0luwaj0x1 can you also squash all commits?

Dockerfile
0luwaj0x1: Agreed

chengfang (approved):

PR #710, opened 2024-05-11
Added support for ssh signed commits and completed gpg signed commit work

dlactin:
Building upon the work done in [PR 428](https://github.com/argoproj-labs/argocd-image-updater/pull/428) to add support for SSH signed commits. 

Updated Documentation to include instructions and examples for adding an SSH signing key along with links to repository provider documentation for setting up commit verification.  

Updated deployment to include volumes for SSH key secrets and the respective volume mounts.

### DIFF
--- a/docs/basics/update-methods.md
+++ b/docs/basics/update-methods.md
@@ -153,7 +153,8 @@ format. To create such a secret from an existing private key, you can use
 
 ```bash
 kubectl -n argocd-image-updater create secret generic git-creds \
-  --from-file=sshPrivateKey=~/.ssh/id_rsa
+  --from-file=sshPrivateKey=~/.ssh/id_rsa \
+  --from-file=sshPublicKey=~/.ssh/id_rsa.pub \
 ```
 
 ### <a name="method-git-repository"></a>Specifying a repository when using a Helm repository in repoURL
@@ -246,6 +247,52 @@ as the author. You can override the author using the
 `--git-commit-user` and `--git-commit-email` command line switches or set
 `git.user` and `git.email`
 in the `argocd-image-updater-config` ConfigMap.
+
+### <a name="method-git-commit-signing"></a>Enabling commit signature verification using an SSH or GPG key
+Commit signing requires the repository be accessed using HTTPS or SSH with a user account.
+Repositories accessed using a GitHub App can not be verified when using the git command line at this time.
+
+Each Git commit associated with an author's name and email address can be signed via a public SSH key or GPG key.
+Commit signing requires a bot account with a GPG or SSH key and the username and email address configured to match the bot account.
+
+Your preferred signing key must be associated with your bot account. See provider documentation for further details:
+* [GitHub](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)
+* [GitLab](https://docs.gitlab.com/ee/user/project/repository/signed_commits/)
+* [Bitbucket](https://confluence.atlassian.com/bitbucketserver/controlling-access-to-code-776639770.html)
+
+Commit Sign Off can be enabled by setting `git.commit-sign-off: "true"`
+
+**SSH:**
+
+Both private and public keys must be mounted and accessible on the `argocd-image-updater` pod.
+
+Set `git.commit-signing-key` `argocd-image-updater-config` ConfigMap to the path of your public key:
+
+```yaml
+data:
+  git.commit-sign-off: "true"
+  git.commit-signing-key: /app/.ssh/id_rsa.pub
+```
+
+The matching private key must be available in the same location.
+
+Create a new SSH secret or add the public key to your existing SSH secret:
+```bash
+kubectl -n argocd-image-updater create secret generic ssh-git-creds \
+  --from-file=sshPrivateKey=~/.ssh/id_rsa \
+  --from-file=sshPublicKey=~/.ssh/id_rsa.pub
+```
+
+**GPG:**
+
+The GPG private key must be installed and available in the `argocd-image-updater` pod.
+Set `git.commit-signing-key` in the `argocd-image-updater-config` ConfigMap to the GPG key ID you want to use:
+
+```yaml
+data:
+  git.commit-sign-off: "true"
+  git.commit-signing-key: 3AA5C34371567BD2
+```
 
 ### <a name="method-git-commit-message"></a>Changing the Git commit message
 
--- a/docs/basics/update-methods.md
+++ b/docs/basics/update-methods.md
@@ -153,8 +153,7 @@ format. To create such a secret from an existing private key, you can use
 
 ```bash
 kubectl -n argocd-image-updater create secret generic git-creds \
-  --from-file=sshPrivateKey=~/.ssh/id_rsa \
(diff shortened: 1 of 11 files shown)


jannfis:
Hey @dlactin, thanks for this PR. 

I was wondering what the value-add for this would be, because Argo CD would not be able to validate those commits signed by SSH just yet?

dlactin:
Hey @jannfis, in our case we require signed commits on our infrastructure repositories. So we are unable to have image updater commit to any of our protected branches without this change. 

Screenshot 2024-05-16 at 9 34 02 AM

jannfis:
Thanks for the clarification, @dlactin. The feature makes sense to me, then. Also, I guess with [Source Verification Policies](https://github.com/argoproj/argo-cd/pull/14964) hopefully coming to Argo CD, we could as well integrate verification of SSH signatures and Image Updater would already know how to do sign using SSH then.

+1 from me for this feature. Please give us some time to review.

ext/git/writer.go
```diff
@@ -116,3 +126,28 @@ func (m *nativeGitClient) Config(username string, email string) error {
...
+func (m *nativeGitClient) SigningConfig(signingkey string) error {
+	// Check if SiginingKey is a GPG key or Public SSH Key
+	keyCheck, err := regexp.MatchString(".*pub$", signingkey)
+	if err != nil {
+		return fmt.Errorf("could not validate Signing Key as GPG or Public SSH Key: %v", err)
+	}
+	if keyCheck {
+		// Setting the GPG format to ssh
+		log.Warnf("Setting GPG Format to SSH")
+		_, err = m.runCmd("config", "gpg.format", "ssh")
```
jannfis, line 140: Setting this via `git config` can easily lead to race conditions (i.e. when two processes calling the function at the same time with different signing requirements). Can you please explore how to set the signing config for the commit command on the command line? I think this function should be gone, and everything should be integrated into the `Commit` function.

jannfis (changes requested):
I got a couple of comments, PTAL.

docs/basics/update-methods.md
```diff
@@ -247,6 +248,52 @@ as the author. You can override the author using the
 `git.user` and `git.email`
 in the `argocd-image-updater-config` ConfigMap.
 
+### <a name="method-git-commit-signing"></a>Enabling commit signature verification using an SSH or GPG key
+Commit signing requires the repository be accessed using HTTPS or SSH with a user account.
+Repositories accessed using a GitHub App can not be verified when using the git command line at this time.
+
+Each Git commit associated with an author's name and email address can be signed via a public SSH key or GPG key.
+Commit signing requires a bot account with a GPG or SSH key and the username and email address configured to match the bot account.
```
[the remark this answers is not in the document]
dlactin: That's a good point; I will update the messaging here. 

This is a GitHub/GitLab branch protection requirement, as you noted.

ext/git/writer.go
```diff
@@ -32,7 +33,16 @@ func (m *nativeGitClient) Commit(pathSpec string, opts *CommitOptions) error {
 		args = append(args, "-a")
 	}
 	if opts.SigningKey != "" {
-		args = append(args, "-S", opts.SigningKey)
+		// Check if SiginingKey is a GPG key or Public SSH Key
+		keyCheck, err := regexp.MatchString(".*pub$", opts.SigningKey)
+		if err != nil {
+			return fmt.Errorf("could not validate Signing Key as GPG or Public SSH Key: %v", err)
+		}
```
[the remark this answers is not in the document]
dlactin: Sounds good, I'll add that today.

ext/git/writer.go
dlactin: Unfortunately, it doesn't look like this is configurable via the commit cli at the moment: 

https://git-scm.com/docs/git-commit 

I have a couple of thoughts: 

1. Move it out of the commit function and somewhere it is only set once
2. Use the default `gpg` signing format and provide some instructions to mount a `~/.gitconfig` file if folks want to use SSH signing instead.

jannfis: I think setting a config is a global option, and on the command line you could do something like:

```
git -c gpg.format=ssh commit ...
```

From `git(1)`:
```
       -c <name>=<value>
           Pass a configuration parameter to the command. The value given will override values from configuration files. The <name> is expected in the same format as listed by git
           config (subkeys separated by dots).

           Note that omitting the = in git -c foo.bar ...  is allowed and sets foo.bar to the boolean true value (just like [foo]bar would in a config file). Including the equals but
           with an empty value (like git -c foo.bar= ...) sets foo.bar to the empty string which git config --type=bool will convert to false.

```

dlactin: Nice find, that works great! 

I will update the PR with that change.

dlactin: All right, let me know what you think about this change. I tested it this afternoon, and things are working as expected.

jannfis:
Just as a heads up, this is good to go. Waiting for https://github.com/argoproj-labs/argocd-image-updater/pull/737 to be merged, then we'd need to rebase this one on top of the Git client changes. It shouldn't be too much work though.

jannfis:
@dlactin Thank you for your patience! #737 has been merged, and as expected left a couple of conflicts for this PR. I can support you to resolve them, or let you do it on your own. Please let me know what you prefer.

dlactin:
> @dlactin Thank you for your patience! #737 has been merged, and as expected left a couple of conflicts for this PR. I can support you to resolve them, or let you do it on your own. Please let me know what you prefer.

Hey, not a problem! 

I won't have time to finish this up until sometime next week so feel free to resolve them if you'd like! Otherwise I'm happy to pick this back up later.

manifests/base/deployment/argocd-image-updater-deployment.yaml
```diff
@@ -77,6 +77,24 @@ spec:
...
+        - name: GIT_COMMIT_SIGNING_KEY
+          valueFrom:
+            configMapKeyRef:
+              key: git.commit-signing-key
+              name: argocd-image-updater-config
+              optional: true
+        - name: GIT_COMMIT_SIGNING_METHOD
+          valueFrom:
+            configMapKeyRef:
+              key: git.commit-signing-key
```
chengfang, line 89: should this be git.commit-signing-method?

manifests/install.yaml
```diff
@@ -158,6 +158,24 @@ spec:
...
+        - name: GIT_COMMIT_SIGNING_KEY
+          valueFrom:
+            configMapKeyRef:
+              key: git.commit-signing-key
+              name: argocd-image-updater-config
+              optional: true
+        - name: GIT_COMMIT_SIGNING_METHOD
+          valueFrom:
+            configMapKeyRef:
+              key: git.commit-signing-key
```
chengfang, line 170: same as above. should this be git.commit-signing-method?

manifests/base/deployment/argocd-image-updater-deployment.yaml
dlactin: yes, just pushed an update!

chengfang (approved):

jannfis (approved):
Awesome, thank you so much @dlactin, and also thank you @chengfang for the additional review!

LGTM.

PR #721, opened 2024-05-25
fix: downgrade alpine version, while aws cli is not available in 3.20

pasha-codefresh:
Related to: https://github.com/alpinelinux/docker-alpine/issues/396


jannfis (approved):
LGTM

PR #636, opened 2023-10-27
Feature/wbc multisource support

askhari:
It relates to PR #631 and includes the same changes as it's based on that code.

Also, this relates to the following issues:

- #551 
- #558 
- #513 

This PR provides multisource support for git write back configurations on Helm application types.
Also it adds a new feature to provide a helm values file for the git write back. This makes it easier to keep value files up to date.
In order to use the Helm values feature, you'll need to provide a "helmvalues" prefix in the write-back-target header. You can take a look to the write-back-target documentation in this code to know how to configure it.

There are test for all the new code, they cover the most basic cases.
I tried to keep the same coding pattern and built the least invasive changes in the code flow.

Please give it a try and let's see if it can be merged. Any suggestions are more than welcome.

Issue #558, opened 2023-04-24
Multiple helm sources: skipping app of type '' because it's not of supported source type


admssa:
**Describe the bug**
The problem is that when you store helm values files in a separate repo you use `spes.sources` instead of `spes.source` in your Application manifests. Thus instead of `sourceType`  string in application status we will have `sourceTypes` array. 

```
status:
  sourceTypes:
  - Helm
```
Unfortunately argocd-image-updater check for `sourceType` only:
https://github.com/argoproj-labs/argocd-image-updater/blob/49825ec4df76b3d2571515525ac71b83efd30acc/pkg/argocd/argocd.go#L520-L533

**To Reproduce**
Replace values surrounded by << >> and create Application with 2 sources and follow the logs:

```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: test
  namespace: argocd
  annotations:
    argocd-image-updater.argoproj.io/write-back-method: git
    argocd-image-updater.argoproj.io/image-list: test=<<REPO_URL/test>>
    argocd-image-updater.argoproj.io/test.force-update: "true"
    argocd-image-updater.argoproj.io/test.helm.image-tag: deployment.image.tag
    argocd-image-updater.argoproj.io/test.helm.image-name: deployment.image.repository
    argocd-image-updater.argoproj.io/test.update-strategy: latest
    argocd-image-updater.argoproj.io/test.allow-tags: regexp:(tested-dev-)
spec:
  project: default
  sources:
    - repoURL: <<rREPO_URL>>
      chart: test
      targetRevision: 0.1.0
      helm:
        releaseName: test
        valueFiles:
          - $values/dev/values/test.yaml
    - repoURL: "git@github.com:<<YOUR_ORG>>/gitops-backend.git"
      targetRevision: HEAD
      ref: values
  destination:
    server: "https://kubernetes.default.svc"
    namespace: dev
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
      allowEmpty: false
    syncOptions:
      - Validate=true
      - CreateNamespace=true
      - PrunePropagationPolicy=foreground
      - PruneLast=true

```

**Version**
0.12.2

**Logs**
```
time="2023-04-24T08:48:37Z" level=info msg="Processing results: applications=0 images_considered=0 images_skipped=0 images_updated=0 errors=0"
time="2023-04-24T08:50:38Z" level=warning msg="skipping app 'test' of type '' because it's not of supported source type" application=test
[...]

PR #725, opened 2024-05-27
fix: set default ImageAlias for Helm app type using helmvalues git write-back method

askhari:
It relates to PR #636 and includes a fix to set a default ImageAlias name on container objects in order to retrieve application's annotation values properly.

Also, this relates to the last comments from [@zagr0](https://github.com/zagr0) on issue #558.

This PR fixes a problem retrieving image-name and image-tag annotations when we try to update a Helm type application using the "helmvalues" way on the "write-back-target" annotation.
It includes a new GetImagesAndAliasesFromApplication function which sets the ImageAlias attribute for the containerImage object. This solves the problem retrieving annotations which uses the container alias name.

As with PR #636 , I tried to keep the same coding pattern and built the least invasive changes in the code flow. That's the reason of using a new function.
Also, there are test for all the new code, they cover the most basic cases.

Please give it a try and let's see if it can be merged. Any suggestions are more than welcome.

@jannfis , @pasha-codefresh , I'm not sure if this is the way to open a fix for release-0.13. What I did was the following:
1. Download the last changes.
2. Changed to release-0.13 branch.
3. Create a new branch fix/set-default-image-alias-with-helmvalues .
4. Create this PR to merge into release-0.13 .

Please, let me know if I need to work on this on any other way.


jannfis:
Hi @askhari, please submit all PRs against the default branch ("`master`"). All changes have to first go into the default branch, and maintainers will cherry-pick any changes to the release branches as required.

Thank you!

askhari:
@jannfis , thank you for the info.

Then, I'll make some ammendments and try to fix the integration test that is failing. this is due I branched from release-0.13 and it has that version in the kustomization.yaml and install.yml files.

askhari:
@jannfis , I think it's all done.
I changed the base branch to be master and also changed the version for the docker image and pass the integration tests.
Again, thank you for the quick response :)

[@NissesSenap](https://github.com/NissesSenap) , thank you for testing it so quickly.

pkg/argocd/update.go
```diff
@@ -416,15 +416,17 @@ func marshalParamsOverride(app *v1alpha1.Application, originalData []byte) ([]by
 		}
 
 		if strings.HasPrefix(app.Annotations[common.WriteBackTargetAnnotation], common.HelmPrefix) {
-			images := GetImagesFromApplication(app)
+			images := GetImagesAndAliasesFromApplication(app)
 
 			for _, c := range images {
-				helmAnnotationParamName, helmAnnotationParamVersion := getHelmParamNamesFromAnnotation(app.Annotations, c.ImageName)
```
pasha-codefresh, line 422: why you need to change this logic? why just not replace it with c.ImageAlias if it is exist ?

pasha-codefresh (changes requested):

VERSION
```diff
@@ -1 +1 @@
-99.9.9
+0.13.0
```
pasha-codefresh, line 1: Revert this change, please

pasha-codefresh (changes requested):

pkg/argocd/update.go
askhari: Thank you for pointing it out :)

There are mainly two reasons for this decision:

1. The code of _getHelmParamNamesFromAnnotation_ it's only used here and anywhere else in the code. I used it at first because it was convenient at the time, but it seems that it may be deprecated if it has no use anymore (I was going to suggest that in another issue later on).
2. Using _GetParameterHelmImageName_ and _GetParameterHelmImageTag_ seems a better option because this two functions uses _normalizedSymbolicName_ function internally, which normalizes the _ImageAlias_ enabling its use in annotations properly because it replaces the "/" character for "_".

If in the end it should be better to leave it as before, just let me know and I'll make revert the changes to use _c.ImageAlias_.

VERSION
askhari: Done! I already reverted it and pushed the branch.
I'll wait to see if this change dissapears from the diff and then mark the conversation as resolved, would that be right?

pkg/argocd/update.go
pasha-codefresh: I just concerned that you actually changing logic, because inside getHelmParamNamesFromAnnotation here more conditions. I would not change it for now. Also this function is well covered with tests, in case if you are going to use  new logic, you have to cover it with unit tests

askhari: Yup, you are absolutely right and I should think thrice before changing things.

I'll change it back to the previous statement changing the _c.ImageName_ for _c.normalizedSymbolicName()_ which will work properly with the annotation names because it already replaces the "/" characters for "_". I'll check if the logic works that way and push the changes back.

Thank you for noticing and explaining it :)

askhari: In order to make the _normalizedSymbolicName_ public to the _update_ package, I'll need to uppercase the first letter and change its references in other parts of the code. Would that be alright?

I should use that function in order to avoid that images names as _"super/app"_ will be used when referencing an annotation name and use "super_app" instead. This will avoid errors referencing annotations with malformed names. I'll give it a try and let me know what do you think.

pasha-codefresh: What do you think about idea, pass ContainerImage to getHelmParamNamesFromAnnotation, and just to same refactoring as you did in your PR, but in such case here still will be ability to cover this function with tests

askhari: I'm not sure if the changes I just pushed are what you meant before.

With this changes, there is no change in logic anywhere and all the current test should cover the same lines. Maybe this solutions works for you?

pasha-codefresh: Do you mind to give me write permissions to you fork? want to do small enhancement

askhari: Sure! I just send the collaboration invitation :)

pasha-codefresh: @askhari i have pushed a change

askhari: @pasha-codefresh , thank you very much for your help. I really appreciate it and with those changes I now understand what you meant by passing the container image. Thank you!

Do you need me to work on some more changes from my side for this PR? I think that with the changes you made, the problem is already solved.

pasha-codefresh: @askhari maybe you can help with manual test?

askhari: Sure! I'll run it on a cluster where we have a test application just for this purpose and will paste here the results to show the validation.

pasha-codefresh: Fine for me

chengfang (approved):

jannfis (approved):
LGTM, thanks @askhari

pasha-codefresh (approved):
LGTM, thank you

pkg/argocd/update.go
askhari: I see that this is already merged. Thank you for your help @pasha-codefresh .

Still, I attach a test I made to validate that the code is retrieving the information properly when using aliases.
Below there is a screenshot which uses an application definition with a docker image called _labduck_ and using an alias _labbat_ to validate that it is retrieving the values properly from the annotations.

Captura de pantalla 2024-06-05 a las 17 59 27

Again thank you for your help to solve this issue.

Cheers!

askhari:
Thank you very much @pasha-codefresh for your help and also to @jannfis and @chengfang for the approvals.

We usually delete branches after merging in our projects to keep it clean, so I'll proceed to delete it now.

Issue #744, opened 2024-06-13
Git write-back-target helmvalues does not replace tag and repo in the image section


zagr0:
**Describe the bug**
Image updater not able to find `tag` and `repository` within `image` hierarchy to replace values, instead it adds new entries `image.tag` and `image.repository`.

Helm `values.yaml` file used:
```
image:
  repository: my.repo.domain/common/app
  tag: 0.0.6-SNAPSHOT_3ced321a
```

The commited result by image updater in git:
```
image:
  repository: my.repo.domain/common/app
  tag: 0.0.6-SNAPSHOT_3ced321a
image.repository: my.repo.domain/common/app
image.tag: 0.0.5_b579c34b
```

**To Reproduce**
Configure application with annotations to update image tag with helmvales:
```
  annotations:
    argocd-image-updater.argoproj.io/image-list: main=my.repo.doamin/common/app
    argocd-image-updater.argoproj.io/main.pull-secret: pullsecret:argocd/registry
    argocd-image-updater.argoproj.io/main.helm.image-name: image.repository
    argocd-image-updater.argoproj.io/main.helm.image-tag: image.tag
    argocd-image-updater.argoproj.io/main.update-strategy: newest-build
    argocd-image-updater.argoproj.io/main.allow-tags: regexp:^[0-9]+.[0-9]+.[0-9+]_[0-9a-f]{8}$
    argocd-image-updater.argoproj.io/write-back-method: git:secret:argocd/git-creds-ssh
    argocd-image-updater.argoproj.io/write-back-target: helmvalues:./helm/app/values.yaml
    argocd-image-updater.argoproj.io/git-repository: git@my.repo.domain:common/deploy.git
    argocd-image-updater.argoproj.io/git-branch: master
``` 
with the `./helm/app/values.yaml` file content from example above.

**Expected behavior**
We expect to have updated tag and repo inside image section in `values.yaml` like:
```
image:
  repository: my.repo.domain/common/app
  tag: 0.0.5_b579c34b
```

**Version**
This is latest build from master, because we need the latest fixes for helmvalues git write-back

askhari:
Thank you @zagr0 for pointing this out.

As you may see in the comments, I opened #748 to fix the issue. I test it in my own environments, but please let me know if you may test it as well.

Let's see if it can be reviewed soon.

Cheers!

askhari:
Hi @zagr0,

Thank you for your feedback :).

[...]

PR #748, opened 2024-06-16
Fix/helm values wbc

askhari:
This PR relates to the previous PRs, issues and comments:
- Issue: #744 
- PR: #636 
- [Comment](https://github.com/argoproj-labs/argocd-image-updater/issues/558#issuecomment-2165248263) from @zagr0 on #558

These changes solve the following issues:
1. _does not search into the YAML hierarchy (#744 )_:  now this feature will search for the key name on the root node of the YAML file. If it doesn't exist, then it will navigate the YAML hierarchy to look for the right key. If it does not find the key, then it will fail.
2. _Currently it messes up the order of the keys in the YAML_:  with these changes, we got rid of the Conflate library and use a yaml.MapSlice object instead. This way we can preserve the order and avoid messing up the YAML order when writing back to the repository.

To preserve retrocompatibility in case someone already use dots in the YAML keys, we first look for the complete key with dots. Then, if it's not found we go through the YAML hierarchy to find the key.

For example, using an annotation like this: _argocd-image-updater.argoproj.io/labcat.helm.image-tag: rollout.image.tag_; first we'll try to find a key named "rollout.image.tag" in the YAML root node. And if we do not find it, then it will search the YAML hierarchy searching the following structure:
```
rollout:
  image:
    tag: <your preferred value>
```

There are new tests for these code.

Please, comments or suggestions are more than welcome.

Cheers!


pkg/argocd/update_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/argoproj-labs/argocd-image-updater/pkg/tag"
 	"github.com/argoproj-labs/argocd-image-updater/test/fake"
 	"github.com/argoproj-labs/argocd-image-updater/test/fixture"
+	"gopkg.in/yaml.v2"
```
chengfang, line 23: To fix the lint error, looks like the above line `gopkg.in/yaml.v2` should be in its own group around line 11.

askhari: @chengfang , thank you for your help. I already fixed it and pushed the change.

I'll mark the conversation as "resolved". Please, let me know if you find any further issue.

go.mod
```diff
@@ -38,20 +37,6 @@ require (
...
-	github.com/google/s2a-go v0.1.4 // indirect
-	github.com/googleapis/enterprise-certificate-proxy v0.2.5 // indirect
-	github.com/googleapis/gax-go/v2 v2.12.0 // indirect
-	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
-	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
-	github.com/xeipuuv/gojsonschema v1.2.0 // indirect
-	go.opencensus.io v0.24.0 // indirect
-	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
-	google.golang.org/api v0.132.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20230822172742-b8732ec3820d // indirect
```
chengfang, line 40: why are there many removed dependencies in go.mod and go.sum? Are they all indirect dependencies solely from `conflate`?

pkg/argocd/update.go
```diff
@@ -516,6 +523,57 @@ func mergeKustomizeOverride(t *kustomizeOverride, o *kustomizeOverride) {
...
+
+	for i, k := range keys {
+		found := false
+		for j, item := range current {
+			if item.Key == k {
+				if i == len(keys)-1 {
+					// If we're at the final key, set the value
+					current[j].Value = value
+					found = true
+					break
```
chengfang, line 555: Can we just return here since we've set the value and nothing else needs to be done?

alexismaior:
I tested locally with a similar configuration and it worked like a charm!

chengfang (approved):

chengfang:
Thanks @askhari  for the fix, and thanks @alexismaior for verifying it!

seagyn:
Thanks for this fix. We've just run into the issue, is there an expected ETA for a new release of image updater or an image we can use in the mean time?

pasha-codefresh:
I am going to create release on Monday

seagyn:
> I am going to create release on Monday

Appreciate it!

PR #750, opened 2024-06-19
Fix issue regarding volume added in #710

subigre:
Fix issue with volume `ssh-git-creds` (introduced in #710) that prevents to install `master` branch

`optional: true` was missing an indent


jannfis (approved):
Good catch, thank you!

LGTM.

PR #753, opened 2024-06-21
fix: fixed path for ssh signing key, error mounting in another volume

dlactin:
Moved ssh key mount for signing key into a new `ssh-keys` directory. This is to avoid trying to mount in an existing .ssh directory.

### DIFF
--- a/docs/basics/update-methods.md
+++ b/docs/basics/update-methods.md
@@ -275,7 +275,7 @@ Set `git.commit-signing-key` `argocd-image-updater-config` ConfigMap to the path
 ```yaml
 data:
   git.commit-sign-off: "true"
-  git.commit-signing-key: /app/.ssh/id_rsa
+  git.commit-signing-key: /app/ssh-keys/id_rsa
   git.commit-signing-method: "ssh"
 ```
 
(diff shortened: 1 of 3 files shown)


docs/basics/update-methods.md
```diff
@@ -275,7 +275,7 @@ Set `git.commit-signing-key` `argocd-image-updater-config` ConfigMap to the path
 ```yaml
 data:
   git.commit-sign-off: "true"
-  git.commit-signing-key: /app/.ssh/id_rsa
+  git.commit-signing-key: /app/ssh-keys/id_rsa
```
chengfang, line 278: `/app/ssh-keys/id_rsa` appears to be a directory. Should we point to the file corresponding to the mount subPath? How about shorten the path to something like `/app/ssh-keys`?

dlactin: It isn't shown well here, but in the `volumeMounts` block we are mounting a subPath from the secret. So this will be a file named `id_rsa` within the `/app/ssh-keys/` directory: 

ex: 

```
        volumeMounts:
        - mountPath: /app/ssh-keys/id_rsa
          name: ssh-signing-key
          readOnly: true
          subPath: sshPrivateKey
```

chengfang (approved):

chengfang:
@dlactin thanks for the fix!

PR #763, opened 2024-07-01
feat: Support other namespaces than argocd in AIU

sribiere-jellysmack:
When trying to update an application in another namespace than argocd, AIU should include namespace in logs, and in writeback file.

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -623,7 +623,7 @@ func getWriteBackConfig(app *v1alpha1.Application, kubeClient *kube.KubernetesCl
 
 func parseDefaultTarget(appNamespace string, appName string, path string) string {
 
-	if ((appNamespace == "argocd") ||  (appNamespace == "")) {
+	if (appNamespace == "argocd") || (appNamespace == "") {
 		defaultTargetFile := fmt.Sprintf(common.DefaultTargetFilePatternWithoutNamespace, appName)
 		return filepath.Join(path, defaultTargetFile)
 	} else {
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -584,7 +584,7 @@ func getWriteBackConfig(app *v1alpha1.Application, kubeClient *kube.KubernetesCl
 	// Default write-back is to use Argo CD API
 	wbc.Method = WriteBackApplication
 	wbc.ArgoClient = argoClient
-	wbc.Target = parseDefaultTarget(app.GetNamespace(), app.Name, getApplicationSource(app).Path)
+	wbc.Target = parseDefaultTarget(app.GetNamespace(), app.Name, getApplicationSource(app).Path, kubeClient.Namespace)
 
 	// If we have no update method, just return our default
 	method, ok := app.Annotations[common.WriteBackMethodAnnotation]
@@ -621,9 +621,9 @@ func getWriteBackConfig(app *v1alpha1.Application, kubeClient *kube.KubernetesCl
 	return wbc, nil
 }
 
-func parseDefaultTarget(appNamespace string, appName string, path string) string {
+func parseDefaultTarget(appNamespace string, appName string, path string, argoCDNameSpace string) string {
 
-	if (appNamespace == "argocd") || (appNamespace == "") {
+	if (appNamespace == argoCDNameSpace) || (appNamespace == "") {
 		defaultTargetFile := fmt.Sprintf(common.DefaultTargetFilePatternWithoutNamespace, appName)
 		return filepath.Join(path, defaultTargetFile)
 	} else {
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -621,15 +621,15 @@ func getWriteBackConfig(app *v1alpha1.Application, kubeClient *kube.KubernetesCl
 	return wbc, nil
 }
 
-func parseDefaultTarget(appNamespace string,appName string, path string) string {
+func parseDefaultTarget(appNamespace string, appName string, path string) string {
 
-    if (appNamespace == "argocd") {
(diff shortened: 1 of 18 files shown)


cmd/ask_pass_test.go
```diff
@@ -56,6 +56,7 @@ func NewTestCommand() *cobra.Command {
 			fmt.Fprintf(c.ErrOrStderr(), "%s is not set\n", git.ASKPASS_NONCE_ENV)
 			return
 		}
+		//nolint
```
chengfang, line 59: Is this required for this issue? It seems lint has been running fine without this, and we can probably revert the above line.

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/argoproj-labs/argocd-image-updater
 
-go 1.21
+go 1.22
 
-toolchain go1.22.3
+toolchain go1.22.4
```
chengfang, line 6: Upgrading go version is a major task on its own. I would prefer to do it as a separate issue later when the time is right.

pkg/argocd/update_test.go
```diff
@@ -1079,7 +1079,8 @@ kustomize:
 `
 		app := v1alpha1.Application{
 			ObjectMeta: v1.ObjectMeta{
-				Name: "testapp",
+				Name:      "testapp",
+				Namespace: "argocd",
```
chengfang, line 1083: Would the test still pass without adding Namespace field, since the behavior should remain the same when the application is in the argocd control plane's namespace?  

If so, we can probably leave some test cases unchanged, some adding "argocd" namespace, and some adding "other" ns, just to cover more cases.

pkg/common/constants.go
```diff
@@ -57,7 +57,8 @@ const (
 )
 
 // DefaultTargetFilePattern configurations related to the write-back functionality
-const DefaultTargetFilePattern = ".argocd-source-%s.yaml"
+const DefaultTargetFilePattern = ".argocd-source-%s_%s.yaml"
```
chengfang, line 60: Would it be easier to work with if we use the same separator "-", instead of "_"?

chengfang (commented):
Can you fix the DCO check? Fixing the DCO in this PR should be as simple as running

git commit --amend -s
in your local PR branch and then force-pushing.

cmd/ask_pass_test.go
sribiere-jellysmack: I had problems without this : SA1019: grpc.DialContext is deprecated: use NewClient instead.
Probably linked with the go upgrade

pkg/common/constants.go
sribiere-jellysmack: No : ArgoCD would not resync the application if we do not follow this pattern

pkg/argocd/update_test.go
sribiere-jellysmack: OK I'm working on this

pkg/common/constants.go
jannfis: This seems to be a majorly breaking change

pkg/argocd/update.go
```diff
@@ -621,10 +621,15 @@ func getWriteBackConfig(app *v1alpha1.Application, kubeClient *kube.KubernetesCl
 	return wbc, nil
 }
 
-func parseDefaultTarget(appName string, path string) string {
-	defaultTargetFile := fmt.Sprintf(common.DefaultTargetFilePattern, appName)
+func parseDefaultTarget(appNamespace string, appName string, path string) string {
 
-	return filepath.Join(path, defaultTargetFile)
+	if (appNamespace == "argocd") || (appNamespace == "") {
```
jannfis, line 626: `argocd` should not be hard coded here, because Argo CD is regularly installed into different namespaces.

pkg/common/constants.go
sribiere-jellysmack: I'm not sure to understand why it is a breaking change?

jannfis: Sorry, I have seen for the condition to decide which file name to actually use/write to later on.

pkg/argocd/update.go
sribiere-jellysmack: So we should add a parameter to parseDefaultTarget so we can compare with the actual namespace where ArgoCD is deployed?

jannfis: We need to use the namespace as specified here: https://github.com/argoproj-labs/argocd-image-updater/blob/ae3eec3ff69487d13da770abd9167c61ac957931/cmd/run.go#L237

sribiere-jellysmack: Done

pkg/argocd/update.go
```diff
@@ -621,10 +621,15 @@ func getWriteBackConfig(app *v1alpha1.Application, kubeClient *kube.KubernetesCl
 	return wbc, nil
 }
 
-func parseDefaultTarget(appName string, path string) string {
-	defaultTargetFile := fmt.Sprintf(common.DefaultTargetFilePattern, appName)
-
-	return filepath.Join(path, defaultTargetFile)
+func parseDefaultTarget(appNamespace string, appName string, path string, kubeClient *kube.KubernetesClient) string {
+    log.Warnf("Namspace : %v", kubeClient)
```
chengfang, line 625: Should this be info level, instead of warn? Can we also make it clear that this is the control plane namespace? For the arg, did you mean `kubeclient.Namespace`?

chengfang (approved):

chengfang:
Thanks @sribiere-jellysmack !

pkg/common/constants.go
chengfang: @sribiere-jellysmack do you have a link to the argocd code or docs that describes the default target file pattern with namespace?  All I can find is https://argo-cd.readthedocs.io/en/latest/user-guide/parameters/#store-overrides-in-git but this does not mention the file name pattern that contains namespace.

PR #771, opened 2024-07-10
chore(tests): Add unit tests to image-updater pkg/health, pkg/version and pkg/registry

Mangaal:


pkg/health/health_test.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+// Unit test function
+func TestStartHealthServer(t *testing.T) {
+	port := 8080
+	errCh := StartHealthServer(port)
```
chengfang, line 12: is the health server stopped after the test?

chengfang (approved):

PR #793, opened 2024-07-22
feat: update several path for image.tag

Tchoupinax:
Hello folks!

I made this improvement to handle a case we have at work. Let's explain it.

In the current case, we provide a path for the image tag with the annotation `.helm.image-tag`. It says where to update the tag for this argoCD application.

Now, I have one ArgoCD application with two deployements, different with a suffix. Let's call them `a` and `b`. So my helm-chart values are prefixed by the alias of each app.

```yaml
a:
  image:
    tag: v1.0.0

b:
  image:
    tag: v1.0.0
```
I want to update both at the same time, the lifecycle is identical. This merge request allows to do it, by adding several paths as value separated by a comma. Then, I can provide

```yaml
xxxx.helm.image-tag=a.image.tag,b.image.tag
```

A unit test has been added about tag and name and moreover, I handled every space to ensure it won't have any bug for a mistype.

Finally, I build the customized image and tested, all is OK for me.
I hope all is clear!

Thank you for your work, I love this project.
Cheers!


chengfang:
@Tchoupinax thanks for proposing it. I'll look into it.

Tchoupinax:
Hello @chengfang,
Could you have time to give a look again? 🙏

pkg/argocd/argocd.go
```diff
@@ -427,12 +427,30 @@ func SetHelmImage(app *v1alpha1.Application, newImage *image.ContainerImage) err
 		mergeParams = append(mergeParams, p)
```
chengfang, line 427: any reason `hpImageSpec` is not updated to take multiple values too? Do we want to keep image.spec, image.name, and image.tag consistant in taking multi-path values?

chengfang:
Some other places assume that image.name and image.spec represent single-path value, and we should think about if it makes sense for those places to also take multi-path values. For instance,

https://github.com/argoproj-labs/argocd-image-updater/blob/4f21adef41a9dfad16ac6c1d7d7f53bc67e8a79d/pkg/argocd/update.go#L449

https://github.com/argoproj-labs/argocd-image-updater/blob/4f21adef41a9dfad16ac6c1d7d7f53bc67e8a79d/pkg/argocd/update.go#L473

pkg/argocd/argocd_test.go
```diff
@@ -1005,6 +1005,89 @@ func Test_SetHelmImage(t *testing.T) {
...
 
+	t.Run("Test set Helm image parameters when two paths are provided", func(t *testing.T) {
+		app := &v1alpha1.Application{
+			ObjectMeta: v1.ObjectMeta{
+				Name:      "test-app",
+				Namespace: "testns",
+				Annotations: map[string]string{
+					fmt.Sprintf(common.HelmParamImageNameAnnotation, "foobar"): "foobar.image.name,foobar2.image.name",
+					fmt.Sprintf(common.HelmParamImageTagAnnotation, "foobar"):  "foobar.image.tag, foobar2.image.tag", // Space is expected
+				},
```
chengfang, line 1016: also need to handle cases like below to exclude empty slice elements caused by extra separators.
`foobar.image.name,foobar2.image.name,`
`foobar.image.name,,,foobar2.image.name`

Tchoupinax: I edited the test to take care about this point and fixed the code according to

pkg/argocd/argocd.go
Tchoupinax: Indeed, I was not aware of this option because I did not use it. I added the same behavior and another test to handle this case!

Tchoupinax:
I will give a look but maybe I need help on your last point (checking if having two paths is OK everywhere)

PR #800, opened 2024-08-01
List applications in all Namespaces when trying to update

sribiere-jellysmack:
For now, AIU only lists application in ArgoCD Namespace before trying to update them

This PR allows AIU to fetch all namespaces for annotated applications


jannfis (changes requested):
Thank you for your PR.

I think this is a majorly breaking change as-is, and it seems incomplete as well.

For one, to list applications in all namespaces, we would have to update RBAC to include cluster wide permissions for Application CRs. Second, it should be up to the user/admin which mode (namespaced or cluster-wide) should be used.

PR #825, opened 2024-08-12
Add AWS credential provider

marcelobartsch-jt:
This code allows argo image updater to use ECR repos natively.

it is very basic code but get the job done, feel free to comment what is missing so I can update and improve this code.

thanks!


jannfis:
Thank you, I really do like this PR and the intent behind it.

However, I do have slight concerns about maintainability here. To maintain this, one needs to have an aws account and ECR set up. How would this be integrated into a potential end-to-end test? Who will be the one to take care of issues and bugs reported to this particular functionality?

Also, when we start incorporating proprietary authentication for vendor no. 1, then we're not far away from requests to incorporating similar mechanisms for vendor no. 2, vendor no. 3 and so on. This would multiply the impact described in the previous paragraph.

All that being said, please don't close or abandon this PR. I just want to raise these concerns to open up the discussion. Thank you!

marcelobartsch-jt:
About E2E test, they can be automated, of course you will need some infrastructure , or tools like localstack which emulate AWS, for other cloud platforms, no idea ,other option is to use mocked services,sadly I'm not that knowledged on testing in golang to implement , but is doable.
Now the big question is , no new features because we don't have the infra for testing, put those features as second class citizens.

Now in the case of new ones, you can just say 'Sure, provide a PR' , but I think is a natural path for people wanting to use this with less complications as possible, because, I could use the 'ext:/scripts/aws.sh' , for example, and just use that, but then again I need to modify the base image or mount it, but that might need aws cli, python and its depedencies.
Now another approach is to put those 'auth' as 'external' and provide a way to compile them apart, and instead of use 'aws' for the credentials, just use 'ext:/local/aws-ecr-auth', so users , like me, don't need to add python, and aws cli, and all it's dependencies to the base image.

Again, just thinking out loud looking to expand the discussion!

damienjacinto:
Hi, do you plan to consider merging this PR? What is still missing?
Having to rely on a script or a secret with external-secrets is not really optimal.

PR #827, opened 2024-08-13
docs: added release process

kostis-codefresh:
Explain how to make releases for Argo Image updater

PR #828, opened 2024-08-13
chore: removed unused workflow

kostis-codefresh:
This workflow is not used anymore. New documentation about releases was added with https://github.com/argoproj-labs/argocd-image-updater/pull/827


jannfis (approved):
LGTM

PR #830, opened 2024-08-13
feat: Use shallow clone to speed up performance

jnovick:
* Only git fetch the one target revision that we need
* This does not utilize the history so there is no reason to clone it
* For large repos, this will speed up performance significantly

### DIFF
--- a/ext/git/client.go
+++ b/ext/git/client.go
@@ -65,6 +65,7 @@ type Client interface {
 	Root() string
 	Init() error
 	Fetch(revision string) error
+	ShallowFetch(revision string, depth int) error
 	Submodule() error
 	Checkout(revision string, submoduleEnabled bool) error
 	LsRefs() (*Refs, error)
--- a/ext/git/git_test.go
+++ b/ext/git/git_test.go
@@ -322,6 +322,25 @@ func TestVerifyCommitSignature(t *testing.T) {
 	}
 }
 
+func TestVerifyShallowFetchCheckout(t *testing.T) {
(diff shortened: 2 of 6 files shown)


ext/git/client.go
```diff
@@ -384,6 +385,40 @@ func (m *nativeGitClient) Fetch(revision string) error {
 	return err
 }
 
+func (m *nativeGitClient) shallowFetch(revision string, depth int) error {
```
jannfis, line 388: Can you please move both functions to a separate file and only leave the change to the interface in this one? Reason is, that we copy over the Git code from Argo CD on every release. If this code stays in an file that exists in Argo CD too, it makes it pretty complicated to maintain. Thanks!

jnovick: No problem, moved it. I chose not to update the existing `fetch` function since I saw it was copied in, but I wasn't sure where to put new logic, but I found it now. 

You have made me wonder if I can contribute to ArgoCD, but updating it to use shallow clones as well. I will be investigating that later.

Thanks, I appreciate the help :D

jannfis: I think we discussed shallow cloning with Argo CD in the past a couple of times and came to no final conclusion yet. It's a pretty invasive change due to repository caching, history requirements for rollbacks and information etc.

For Image Updater however, I think there is no issue with shallow cloning and I believe it's beneficial.

jannfis (approved):
LGTM, thank you @jnovick! Appreciate your contributions.

PR #831, opened 2024-08-13
List applications in all Namespaces when trying to update

sribiere-jellysmack:
It looks like I closed PR 800 (I don't remember doing that) 

So I Recreated it 

For now, AIU only lists application in ArgoCD Namespace before trying to update them

This PR allows AIU to fetch all namespaces for annotated applications


sribiere-jellysmack:
Let's answer @jannfis ' comment on PR800

Now there is a parameter to chose wether we want namespaced or clusterwide listing.
Regarding CLusterRol and Binding, I will submit a PR on https://github.com/argoproj/argo-helm

sribiere-jellysmack:
Any comments please?

PR #832, opened 2024-08-13
feat!: Filter labels on the server instead of client to allow more label filtering options

jnovick:
Filter labels on the server instead of client
* This will allow supporting filtering by more label options
  * Multiple labels
  * Negative matching
  * Label exists regardless of value
* This should reduce cpu, memory, and networking resources needed
* Updated docs to reflect new capabilities

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -236,7 +236,7 @@ func newRunCommand() *cobra.Command {
 	runCmd.Flags().IntVar(&cfg.MaxConcurrency, "max-concurrency", 10, "maximum number of update threads to run concurrently")
 	runCmd.Flags().StringVar(&cfg.ArgocdNamespace, "argocd-namespace", "", "namespace where ArgoCD runs in (current namespace by default)")
 	runCmd.Flags().StringSliceVar(&cfg.AppNamePatterns, "match-application-name", nil, "patterns to match application name against")
-	runCmd.Flags().StringVar(&cfg.AppLabel, "match-application-label", "", "label to match application labels against")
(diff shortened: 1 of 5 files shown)


jannfis:
Will this work when using the Argo CD API instead of K8s to get the list of applications, too?

jnovick:
> Will this work when using the Argo CD API instead of K8s to get the list of applications, too?

I tested submitting requests to my ArgoCD instance in the same format as the kubernetes api and it worked. I would love assistance validating it with the code, but I believe this should work. I am relatively new to adding contributions to this code so I was a bit confused why there is a kubernetes and an ArgoCD implementation of `GetApplications`, but I updated both. Which one is actually used? Or are they both used?

I will be building this image tomorrow to test it myself (Along with the other PR I submitted today). If I see any issues, I will be sure to fix it in this PR and update you, but my local tests seemed to work.

jannfis:
Hey, yeah, seems you have caught the right spot already. Thanks. I think I'm ok with this change :)

Can you please fix the linter issue? Thanks.

jannfis (approved):
LGTM, this change makes sense.

It is a breaking change I suppose, so should be mentioned in the release notes CC @chengfang @pasha-codefresh

jnovick:
I'm not sure it is a breaking change since any label set before will work still with the same behavior externally, but I do agree that it is worth putting in the release notes. I can tell you that putting this into our environment today has allowed us to shard the image updater by label which has allowed us to do a lot more updates in parallel. The change works exactly how I hoped that it would. Thank you so much for the quick response to my PRs.

PR #838, opened 2024-08-22
Allow setting new value in Helm values file

jnovick:
* This was previously supported, but broke when the function was refactored to preserve order of yaml file (#748)
* Behavior does not change if overriding an existing value
* If the values file is empty or does not have a previous value, add the value instead of erroring

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -465,11 +465,11 @@ func marshalParamsOverride(app *v1alpha1.Application, originalData []byte) ([]by
 					return nil, fmt.Errorf("%s parameter not found", helmAnnotationParamVersion)
 				}
 
-				err = setHelmValue(helmNewValues, helmAnnotationParamName, helmParamName.Value)
+				err = setHelmValue(&helmNewValues, helmAnnotationParamName, helmParamName.Value)
 				if err != nil {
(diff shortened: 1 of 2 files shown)


jnovick:
@jannfis I have published an image to my own registry and tested it works as desired and the tests also show that it works. Let me know if you want me to make any changes.

@askhari If you have concerns, let me know since you implemented the change that broke the ability to write to a new file. I want to make sure you are aware in case this breaks a behavior that you are expecting.

chengfang (approved):

askhari:
> @jannfis I have published an image to my own registry and tested it works as desired and the tests also show that it works. Let me know if you want me to make any changes.
> 
> @askhari If you have concerns, let me know since you implemented the change that broke the ability to write to a new file. I want to make sure you are aware in case this breaks a behavior that you are expecting.

@jnovick , thank you for the notice :) . I see that now this fix is already merged. That's good news.
I didn't realize that I broke that behavior somehow. I tried to implement that piece of code to use its own workflow to avoid breaking things :| . As it was a new feature, at the beginning it was not intended to create new files or entries in the file because we didn't have that requirement. So, it's great that now it will allow to add new keys if needed.

Thank you for your help and improvement of the code :) .

Issue #809, opened 2024-08-05
Multiple alias updates doesn't work (Write back method: git, Write back method: helmvalues) 


dshmelev:
**Describe the bug**
ArgoCD Image Updater does not update the helm values file when multiple aliases of the same image are in application annotation.

**To Reproduce**

Create an application with annotations:
```
    argocd-image-updater.argoproj.io/write-back-method: git:secret:kube-gitops/argocd-repo-creds
    argocd-image-updater.argoproj.io/write-back-target: "helmvalues:/gitops/values/dev.yaml"
    argocd-image-updater.argoproj.io/git-repository: https://github.com/example/example.git
    argocd-image-updater.argoproj.io/git-branch: main
    argocd-image-updater.argoproj.io/image-list: foo=123456789.dkr.ecr.eu-west-1.amazonaws.com/application,bar=123456789.dkr.ecr.eu-west-1.amazonaws.com/application
    argocd-image-updater.argoproj.io/foo.helm.image-name: foo.spec.image.name
    argocd-image-updater.argoproj.io/foo.helm.image-tag: foo.spec.image.tag
    argocd-image-updater.argoproj.io/bar.helm.image-name: bar.spec.image.name
    argocd-image-updater.argoproj.io/bar.helm.image-tag: bar.spec.image.tag
```

Update 123456789.dkr.ecr.eu-west-1.amazonaws.com/application image and check the latest git commit.

Git commit contains updates only for one application.
```
bar.spec.image.name
bar.spec.image.tag
```

I wrote a short test that could help you to investigate an issue

```
	t.Run("Valid Helm source with Helm values file with multiple aliases", func(t *testing.T) {
		expected := `
foo.image.name: nginx
foo.image.tag: v1.0.0
bar.image.name: nginx
bar.image.tag: v1.0.0
replicas: 1
`
		app := v1alpha1.Application{
			ObjectMeta: v1.ObjectMeta{
				Name: "testapp",
				Annotations: map[string]string{
					"argocd-image-updater.argoproj.io/image-list":          "foo=nginx, bar=nginx,",
					"argocd-image-updater.argoproj.io/write-back-method":   "git",
					"argocd-image-updater.argoproj.io/write-back-target":   "helmvalues:./test-values.yaml",
					"argocd-image-updater.argoproj.io/foo.helm.image-name": "foo.image.name",
					"argocd-image-updater.argoproj.io/foo.helm.image-tag":  "foo.image.tag",
[...]

PR #846, opened 2024-08-30
fix: Multiple aliases for the same image only 1 parameter is updated

chengfang:
Fixes https://github.com/argoproj-labs/argocd-image-updater/issues/809

For multiple aliases pointing to the same image, parameters for each alias should be updated according to the application annotations. The test is contributed by dshmelev.


jannfis (approved):
LGTM! Thanks.

Issue #847, opened 2024-08-30
CreateApplicationEvent failure when App is in other namespace


jortkoopmans:
**Describe the bug**
When running (latest/master branch) updater, with Applications in other namespaces, Events will fail to be published, with a message like:
```
time="2024-08-29T17:04:39Z" level=warning msg="Event could not be sent: Event \"my-application.17f041de2f74d240\" is invalid: involvedObject.namespace: Invalid value: \"the-application-namespace\": does not match event.namespace" application=my-application
```

**To Reproduce**
- Configure an Application in another namespace than where the argocd-image-updater is running
- Configure update method argocd (not using kubernetes client, this is currently broken I think, but that's another issue)
- Trigger an update, that requires an Event to be published

Result is that the event is not published, this warning is logged instead.

**Expected behavior**
Events are published

**Additional context**
This is related to recent functionality on Apps in other namespaces.
Looking at the code, it seems to boil down to the kubernetes client being initiated in a particular namespace, which is not matching the Application.

I have a branch that resolves this issue (for me at least), which I will link to this issue for review.

**Version**
v0.14.0 and or built from master.

**Logs**
See above

chengfang:
The other issue related to 'apps in any namespace': https://github.com/argoproj-labs/argocd-image-updater/issues/601

jannfis:
Closed by #848

PR #848, opened 2024-08-30
feat: Sync namespace for Event to the Application namespace (#847)

jortkoopmans:
Sync namespace for Event to the Application namespace (#847)

Note that this does assume that the kubernetes client has permissions to create Events on the particular namespace (this typically depends on cluster RBAC)

### DIFF
--- a/manifests/base/rbac/argocd-image-updater-clusterrole.yaml
+++ b/manifests/base/rbac/argocd-image-updater-clusterrole.yaml
@@ -1,5 +1,5 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
 metadata:
   labels:
     app.kubernetes.io/name: argocd-image-updater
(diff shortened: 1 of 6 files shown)


jannfis (commented):
Will we need RBAC changes for that to allow the image updater the creation of events in those namespaces?

jortkoopmans:
> Will we need RBAC changes for that to allow the image updater the creation of events in those namespaces?

Thank you for the quick feedback.
Yes, some additional permissions are required, I've added a commit proposing to move the Event create permission to a ClusterRole.
Theoretically a cluster operator might prefer creating separate roles for specific namespaces where Applications will be residing, but managing those roles can become tedious.

jannfis (approved):
LGTM! Thanks for your contribution!

chengfang:
@jortkoopmans can you please take a look at the regression reported in issue #859?  Thanks.

PR #849, opened 2024-08-30
fix: warnings from golangci-lint: S1009 should omit nil check; non-constant format string in call to fmt.Errorf

chengfang:


jannfis (approved):
LGTM

PR #856, opened 2024-09-04
docs: use the correct annotation name platforms instead of platform; fix link to registries.md; improve table formatting

chengfang:


jannfis (approved):
LGTM

Issue #711, opened 2024-05-16
ArgoCD does not consider images for arch "linux/arm64/v8"


arontsang:
**Describe the bug**
linux/arm64/v8 != linux/arm64

**To Reproduce**
Create a manifest for a multi-arch image with one of the arch being "linux/arm64/v8"
Run ArgoCD image updater in k8s using the "linux/arm64" (note missing v8) image
Create application using digest update against multi-arch image

**Expected behavior**
ArgoCD pattern matches linux/arm64 with linux/arm64/v8

**Version**
0.13.0

arontsang:
Work around.

Explicitly add annotation:

`argocd-image-updater.argoproj.io/app.platforms: 'linux/arm64/v8'`

mehrdadpfg:
we had the same issue 
the problem is that the log indicates
```ignoring OCI manifest sha256:1b86e767c10ce0c0e0e64e183a0065d114adbd38b13638811229a19a2d23c598. Manifest platform: linux/arm64, requested: linux/arm64"```

and it does not tell that actually they are different
@arontsang solution worked for us as well

jtklahti:
Same thing was in #665

chengfang:
Please review the linked PR https://github.com/argoproj-labs/argocd-image-updater/pull/858 , and check if it fixes the problem, and if the change of behavior is acceptable.

PR #858, opened 2024-09-05
fix: allow the specific image platform os/arch/variant to meet the requirement for a broader image platform os/arch

chengfang:
Fixes https://github.com/argoproj-labs/argocd-image-updater/issues/711

* change `WantsPlatform()` method to allow match of broader platform
* add unit tests for the above change
* improve logs to include platform variant so it's clear why certain match is false
* this fix causes some change of behavior in comparing options platforms vs platforms of available images, as shown in the changes to `pkg/image/options_test.go`

### DIFF
--- a/pkg/image/options_test.go
+++ b/pkg/image/options_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"runtime"
+	"strings"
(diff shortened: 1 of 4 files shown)


jannfis (approved):
LGTM, great stuff!

Issue #720, opened 2024-05-24
Unable to access /app/.aws for configuring ECR credentials after 0.13.0 upgrade


schlags:
**Describe the bug**
It seems like this upgrade has (maybe?) broken functionality for using an authentication script for ECR like the following:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: argocd-image-updater-authscripts
data:
  # The following script must have a newline at the end of the output to work correctly, so we explicitly echo the output
  # https://argocd-image-updater.readthedocs.io/en/stable/basics/authentication/#using-a-script-to-generate-credentials
  ecr-login.sh: |
    #!/bin/sh

    userpass=$(aws ecr --region {replace with your region} get-authorization-token --output text --query 'authorizationData[].authorizationToken' | base64 -d) && echo $userpass
```

**To Reproduce**
1. Upgrade to v0.13.0
2. Define the above snippet for the `argocd-image-updater-authscripts` ConfigMap.
3. Set the annotations on an application for a private ECR registry
4. See logs in the image updater service along the lines of 
	```
	Could not set registry endpoint credentials: invalid script output, must be single line with syntax <username>:<password>
	```
5. Shell into the pod and run the command directly and see the following output:
	```

	/ $ /scripts/ecr-login.sh
	
	[Errno 30] Read-only file system: '/app/.aws'
	```
**Expected behavior**

Like the behavior in v0.12.2 this should be acceptable to run aws commands on the pod. Running this command locally does indeed output a single line that follows the `username:password` convention, but the issue is with the new `securityContext` definition on the container spec.

```
AWS:authorization-token
``` 

**Additional context**

Reverting back to v0.12.2 immediately fixed the issue, and logs do not return.

**Version**
Please tell us about the version you encountered the issue with

v0.13.0

**Logs**
Please paste any relevant logs here

```
time="yyyy-MM-ddTHH:mm:ssXXX" level=error msg="Could not set registry endpoint credentials: invalid script output, must be single line with syntax <username>:<password>" alias=* application=* image_name=*/* image_tag=* registry=REDACTED.dkr.ecr.us-west-2.amazonaws.com
```

schlags:
> Please note that the output should contain a newline character. 
https://argocd-image-updater.readthedocs.io/en/stable/basics/authentication/#using-a-script-to-generate-credentials

Is this new as of v0.13.0?

chapayevdauren:
the same here

[...]

Issue #112, opened 2020-10-23
Support ECR authentication


hobbsh:
**Is your feature request related to a problem? Please describe.**
It looks like there is a lack of support for getting the API token required to authenticate against the ECR API. It would be great to be able to make use of the IAM Instance Profile assigned to a node which has `GetAuthorizationToken` permissions. It might look similar to this: https://github.com/awslabs/amazon-ecr-credential-helper/blob/master/ecr-login/api/client.go.

**Describe the solution you'd like**
The image updater supports ECR authentication

**Describe alternatives you've considered**
There do not seem to be any alternatives without code changes.

hobbsh:
For additional context, this is the logged error I get when trying to use ECR:

```
time="2020-10-23T01:53:29Z" level=error msg="Could not get tags from registry: Get \"https://ACCOUNT_ID.dkr.ecr.us-west-2.amazonaws.com/v2/dev/my-image/tags/list\": http: non-successful response (status=401 body=\"Not Authorized\\n\")" alias=backend application=my-application image_name=dev/my-image image_tag=304dbc93e21898e95272d7cb81b61a671a8b7365 registry=ACCOUNT_ID.dkr.ecr.us-west-2.amazonaws.com
```

hobbsh:
We can use `aws ecr get-authorization-token --output text --query 'authorizationData[].authorizationToken'` and jam it into a k8s secret (it yields a base64 encoded `username:password` string) however the token is only valid for 12 hours so there would need to be a cronjob that updates this secret every 12 hours.

jannfis:
We currently have three methods for getting credentials right now - `secret`, `pullsecret` and `env`. Would a fourth option, that calls a script and re-uses its output as credentials, be sufficient for this use case?

For example we could introduce a new method `ext:/path/to/script.sh`, and you could either build your own image from the image updater's docker file or use an init container to copy the script to some location into the image updater's container.

jannfis:
A feature implementing my above proposal has been merged with #121 and will be part of v0.8 release.

jannfis:
I have no means to test against a real ECR instance, but would a script require some kind of parametrization in order to be able to retrieve credentials for the correct registry?

vistrcm:
Hello, 
I can try to help with testing. ECR support is crucial for me too.

One question before I can test: do credentials cached somehow? I mean, does argocd-image-updater read credentials from the secret, env variable or execute the script every time or only once and then use these results?

[...]

PR #880, opened 2024-10-03
docs: add note about write permissions for auth scripts on the filesystem

lexops:
This PR adds a note to the documentation addressing the issue where authentication scripts may require write access to the filesystem, which can lead to errors like `[Errno 30] Read-only file system`.


chengfang:
related discussion:
https://github.com/argoproj-labs/argocd-image-updater/issues/720
https://github.com/argoproj-labs/argocd-image-updater/issues/112

Issue #884, opened 2024-10-08
Git write back to helm values is incorrect during the first pass and corrupts existing data


kirgene:
**Describe the bug**
Using git write back with `helmvalues` and existing `values.yaml` file with user data leads to incorrect data writes and corruption of user data.

**To Reproduce**
1. Having this application:
```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: test-app-custom
  namespace: argo-cd
  annotations:
    argocd-image-updater.argoproj.io/write-back-method: git:secret:argo-cd/repo
    argocd-image-updater.argoproj.io/test-app.secret: argo-cd/docker#auth
    argocd-image-updater.argoproj.io/image-list: 'image1=XXXX.dkr.ecr.us-west-2.amazonaws.com/images/test-app,image2=XXXX.dkr.ecr.us-west-2.amazonaws.com/images/test-app2'
    argocd-image-updater.argoproj.io/image1.helm.image-name: image.name
    argocd-image-updater.argoproj.io/image1.helm.image-tag: image.tag
    argocd-image-updater.argoproj.io/image2.helm.image-name: image2.name
    argocd-image-updater.argoproj.io/image2.helm.image-tag: image2.tag
    argocd-image-updater.argoproj.io/image1.update-strategy: semver
    argocd-image-updater.argoproj.io/image2.update-strategy: semver
    argocd-image-updater.argoproj.io/write-back-target: "helmvalues:/apps/test-app/values.yaml"
    argocd-image-updater.argoproj.io/git-repository: "https://github.com/XXXX/test-gitops-test.git"
    argocd-image-updater.argoproj.io/git-branch: "dev"

spec:
  destination:
    namespace: test-app
    server: https://kubernetes.default.svc
  project: default
  sources:
  - chart: test-app
    helm:
      releaseName: test-app-custom
      valueFiles:
      - $vars/apps/test-app/values.yaml
    repoURL: XXXX.dkr.ecr.us-west-2.amazonaws.com/charts
    targetRevision: 0.4.0
  - ref: vars
    repoURL: https://github.com/XXXX/test-gitops-test
```
2. Content of the `/apps/test-app/values.yaml` in git repo:
```yaml
# comment
test-value1: one
test-value2: two
```
3. After the **first** pass of argocd image updater the the `values.yaml` file is modified like this:
```yaml
test-value1:
  tag: 1.1.0
test-value2: two
image2:
[...]

PR #885, opened 2024-10-11
fix: Git write back to helm values is incorrect during the first pass and corrupts existing data

chengfang:
Closes https://github.com/argoproj-labs/argocd-image-updater/issues/884


ishitasequeira (approved):
LGTM!!

pasha-codefresh (approved):

PR #888, opened 2024-10-15
fix(update-methods.md):add escaping to the template so helm doesn't consume it

galamdring:
If the {{ and }} are not escaped, Helm will consume those, and the whole template will be lost.

The original example: 
```yaml
argocd-image-updater.argoproj.io/git-branch: main:image-updater{{range .Images}}-{{.Name}}-{{.NewTag}}{{end}}
```

will render as image-updater because the whole template is lost. The fix to this is to escape the {{ and }} like so:
```yaml
argocd-image-updater.argoproj.io/git-branch: main:image-updater{{"{{"}}range .Images{{"}}"}}-{{"{{"}}.Name{{"}}"}}-{{"{{"}}.NewTag{{"}}"}}{{"{{"}}end{{"}}"}}
```


docs/basics/update-methods.md
```diff
@@ -214,8 +214,10 @@ the annotation. For example, the following would create a branch named
 `image-updater-foo/bar-1.1` based on `main` in the event an image with
 the name `foo/bar` was updated to the new tag `1.1`.
 
+Helm will try to consume this template unless you escape the opening and closing brackets.
+
 ```yaml
-argocd-image-updater.argoproj.io/git-branch: main:image-updater{{range .Images}}-{{.Name}}-{{.NewTag}}{{end}}
+argocd-image-updater.argoproj.io/git-branch: main:image-updater{{"{{"}}range .Images{{"}}"}}-{{"{{"}}.Name{{"}}"}}-{{"{{"}}.NewTag{{"}}"}}{{"{{"}}end{{"}}"}}
 ```
```
chengfang, line 221: Would it make sense to keep the original form for non-helm cases, and also for easy viewing of the real value?  We can add the helm-escaped form for helm cases.

PR #897, opened 2024-10-25
Add (or restore) support for helm image-spec annotation in writeback to helmvalues

mwijngaard:
Helmvalues writeback currently does not support the image-spec annotation. This should fix that.


mwijngaard:
It seems I had a stale checkout and this is already fixed

Issue #903, opened 2024-10-31
There are info level logs even though we set log.level to `warn` in configmap argocd-image-updater-config


wd:
**Describe the bug**

There are info level logs even though we set log.level to `warn` in configmap argocd-image-updater-config

**To Reproduce**
Install argocd-image-updater from helm chart, and set the logLevel to warn 
https://github.com/argoproj/argo-helm/blob/argo-cd-7.5.1/charts/argocd-image-updater/values.yaml#L146

There will be info logs printed. 

**Expected behavior**
Only logs above the warning should be printed. 

**Additional context**

**Version**
v0.15.0

**Logs**

```
time="2024-10-31T06:36:23Z" level=info msg="git checkout --force master" dir=/tmp/git-pixel-dev3424132615 execID=1b357
time="2024-10-31T06:36:24Z" level=info msg=Trace args="[git checkout --force master]" dir=/tmp/git-pixel-dev3424132615 operation_name="exec git" time_ms=338.770766
time="2024-10-31T06:36:24Z" level=info msg="git clean -ffdx" dir=/tmp/git-pixel-dev3424132615 execID=c4aba
time="2024-10-31T06:36:24Z" level=info msg=Trace args="[git clean -ffdx]" dir=/tmp/git-pixel-dev3424132615 operation_name="exec git" time_ms=30.726668
time="2024-10-31T06:36:24Z" level=info msg="Starting configmap/secret informers"
time="2024-10-31T06:36:25Z" level=info msg="Configmap/secret informer synced"
time="2024-10-31T06:36:25Z" level=info msg="secrets informer cancelled"
```

PR #904, opened 2024-10-31
fix: respect log level settings

wd:
This is an approach to fix https://github.com/argoproj-labs/argocd-image-updater/issues/903

However, there are still some info level logs printed. 

```
time="2024-10-31T08:51:43Z" level=info msg=Trace args="[git config user.name argocd]" dir=/tmp/git-rainmaker3532605552 operation_name="exec git" time_ms=1.117718
```

These logs are generated by codes listed below
1. https://github.com/argoproj-labs/argocd-image-updater/blob/master/ext/git/client.go#L814
2.  https://github.com/argoproj/argo-cd/blob/master/util/exec/exec.go#L54
3. https://github.com/argoproj/gitops-engine/blob/master/pkg/utils/tracing/logging.go#L43

They provided a way to set the log level, which is here https://github.com/argoproj/argo-cd/blob/master/util/log/logrus.go#L55. We can use the environment variable `ARGOCD_LOG_LEVEL` to set the log level. This can be done through the helm chart. 

Update:
I test with set below for the helm chart, and it works. 
```
extraEnv:
  - name: ARGOCD_LOG_LEVEL
    value: "warn"
```

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -40,20 +40,23 @@ func NewContext() *LogContext {
 
 // SetLogLevel sets the log level to use for the logger
 func SetLogLevel(logLevel string) error {
+	var level logrus.Level
 	switch strings.ToLower(logLevel) {
 	case "trace":
-		logger.SetLevel(logrus.TraceLevel)
+		level = logrus.TraceLevel
 	case "debug":
-		logger.SetLevel(logrus.DebugLevel)
+		level = logrus.DebugLevel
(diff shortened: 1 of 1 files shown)


jannfis:
Nice! Can you please fix the DCO by signing off your commit?

wd:
@jannfis Done. I fixed the CI.

jannfis (approved):
LGTM, thanks!

PR #912, opened 2024-11-03
fix: delay setting git.username and git.email config till commit time

chengfang:
In https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/argocd/git.go#L161-L167, it sets git.username and git.email config to the git client. These 2 configs are used by commit operation later. In some cases, the commit operation will not be needed, either because there are errors before the commit, or because the commit should be skipped due to absence of real changes.

This PR delays setting these 2 configs, which involves calling git command twice, till before the commit operation. In addition to the savings from skipping the 2 command calls, the logs will also be cleaner.

### DIFF
--- a/pkg/argocd/git.go
+++ b/pkg/argocd/git.go
@@ -158,14 +158,6 @@ func commitChangesGit(app *v1alpha1.Application, wbc *WriteBackConfig, changeLis
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


ishitasequeira (approved):
LGTM!!

jannfis (approved):
LGTM!

PR #922, opened 2024-11-08
fix(build): add step to Makefile to generate signing gpg key

chengfang:
This PR fixes the github action draft release failure in signing binaries.


ishitasequeira (approved):
LGTM!!

PR #933, opened 2024-11-12
refactor: Add registry pkg registry scanner

ishitasequeira:
Add registry pkg registry scanner. Skipped content related to ArgoCD Applications.


registry-scanner/pkg/registry/registry_test.go
```diff
@@ -0,0 +1,157 @@
...
+	"testing"
+	"time"
+
+	//nolint:staticcheck
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Test relies on image package which is not available yet. Will uncomment as soon as it is available.
+// func Test_GetTags(t *testing.T) {
```
ishitasequeira, line 14: Test relies on image package which is not available yet. Will uncomment as soon as it is available.

jannfis (approved):
LGTM

PR #952, opened 2024-12-03
Add image folder to registry scanner

ishitasequeira:
Add image folder to registry scanner module.


registry-scanner/go.mod
```diff
@@ -3,20 +3,81 @@ module github.com/argoproj-labs/argocd-image-updater/registry-scanner
 go 1.22.3
```
chengfang, line 3: image-updater currently has go version 1.22.0. I think they should be consistent.

registry-scanner/go.mod
```diff
@@ -3,20 +3,81 @@ module github.com/argoproj-labs/argocd-image-updater/registry-scanner
...
 
 require (
+	github.com/argoproj-labs/argocd-image-updater v0.15.1
```
chengfang, line 6: Will this sub-module build and work without dependency on image-updater? I think we should avoid depending on image-updater for easy consumption.

registry-scanner/go.mod
```diff
@@ -3,20 +3,81 @@ module github.com/argoproj-labs/argocd-image-updater/registry-scanner
...
 require (
 	github.com/Masterminds/semver/v3 v3.2.1
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
-	golang.org/x/sys v0.21.0 // indirect
+	golang.org/x/sys v0.24.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace k8s.io/client-go => k8s.io/client-go v0.31.0
```
chengfang, line 83: some dependency versions are different from those in image-updater. In general, I think we should keep them in sync. Also adding this sub-module to dependabot.yaml may help a bit.

ishitasequeira: Updated all the dependencies to be in sync and also added the package to dependabot.yaml

registry-scanner/go.mod
ishitasequeira: I think I missed updating one import to the correct package leading to the addition of image-updater dependency. Updated the import to correct package and removed dependency on image-updater.

chengfang (approved):

PR #962, opened 2024-12-05
Update references of pkg/tag pkg/log pkg/options to use modules from registry-scanner

ishitasequeira:
Update references of pkg/tag pkg/log pkg/options to use modules from registry-scanner


go.mod
```diff
@@ -175,7 +173,8 @@ require (
 )
 
 replace (
-	github.com/argoproj-labs/argocd-image-updater/registry-scanner => ../registry-scanner
+	// Uncomment for local testing
+	// github.com/argoproj-labs/argocd-image-updater/registry-scanner => ./registry-scanner/
```
ishitasequeira, line 177: Added a comment for stating that we need to uncomment this line for local testing if we have changes in registry-scanner module.

chengfang (approved):

PR #965, opened 2024-12-09
Improve Application updates when write-back method is ArgoCD

aaguilartablada:
When the write-back method is ArgoCD, the Image Updater sends an update to the ArgoCD Application on each iteration. No matter whether the new image has already been set or not, it always sends the update. In a situation where autoSync is disabled, this means that it continuously sends updates until somebody syncs the Application.

To reduce updates and improve resource usage, this pull request slightly changes the behavior. The Image Updater will check if the new image is already set in the Application source to apply the update or ignore it.

We have tested the changes in our ArgoCD instance to check the feature and you can see the results in the image below. Updates have been reduced. Once they are applied, updates disappear. CPU and memory usage have been improved.

Captura de pantalla 2024-12-09 a las 10 22 03

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -288,6 +288,18 @@ func UpdateApplication(updateConf *UpdateConfiguration, state *SyncIterationStat
 		if needsUpdate(updateableImage, applicationImage, latest) {
 
 			imgCtx.Infof("Setting new image to %s", applicationImage.WithTag(latest).GetFullNameWithTag())
+
+			// Check if new image is alredy set in Application Spec when write back is set to argocd
+			// and compare with new image
+			appImage, err := getAppImage(&updateConf.UpdateApp.Application, applicationImage.WithTag(latest))
+			if err != nil {
+				continue
+			}
+			if appImage == applicationImage.WithTag(latest).GetFullNameWithTag() {
+				imgCtx.Infof("New image %s already set in spec", applicationImage.WithTag(latest).GetFullNameWithTag())
+				continue
+			}
+
 			needUpdate = true
 
 			err = setAppImage(&updateConf.UpdateApp.Application, applicationImage.WithTag(latest))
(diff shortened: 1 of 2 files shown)


pkg/argocd/update.go
```diff
@@ -288,6 +288,18 @@ func UpdateApplication(updateConf *UpdateConfiguration, state *SyncIterationStat
 		if needsUpdate(updateableImage, applicationImage, latest) {
 
 			imgCtx.Infof("Setting new image to %s", applicationImage.WithTag(latest).GetFullNameWithTag())
```
chengfang, line 290: can we move this log line a few lines below around line 304, after `needUpdate = true`?

pkg/argocd/update.go
```diff
@@ -288,6 +288,18 @@ func UpdateApplication(updateConf *UpdateConfiguration, state *SyncIterationStat
...
+				continue
+			}
+			if appImage == applicationImage.WithTag(latest).GetFullNameWithTag() {
+				imgCtx.Infof("New image %s already set in spec", applicationImage.WithTag(latest).GetFullNameWithTag())
+				continue
+			}
+
 			needUpdate = true
 
 			err = setAppImage(&updateConf.UpdateApp.Application, applicationImage.WithTag(latest))
```
chengfang, line 305: `applicationImage.WithTag(latest)` and `applicationImage.WithTag(latest).GetFullNameWithTag()` are called a few times, maybe we should extract into variables to avoid computing it repeatedly.

pkg/argocd/update.go
```diff
@@ -288,6 +288,18 @@ func UpdateApplication(updateConf *UpdateConfiguration, state *SyncIterationStat
 		if needsUpdate(updateableImage, applicationImage, latest) {
 
 			imgCtx.Infof("Setting new image to %s", applicationImage.WithTag(latest).GetFullNameWithTag())
+
+			// Check if new image is alredy set in Application Spec when write back is set to argocd
```
chengfang, line 292: *already*

pkg/argocd/update.go
aaguilartablada: done

pkg/argocd/update.go
```diff
@@ -286,20 +286,32 @@ func UpdateApplication(updateConf *UpdateConfiguration, state *SyncIterationStat
...
 
 			if err != nil {
 				imgCtx.Errorf("Error while trying to update image: %v", err)
 				result.NumErrors += 1
 				continue
 			} else {
-				containerImageNew := applicationImage.WithTag(latest)
-				imgCtx.Infof("Successfully updated image '%s' to '%s', but pending spec update (dry run=%v)", updateableImage.GetFullNameWithTag(), containerImageNew.GetFullNameWithTag(), updateConf.DryRun)
-				changeList = append(changeList, ChangeEntry{containerImageNew, updateableImage.ImageTag, containerImageNew.ImageTag})
+				imgCtx.Infof("Successfully updated image '%s' to '%s', but pending spec update (dry run=%v)", updateableImage.GetFullNameWithTag(), appImageWithTag.GetFullNameWithTag(), updateConf.DryRun)
```
chengfang, line 313: looks like `appImageWithTag.GetFullNameWithTag()` in the above log line can also be replaced with the variable `appImageWithTagFull`. Can you rebase and fix it if you agree, to have CI run with the latest?

aaguilartablada: you are completely right. fixed!

chengfang (approved):

chengfang:
@aaguilartablada thanks for contributing this valuable enhancement!

chengfang:
0.15.2 was released on 2025-01-06: https://github.com/argoproj-labs/argocd-image-updater/releases

PR #967, opened 2024-12-13
update k8s.io/client-go from v1.5.2 to v0.31.0

kranurag7:
https://pkg.go.dev/k8s.io/client-go#section-readme
<details><summary>Details</summary>
<p>

image

</p>
</details> 

migrates to newer scheme to match kubernetes releases.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.31.0
 	k8s.io/apimachinery v0.31.0
-	k8s.io/client-go v1.5.2
+	k8s.io/client-go v0.31.0
 	sigs.k8s.io/kustomize/api v0.17.2
 	sigs.k8s.io/kustomize/kyaml v0.17.1
 )
--- a/registry-scanner/go.mod
+++ b/registry-scanner/go.mod
@@ -17,7 +17,7 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	k8s.io/api v0.31.0
 	k8s.io/apimachinery v0.31.0
-	k8s.io/client-go v1.5.2
+	k8s.io/client-go v0.31.0
 )
 
 require (


chengfang:
There is a replace directive at line 187 `k8s.io/client-go => k8s.io/client-go v0.31.0`. With the changes in this PR, we should be able to remove that replacement line.

The same change should also be applied to sub-module `registry-scanner/go.mod`

kranurag7:
@chengfang thanks for the review. 

I've addressed it via https://github.com/argoproj-labs/argocd-image-updater/pull/967/commits/a999ff310c183b923bd083ea41e75263a03ee60d 

two more changes that were done are:
1. A lot of replaces are not required now given we're using that version already. Please let me know if there's a historical context of doing it that way. 
2. I also bumped `k8s.io/*` deps to 0.31.2 and `argocd/v2` to 2.13.2

chengfang:
I would limit the changes to only those needed by client-go updates, not to overload this PR.

kranurag7:
> I would limit the changes to only those needed by client-go updates, not to overload this PR.

Thanks Again, I've done that. I can remove replaces in a follow-up PR.

chengfang:
Thank you. We usually follow argo-cd project, https://github.com/argoproj/argo-cd/blob/master/go.mod for common dependencies. Sometimes lags behind a bit but try to avoid upgrading earlier.

chengfang:
there many lines of changes in `go.sum`. Are they required for the client-go upgrade? When running `go get k8s.io/client-go@v0.31.0 && go mod tidy`, I don't see any changes in `go.sum`.

kranurag7:
I did the same again on main branch and I think the changes in go.sum are coming from changing go.mod files.

kranurag7:
`go get k8s.io/client-go@v0.31.0 && go mod tidy` this is correct that you won't get any changes in go.sum if you do this.

but we are also removing one entry (k8s.io/client-go) from `replaces` and when I run `go.sum` after removing that, we get diff in go.sum

chengfang:
> `go get k8s.io/client-go@v0.31.0 && go mod tidy` this is correct that you won't get any changes in go.sum if you do this.
> 
> but we are also removing one entry (k8s.io/client-go) from `replaces` and when I run `go.sum` after removing that, we get diff in go.sum

Looks like the original seemingly duplicate replacement line for client-go is there to override what `k8s.io/kubernetes v1.31.0` brings along with it. I think we should keep this replacement line intact to avoid any additional transitive deps from `k8s.io/kubernetes v1.31.0`. So the net changes will be in the 2 `go.mod` files. Sorry for the confusion.

kranurag7:
thanks I've updated it again only making changes in go.mod files.

chengfang (approved):

chengfang:
@kranurag7 an upgrade to argo-cd 2.13.1 would be nice and welcome. But I would not change these replacement lines since they are needed to override `k8s.io/kubernetes` libraries.

PR #971, opened 2024-12-18
Update CI to run tests for registry-scanner separately

ishitasequeira:
Update CI to run tests for registry-scanner separately

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -117,3 +117,27 @@ jobs:
       - name: Run manifest tests
         run: |
           make test-manifests
+
+  registry-scanner:
+    name: Ensure registry-scanner Go modules synchronicity and run tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Setup Golang
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+      - name: Download all Go modules
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-tests.yaml
```diff
@@ -117,3 +117,27 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Setup Golang
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.22'
+      - name: Download all Go modules
+        run: |
+          go mod download
```
chengfang, line 133: do we need to first `cd registry-scanner` here too?

.github/workflows/ci-tests.yaml
```diff
@@ -117,3 +117,27 @@ jobs:
...
+      - name: Check for tidyness of go.mod and go.sum
+        run: |
+          go mod tidy
+          git diff --exit-code -- .
+      - name: Run tests
+        env:
+          GNUPG_DISABLED: true
+        run: |
+          cd registry-scanner
```
chengfang, line 142: if we already change directory to registry-scanner in previous step, then we don't need `cd registry-scanner` here. Effectively move the cd command a few steps earlier.

.github/workflows/ci-tests.yaml
ishitasequeira: Good point. Moved the cd here.

PR #972, opened 2024-12-18
Update references of cache, env, health to use modules from registry-scanner

ishitasequeira:
Update references of cache, env, health to use modules from registry-scanner


cmd/run.go
```diff
@@ -12,11 +12,11 @@ import (
...
 	"github.com/argoproj-labs/argocd-image-updater/pkg/argocd"
 	"github.com/argoproj-labs/argocd-image-updater/pkg/common"
-	"github.com/argoproj-labs/argocd-image-updater/pkg/env"
-	"github.com/argoproj-labs/argocd-image-updater/pkg/health"
 	"github.com/argoproj-labs/argocd-image-updater/pkg/metrics"
 	"github.com/argoproj-labs/argocd-image-updater/pkg/registry"
 	"github.com/argoproj-labs/argocd-image-updater/pkg/version"
+	"github.com/argoproj-labs/argocd-image-updater/registry-scanner/pkg/env"
+	"github.com/argoproj-labs/argocd-image-updater/registry-scanner/pkg/health"
 	"github.com/argoproj-labs/argocd-image-updater/registry-scanner/pkg/log"
```
chengfang, line 20: Is the health server referenced by registry-scanner for it to be relocated?

registry-scanner/pkg/common/constants.go
```diff
@@ -56,6 +56,11 @@ const (
 	HelmPrefix                = "helmvalues"
 )
 
+// DefaultTargetFilePattern configurations related to the write-back functionality
+const DefaultTargetFilePattern = ".argocd-source-%s_%s.yaml"
+const DefaultTargetFilePatternWithoutNamespace = ".argocd-source-%s.yaml"
+const DefaultHelmValuesFilename = "values.yaml"
+
```
chengfang, line 63: I see both image-updater and registry-scanner have `constants.go`. Do we want to keep only fields required by registry-scanner in registry-scanner/constants.go, and remove those not used by registry-scanner?

cmd/run.go
ishitasequeira: It is not. Do you feel we should keep it in image-updater itself?

registry-scanner/pkg/common/constants.go
ishitasequeira: Yes, I am working on a logic to separate them out. It needs a bit of refactoring. I will have the PR updated.

ishitasequeira: I will move this change to a different PR.

cmd/run.go
chengfang: Yeah, since it's checks the health of argocd-image-updater, and it won't be able to run within registry-scanner. We can address that in a separate PR.

chengfang (approved):

PR #981, opened 2024-12-20
Improve Application updates when write-back method is ArgoCD (#965)

chengfang:
cherry-pick from master to release-0.15 to improve application updates when write-back method is ArgoCD


jannfis:
I think is too big of a change for a patch release. Probably should release that with a 0.16.0?

jannfis:
Ah. YOLO :)

jannfis (approved):
LGTM

Issue #501, opened 2022-12-08
Retain YAML formatting when using the Git write-back method 


sa-spag:
**Is your feature request related to a problem? Please describe.**
When the Image Updater updates a `.argocd-source-<appName>.yaml` or `kustomiztion.yaml` file, it does not retain their indentation. For instance, it may generate the following diff:

```diff
 apiVersion: kustomize.config.k8s.io/v1beta1
 resources:
-  - some-resource.yaml
+- some-resource.yaml
 images:
...
```

**Describe the solution you'd like**
Instead, the Image Updater should retain the existing formatting.

It turns out the issue is pretty common for other projects that edit YAML (see issues linked to https://github.com/kubernetes-sigs/kustomize/issues/3946). Since [version 0.11.1](https://github.com/kubernetes-sigs/kustomize/releases/tag/kyaml%2Fv0.11.1), kyaml enables retaining the source file indention (see  https://github.com/kubernetes-sigs/kustomize/pull/4043). This feature has been integrated in Flux's Image Automation Controller (see https://github.com/fluxcd/image-automation-controller/pull/366), and I believe we can implement a similar feature for the Image Updater.

**Describe alternatives you've considered**
None.

**Additional context**
None.

zagr0:
Would be really great to have the original file content retained, and the only replaced parts are changed.
In our scenario when we write back to git helm values file for new tag, all the values become sorted and reformatted.

chengfang:
Need to check how to implement this in https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/argocd/git.go#L338

hgambarian:
I still have the same problem, Im not using Kustomize but helm, and when it changes the image tag . it already formats the yaml and remove the whitespaces, how can I make sure it will not touch anything except image tag?

frozenbonito:
Indentation is now preserved, but blank lines in `kustomization.yaml` are still removed.
Since blank lines are sometimes intentionally inserted for readability, would it be possible to preserve them as well?

mhrznamn068:
> I still have the same problem, Im not using Kustomize but helm, and when it changes the image tag . it already formats the yaml and remove the whitespaces, how can I make sure it will not touch anything except image tag?

I have similar issue with helm

PR #1002, opened 2025-01-09
Keep yaml sequence indentation when writing back to Git to a kustomize file

AlessandroZanatta:
Hi,

I've created a PR that should solve #501. First time I (really) contribute to an open-source project, so feel free to criticize anything I may have done wrong!

I have based this PR on the v0.15.2 branch instead of master as master was failing for me (throwing panics at runtime in my cluster). Let me know whether this is fine.

I've basically changed the way the `updateKustomizeFile` function in `pkg/argocd/git.go` works. It now uses `kio.ByteReadWriter`, which allows to preserve sequence indentation. 
I also had to slightly modify how the comparison is done: `kio` adds some `metadata.annotations` to the yaml document, hence my approach was to get the output from the writer (which clears those annotations), and then parse it again with `kyaml`, hence ensuring that indentation is the same across the two files.

I'm also planning of sending a PR to `kyaml` to add a boolean that optionally retains the initial triple dashes of a document, if present (though I have yet to dig on `kyaml` code). To me, that would be a very nice achievement, as git diffs would become truly the smallest they can be!

I hope that my explanation (and code) make sense!

EDIT: added a test case for the change

### DIFF
--- a/pkg/argocd/git.go
+++ b/pkg/argocd/git.go
@@ -382,7 +382,7 @@ func updateKustomizeFile(filter kyaml.Filter, path string) (error, bool) {
 		PreserveSeqIndent: true,
 	}
 
-	// Read input file
+	// Read from input buffer
 	newYSlice, err := rw.Read()
 	if err != nil {
 		return err, false
@@ -405,7 +405,7 @@ func updateKustomizeFile(filter kyaml.Filter, path string) (error, bool) {
 		return err, false
 	}
 
-	// yCpy contains metadata used by kio to preserve sequence indentation,
+	// newY contains metadata used by kio to preserve sequence indentation,
 	// hence we need to parse the output buffer instead
 	newParsedY, err := kyaml.Parse(out.String())
 	if err != nil {
@@ -416,12 +416,13 @@ func updateKustomizeFile(filter kyaml.Filter, path string) (error, bool) {
 		return err, false
 	}
 
-	// Diff the updated document with the original
+	// Compare the updated document with the original document
 	if originalData == newData {
 		log.Debugf("target parameter file and marshaled data are the same, skipping commit.")
 		return nil, true
 	}
 
+	// Write to file the changes
 	if err := os.WriteFile(path, out.Bytes(), 0600); err != nil {
 		return err, false
 	}
--- a/pkg/argocd/git.go
+++ b/pkg/argocd/git.go
@@ -350,72 +350,74 @@ func writeKustomization(app *v1alpha1.Application, wbc *WriteBackConfig, gitC gi
 // to the file. This is the same behavior as kyaml.UpdateFile, but it preserves the original order of YAML fields
 // and indentation of YAML sequences to minimize git diffs.
 func updateKustomizeFile(filter kyaml.Filter, path string) (error, bool) {
-	// Read the yaml
-	y, err := kyaml.ReadFile(path)
-	if err != nil {
-		return err, false
-	}
-
-	originalData, err := y.String()
-	if err != nil {
-		return err, false
-	}
-
 	// Open the input file for read
-	in, err := os.Open(path)
+	yRaw, err := os.ReadFile(path)
 	if err != nil {
 		return err, false
 	}
-	defer in.Close()
+
+	// Read the yaml document from bytes
+	originalYSlice, err := kio.FromBytes(yRaw)
+	if err != nil {
+		return err, false
+	}
+
+	// Check that we are dealing with a single document
+	if len(originalYSlice) != 1 {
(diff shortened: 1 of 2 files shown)


AlessandroZanatta:
I can already see that I screwed up, as all the commits from v0.15.2 are listed above.

Also, should I add a `Signed-off-by: ...` to my commit message to make the DCO action pass?

AlessandroZanatta:
Sorry about the spam with the commit on my repo. I was hoping for GitHub not to mention a commit PR link on an unrelated repository in the actual PR... seems I was wrong!

chengfang:
Thanks for implementing this enhancement. Will take a closer look.

pkg/argocd/git_test.go
```diff
@@ -267,6 +267,18 @@ func Test_updateKustomizeFile(t *testing.T) {
...
+		{
+			name: "indented",
+			content: `images:
+  - name: foo
+    digest: sha12345
+`,
+			wantContent: `images:
+  - name: foo
+    digest: sha23456
 `,
```
chengfang, line 282: In the test, can you use a different indentation, for example,
```yaml
		{
			name: "indented",
			content: `
images:
  - name: foo
    digest: sha12345
`,
			wantContent: `
images:
  - name: foo
    digest: sha23456
`,
			filter: filter,
		},
`,
```
so the list element is indented by 2 spaces, or some other variations. I think the current behavior is to write list elements without indentation, so test data of list elements without indentation will always pass even without this fix.

pkg/argocd/git.go
```diff
@@ -359,8 +361,34 @@ func updateKustomizeFile(filter kyaml.Filter, path string) (error, bool) {
...
+	yCpySlice, err := rw.Read()
+	if err != nil {
+		return err, false
+	}
+
+	// We expect a single yaml
+	if len(yCpySlice) != 1 {
+		return errors.New("target parameter file should contain a single YAML document"), false
+	}
+	yCpy := yCpySlice[0]
```
chengfang, line 389: in the beginning of the func, we called `kyaml.ReadFile` to get the originl RNode and originalData. Then here we call `os.Open` to get the input stream to kio ReadWriter. I think we could remove the first read, and use the 2nd read to get the original RNode and originalData before any modification. WDTY?

pkg/argocd/git_test.go
AlessandroZanatta: I'm not sure I understand what you mean. I have included a test that checks that a given indentation is retained after an update, and it looks like your example. Could you please clarify?

chengfang: You're right. Sorry, I might have misread it.

pkg/argocd/git.go
AlessandroZanatta: @chengfang Just a heads up in case my commit got lost amidst other stuff. I've updated the code, feel free to review it when you've got the time to do so, thanks!

pkg/argocd/git.go
```diff
@@ -345,43 +347,82 @@ func writeKustomization(app *v1alpha1.Application, wbc *WriteBackConfig, gitC gi
...
+
+	// Create a reader, preserving indentation of sequences
+	var out bytes.Buffer
+	rw := &kio.ByteReadWriter{
+		Reader:            bytes.NewBuffer(yRaw),
+		Writer:            &out,
+		PreserveSeqIndent: true,
+	}
+
+	// Read input file
```
chengfang, line 385: shall we change it to "// Read from input buffer" instead of file?

pkg/argocd/git.go
```diff
@@ -345,43 +347,82 @@ func writeKustomization(app *v1alpha1.Application, wbc *WriteBackConfig, gitC gi
...
 		return err, false
 	}
 
-	override, err := yCpy.String()
+	// Write the yaml document to the output buffer
+	if err = rw.Write([]*kyaml.RNode{newY}); err != nil {
+		return err, false
+	}
+
+	// yCpy contains metadata used by kio to preserve sequence indentation,
```
chengfang, line 408: did you mean `newY` instead of `yCpy` in the above comment?
Just 2 optional nit comments. Overall looks good.

pkg/argocd/git.go
AlessandroZanatta: Updated!

pkg/argocd/git.go
AlessandroZanatta: Yes, I missed that, thanks. Updated!

chengfang (approved):

hgambarian:
Hey, Is it possible to fix this for helm values as well?

PR #1003, opened 2025-01-09
CVE-2024-45338 in image-updater: Non-linear parsing of case-insensitive content in golang.org/x/net/html

chengfang:
https://github.com/advisories/GHSA-w32m-9786-jp63


jannfis (approved):
LGTM!

PR #1007, opened 2025-01-10
chore(deps): upgrade registry-scanner submodule version to the latest

chengfang:


jannfis (approved):
LGTM.

I was asking myself whether we should use a local reference in replace rules?

chengfang:
Towards the end of go.mod, there is a commented replacing line to be used for local testing.
https://github.com/argoproj-labs/argocd-image-updater/blob/ae473dd84031473a758bd93e9ab9558e2a2385e5/go.mod#L176-L177

PR #1014, opened 2025-01-17
Keep the "library" namespace if it was specified in the image identifier

### LINKED ISSUE #1012: Docker Hub default namespace is removed on write back
I would like to be able to specify every image name as a fully qualified name, that is, `registry.url/namespace/imagename:tag`.

However, this is not possible for `docker.io/library/<...>`, as the `library/` part [gets removed here](https://github.com/argoproj-labs/argocd-image-updater/blob/b1555062f8fced539d5685203d4115e910270bba/registry-scanner/pkg/image/image.go#L45) (I think).

When using git write-back-method and the kustomization write-back-target, that is:
```
    argocd-image-updater.argoproj.io/write-back-method: git
    argocd-image-updater.argoproj.io/write-back-target: kustomization
```
it results in the following kustomization update:
```
images:
  - name: docker.io/postgres
    newTag: 16.6-alpine
```
which does not match the image tag in the pod, such as `docker.io/library/postgres:16-alpine`, hence the pod is not rolled to the new image by argocd.

Would it be possible to avoid removing the default namespace? The comment [here](https://github.com/argoproj-labs/argocd-image-updater/blob/b1555062f8fced539d5685203d4115e910270bba/registry-scanner/pkg/image/image.go#L44) does not specify why this gets removed, could you please clarify? If it's something that can be fixed, I'd be happy to send a PR for it!

I've also tried specifying the Docker Hub registry in the configuration with a blank `defaultNs`, but that resulted in errors from the registry:

```
time="2025-01-14T08:42:37+01:00" level=error msg="Could not get tags from registry: errors:\ndenied: requested access to the resource is denied\nunauthorized: authentication required\n" alias=postgres application=anime image_name=postgres image_tag=16-alpine registry=docker.io
```

I do not see a way to enforce retaining the `library` part in the image name in the docs (unless I missed it, that is).

The alternative is leaving the `library/` part out of the image names, which I do not particularly like given that all the other images are fully qualified.

**Additional context**
I think it's a pretty niche feature, but also a slight improvement that could be added to this amazing project!

AlessandroZanatta:
I have implemented a change that should fix #1012.

Let me know if checking with `string.Contains` is enough. It theoretically does not comply with the OCI specification, as it assumes that there is a single namespace, but most registries do not allow nested namespaces ([ref](https://docs.docker.com/reference/cli/docker/image/tag/)).


AlessandroZanatta:
@jannfis @chengfang Could you please have a look at it please? I think it should be a quick PR

AlessandroZanatta:
Ping @jannfis @chengfang @pasha-codefresh @ishitasequeira

PR #1020, opened 2025-01-25
fix: Prevent segfault with empty kustomization

Ongy:
I'm not entirely sure when this happens, but the kustomize can be nil. Accessing Images then leads to a nil-deref.
With this approach, the updater defaults to empty images when it cannot get the real ones.


chengfang (approved):

Ongy:
I already added `Signed-off-by: Markus Ongyerth <ongyerth@google.com>` to the commit itself. Does it check the message here?

Though it might also be unhappy because that's not the same email as in the commit itself. I'll update that later today and see.

jannfis (approved):
I've set the DCO to `pass`, as I believe we're going to deprecate it soon anyway.

Change LGTM. Thanks!

Issue #1034, opened 2025-02-07
Duplicate image entries with different formats when updating two images


etanase:
**Describe the bug**
When updating the hash of one image using git method and digest strategy, argocd creates a copy of the other image without the ":latest" tag. This results in the containers not being re-created and remaining on the old images. This can be resolved by manually removing the hashes without the latest tag (after each update). For example, this is what the yaml looks like after attempting to update both images:
```yaml
# .argocd-source-app.yaml
kustomize:
  images:
  - ghcr.io/org/repo/image-a:latest@sha256:new-hash
  - ghcr.io/org/repo/image-b:latest@sha256:new-hash
  - ghcr.io/org/repo/image-a@sha256:old-hash
  - ghcr.io/org/repo/image-b@sha256:old-hash
```

**To Reproduce**
Have an application that uses two different images, and use these parameters:
```yaml
argocd-image-updater.argoproj.io/update-strategy: digest
argocd-image-updater.argoproj.io/write-back-method: git
```

**Expected behavior**
The image is updated with the latest tag, and no other entries are created.

**Version**
v0.15.2

chengfang:
I tested with my sample app https://github.com/chengfang/image-updater-examples/tree/main/image-list-kustomize, and was able to reproduce the issue. See the faulty generated file https://github.com/chengfang/image-updater-examples/blob/main/image-list-kustomize/source/.argocd-source-image-list-kustomize.yaml

In more details, if there is no existing write-back target file (.argocd-source-*.yaml) in the target repo, a new file was created correctly, without the issue seen above.

Then I manually deleted sha portion from one of the image, to trigger image-updater to update only one  of the images. After reinstalling and app and running image-updater, I was able to see the issue in the modified `.argocd-source-*.yaml` file.

meme8383:
I was also able to resolve the issue in my application by using the newest-build update strategy and removing the :latest tag from my deployments.

meme8383:
Fixed in 0.16.0

PR #1037, opened 2025-02-13
fix(write-back): Duplicate image entries with different formats when updating two images

chengfang:
Fixes https://github.com/argoproj-labs/argocd-image-updater/issues/1034

Modified `mergeKustomizeOverride` method to have more control over setting and adding parameters to `.argocd-source-*.yaml` target file.

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -551,13 +551,24 @@ func mergeHelmOverride(t *helmOverride, o *helmOverride) {
 }
 
 func mergeKustomizeOverride(t *kustomizeOverride, o *kustomizeOverride) {
-	for _, image := range *o.Kustomize.Images {
-		idx := t.Kustomize.Images.Find(image)
-		if idx != -1 {
-			(*t.Kustomize.Images)[idx] = image
-			continue
+	for _, newImage := range *o.Kustomize.Images {
+		found := false
+		newContainerImage := image.NewFromIdentifier(string(newImage))
+		for idx, existingImage := range *t.Kustomize.Images {
+			existingContainerImage := image.NewFromIdentifier(string(existingImage))
+			if newContainerImage.ImageName == existingContainerImage.ImageName &&
+				newContainerImage.RegistryURL == existingContainerImage.RegistryURL {
+				found = true
+				if existingContainerImage.ImageTag == nil ||
+					(newContainerImage.ImageTag != nil && !(existingContainerImage.ImageTag).Equals(newContainerImage.ImageTag)) {
+					(*t.Kustomize.Images)[idx] = newImage
+				}
+				break
+			}
 		}
-		*t.Kustomize.Images = append(*t.Kustomize.Images, image)
+		if !found {
+			*t.Kustomize.Images = append(*t.Kustomize.Images, newImage)
+		}
 	}
 }
 
--- a/pkg/argocd/update_test.go
+++ b/pkg/argocd/update_test.go
@@ -1219,6 +1219,49 @@ kustomize:
 		assert.Equal(t, strings.TrimSpace(expected), strings.TrimSpace(string(yaml)))
 	})
 
+	t.Run("Merge images param", func(t *testing.T) {
+		expected := `
+kustomize:
+  images:
+  - existing:latest
+  - updated:latest
+  - new
+`
+		app := v1alpha1.Application{
+			ObjectMeta: v1.ObjectMeta{
+				Name: "testapp",
+				Annotations: map[string]string{
+					"argocd-image-updater.argoproj.io/image-list": "nginx",
+				},
+			},
+			Spec: v1alpha1.ApplicationSpec{
+				Source: &v1alpha1.ApplicationSource{
+					RepoURL:        "https://example.com/example",
+					TargetRevision: "main",
+					Kustomize: &v1alpha1.ApplicationSourceKustomize{
+						Images: v1alpha1.KustomizeImages{
+							"new",
+							"updated:latest",
+						},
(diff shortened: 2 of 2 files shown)


pkg/argocd/update_test.go
```diff
@@ -3643,3 +3643,58 @@ func Test_GetRepositoryLock(t *testing.T) {
...
+			[]v1alpha1.KustomizeImage{"nginx:foo"},
+			[]v1alpha1.KustomizeImage{"nginx:foo"}},
+		{"with-tag-1", []v1alpha1.KustomizeImage{"nginx"},
+			[]v1alpha1.KustomizeImage{"nginx:latest"},
+			[]v1alpha1.KustomizeImage{"nginx:latest"}},
+		{"with-tag-sha", []v1alpha1.KustomizeImage{"nginx:latest"},
+			[]v1alpha1.KustomizeImage{"nginx:latest@sha256:91734281c0ebfc6f1aea979cffeed5079cfe786228a71cc6f1f46a228cde6e34"},
+			[]v1alpha1.KustomizeImage{"nginx:latest@sha256:91734281c0ebfc6f1aea979cffeed5079cfe786228a71cc6f1f46a228cde6e34"}},
+
+		{"2-images", []v1alpha1.KustomizeImage{"nginx:latest",
```
keithchong, line 3664: Cheng, do you have an example where the new image tag replaces the old image tag to test out the conditional statements in your new code in update.go?

pkg/argocd/update.go
```diff
@@ -551,13 +551,24 @@ func mergeHelmOverride(t *helmOverride, o *helmOverride) {
...
+	for _, newImage := range *o.Kustomize.Images {
+		found := false
+		newContainerImage := image.NewFromIdentifier(string(newImage))
+		for idx, existingImage := range *t.Kustomize.Images {
+			existingContainerImage := image.NewFromIdentifier(string(existingImage))
+			if newContainerImage.ImageName == existingContainerImage.ImageName &&
+				newContainerImage.RegistryURL == existingContainerImage.RegistryURL {
+				found = true
+				if existingContainerImage.ImageTag == nil ||
+					!(existingContainerImage.ImageTag).Equals(newContainerImage.ImageTag) {
```
keithchong, line 563: From what I see, at least from the tests that I'm trying to run, the new image tag can be nil and so the newImage gets added

pkg/argocd/update.go
```diff
@@ -551,13 +551,24 @@ func mergeHelmOverride(t *helmOverride, o *helmOverride) {
...
-		if idx != -1 {
-			(*t.Kustomize.Images)[idx] = image
-			continue
+	for _, newImage := range *o.Kustomize.Images {
+		found := false
+		newContainerImage := image.NewFromIdentifier(string(newImage))
+		for idx, existingImage := range *t.Kustomize.Images {
+			existingContainerImage := image.NewFromIdentifier(string(existingImage))
+			if newContainerImage.ImageName == existingContainerImage.ImageName &&
+				newContainerImage.RegistryURL == existingContainerImage.RegistryURL {
```
keithchong, line 560: From the tests that I performed, RegistryURL is nil all the time.  Is there another test that will test this condition with non-nil values?

pkg/argocd/update_test.go
```diff
@@ -3643,3 +3643,58 @@ func Test_GetRepositoryLock(t *testing.T) {
 	require.NotNil(t, state.repositoryLocks[repo2])
 	require.Equal(t, lock3, state.repositoryLocks[repo2])
 }
+
+func Test_mergeKustomizeOverride(t *testing.T) {
```
keithchong, line 3647: I was trying to test the new function from a higher level, at calling `marshalParamsOverride` instead.  Do you think it is worthwhile to do it from that call?  eg. see the first couple of scenarios from the existing test `Test_MarshalParamsOverride`

pkg/argocd/update_test.go
chengfang: the sub-tests: `with-tag-1`, `with-tag-sha`, and `2-images` all replace the existing tag with the new tag.

pkg/argocd/update.go
chengfang: In real applications, new image tag is unlikely to be nil. If it is, something already went wrong earlier.

But if the new tag is nil, it will cause nil pointer dereferece and panic. We could add a nil check to prevent that from happening:

```go
if existingContainerImage.ImageTag == nil ||
(newContainerImage.ImageTag != nil && !(existingContainerImage.ImageTag).Equals(newContainerImage.ImageTag)) {
```

pkg/argocd/update.go
chengfang: `RegistryURL` has a default value "" and should not be nil.

The first 4 sub-tests don't use any registry url, so they should be able to test this case.

pkg/argocd/update_test.go
chengfang: I can add a sub-test to `Test_MarshalParamsOverride` by setting the new images in app spec, to test out the changes from that level.

keithchong: Ok, I see the call to marshalParamsOverride

pkg/argocd/update_test.go
keithchong: I rechecked this again and yeah, the first test `with-tag`, the existing and the new tag are the same.  The next three tests, by definition of your test case struct, the tags are different.

keithchong (approved):
LGTM after re-review.

jannfis (approved):
LGTM

PR #1039, opened 2025-02-13
feat: preserve formatting and comments when updating Helm values file

kirgene:
## Description

This PR enhances the Helm values file update functionality by preserving the original formatting and comments. Previously, updating the file would overwrite manual formatting and comments, which could be problematic for users who rely on these customizations for clarity and maintenance. By retaining inline comments and specific formatting, this change improves the user experience and reduces potential confusion after automated updates.

Additionally, this update refines how `image-name` and `image-tag` keys are handled:

- By default, key path in `image-name` and `image-tag` creates a **nested structure**. For example:

  ```yaml
  argocd-image-updater.argoproj.io/nginx.helm.image-name: image.name
  argocd-image-updater.argoproj.io/nginx.helm.image-tag: image.tag
  ```

  Produces:

  ```yaml
  image:
    name: app-name
    tag: v1.0.0
  ```

- If a user wants to update Helm values that contain dots, the key path must be quoted (either with double or single quotes):
  ```yaml
  argocd-image-updater.argoproj.io/nginx.helm.image-name: '"image.name"'
  argocd-image-updater.argoproj.io/nginx.helm.image-tag: '"image.tag"'
  ```
  This will correctly produce:
  ```yaml
  image.name: app-name
  image.tag: v1.0.0
  ```

These changes ensure greater flexibility and correctness in handling Helm values updates while maintaining user-defined formatting.

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"text/template"
@@ -21,7 +22,11 @@ import (
 
 	"github.com/argoproj/argo-cd/v2/pkg/apiclient/application"
 	"github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
-	"gopkg.in/yaml.v2"
+
+	"github.com/goccy/go-yaml"
+	"github.com/goccy/go-yaml/ast"
+	"github.com/goccy/go-yaml/parser"
+	"github.com/goccy/go-yaml/token"
 )
 
 // Stores some statistics about the results of a run
@@ -459,8 +464,7 @@ func marshalParamsOverride(app *v1alpha1.Application, originalData []byte) ([]by
 		if strings.HasPrefix(app.Annotations[common.WriteBackTargetAnnotation], common.HelmPrefix) {
 			images := GetImagesAndAliasesFromApplication(app)
 
-			helmNewValues := yaml.MapSlice{}
-			err = yaml.Unmarshal(originalData, &helmNewValues)
+			helmNewValues, err := parser.ParseBytes(originalData, parser.ParseComments)
 			if err != nil {
 				return nil, err
 			}
@@ -488,7 +492,7 @@ func marshalParamsOverride(app *v1alpha1.Application, originalData []byte) ([]by
 					if helmParamVersion == nil {
 						return nil, fmt.Errorf("%s parameter not found", helmAnnotationParamVersion)
 					}
-					err = setHelmValue(&helmNewValues, helmAnnotationParamVersion, helmParamVersion.Value)
+					err = setHelmValue(helmNewValues, helmAnnotationParamVersion, helmParamVersion.Value)
 					if err != nil {
 						return nil, fmt.Errorf("failed to set image parameter version value: %v", err)
 					}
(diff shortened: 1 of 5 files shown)


pkg/argocd/update_test.go
```diff
@@ -1538,20 +1536,6 @@ replicas: 1
...
-		expected = `
-test-value1: one
-image:
-  spec:
-    foo: nginx:v1.0.0
-`
-		yaml, err = marshalParamsOverride(&app, originalData)
-		require.NoError(t, err)
-		assert.NotEmpty(t, yaml)
-		assert.Equal(t, strings.TrimSpace(strings.ReplaceAll(expected, "\t", "  ")), strings.TrimSpace(string(yaml)))
```
chengfang, line 1554: Any reason this part of the test is removed? I think we should keep any existing test as much as we can, and add new tests for introduced new features. 

The above test verifies that 
* the updater can infer the correct param name (whether to interpret dotted names as nested or flat ones) based on the existing format in use. 
* any added new parameters will not corrupt or interfere with any existing parameters.

pkg/argocd/update_test.go
```diff
@@ -1640,25 +1624,6 @@ replicas: 1
...
-    name: nginx
-redis:
-  image:
-    tag: v1.0.0
-    name: redis
-`
-		yaml, err = marshalParamsOverride(&app, originalData)
-		require.NoError(t, err)
-		assert.NotEmpty(t, yaml)
-		assert.Equal(t, strings.TrimSpace(strings.ReplaceAll(expected, "\t", "  ")), strings.TrimSpace(string(yaml)))
```
chengfang, line 1661: Any reason this part of the test is removed? I think we should keep any existing test as much as we can, and add new tests for introduced new features. 

The above test verifies that 
* the updater can infer the correct param name (whether to interpret dotted names as nested or flat ones) based on the existing format in use. 
* any added new parameters will not corrupt or interfere with any existing parameters.

pkg/argocd/update.go
```diff
@@ -561,76 +565,118 @@ func mergeKustomizeOverride(t *kustomizeOverride, o *kustomizeOverride) {
...
+
+	return result
 }
 
 // set value of the parameter passed from the annotations.
-func setHelmValue(currentValues *yaml.MapSlice, key string, value interface{}) error {
-	// Check if the full key exists
-	if idx, found := findHelmValuesKey(*currentValues, key); found {
-		(*currentValues)[idx].Value = value
-		return nil
```
chengfang, line 579: the current behaviro is, if the full key (e.g., `images.nginx.name`) already exists in the target helm file, take this existing key and update its value. Otherwise, assume the users want nested parameter names. I think this should be perserved when the annotation value `helm.image-name` and `helm.image-tag` are not explicitly quoted.

Even when the 2 annotation values are explicitly quoted, it's still worth considering whether we should honor the existing key format (flat dotted or nested) over the application annotation.

bazzargh:
FYI, I was just doing a similar thing because we need it for work, then noticed this PR. https://github.com/bazzargh/argocd-image-updater/tree/bazzargh/yamlv3 - my approach was a bit different; I switched from yamlv2 to yamlv3, and yaml.MapSlice to yaml.Node, which will preserve the various scalar types and comments (to an extent. it does not guarantee the same indentation, blank lines between comments may go, etc). One thing I do that this PR doesn't is support Aliases (from reading this, I think this PR supports leaving aliases alone if they're not part of the update path, but will not follow aliases to update a value, for example; our developers have expressed frustration with the way they can't use aliases with this tool)

Your PR does a couple of things I didn't - I also wanted to build the expected/input structures by parsing yaml instead of inlining the very verbose code, but held off to stick with the original style. I also didn't change the odd handling of top level dotted keys and prefer what you did (tho clearly it's a breaking change)

I only just coded this up, some co-workers were interested in having codefresh pick this up; we might run some tests with it this week to get some production experience before submitting a PR; also I wanted to add tests for the alias processing.

Anyway - just letting you know this is in the works in case it changes what you want to do here.

bewinsnw:
Moved that over to my work account: https://github.com/bewinsnw/argocd-image-updater/tree/bazzargh/yamlv3 , also fixed up the tests to be yaml/yaml comparisons not compare the Node tree, and tests showing this preserves comments, multiline strings, anchors, and follows anchors for setting values too.

Issue #821, opened 2024-08-09
Image updater updates the wrong helm values when having more than one chart in an Argocd Application 


emagiz:
**Describe the bug**
When having more than one helm chart deployed in the same Application file, the annotation `    argocd-image-updater.argoproj.io/XXXX.helm.image-name` does not update the properties for the correct chart, but for the first one.
**To Reproduce**
Application.yaml
`apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: name
  namespace: namespace
  annotations:
    argocd-image-updater.argoproj.io/image-list: XXXXX=
    argocd-image-updater.argoproj.io/XXXXX.platforms: linux/amd64
    argocd-image-updater.argoproj.io/XXXXX.pull-secret: pullsecret:
    argocd-image-updater.argoproj.io/XXXXX.update-strategy: semver
    argocd-image-updater.argoproj.io/XXXXX.allow-tags: regexp:^[0-9]+
    argocd-image-updater.argoproj.io/XXXXX.ignore-tags: latest
    argocd-image-updater.argoproj.io/XXXXX.helm.image-tag: YYYY.image.tag (YYYY should be the root yaml property in the values file of chart 2)
    argocd-image-updater.argoproj.io/XXXXX.helm.image-name: YYYY.image.repository
spec:
  project: 
  sources:
    - repoURL: ""
      path: main
      targetRevision: test
      helm:
        valueFiles:
          - $values-chart1/path
    - repoURL: ""
      targetRevision: test
      ref: values-chart1
    - repoURL: ""
      path: helm
      targetRevision: test
      helm:
        valueFiles:
          - $values-chart2/
    - repoURL: ""
      targetRevision: test
      ref: values-chart2
  destination:
    server: "https://kubernetes.default.svc"
    namespace: namespace
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
`

**Expected behavior**
`apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: name
  namespace: namespace
  annotations:
    argocd-image-updater.argoproj.io/image-list: XXXXX=
    argocd-image-updater.argoproj.io/XXXXX.platforms: linux/amd64
    argocd-image-updater.argoproj.io/XXXXX.pull-secret: pullsecret:
    argocd-image-updater.argoproj.io/XXXXX.update-strategy: semver
[...]

PR #1040, opened 2025-02-14
feat: add helm release helm param to fix multi helm sources updates

etienne-napoleone:
Introduces an implementation idea to fix the bug where multiple helm sources applications doesn't update the correct pod with the correct image.

The approach is to introduce a `argocd-image-updater.argoproj.io/%s.helm.release-name` label to specify which helm release is targeted by an alias.

I'm not sure it's the correct approach, but tested on my side and it seems to be working as expected.

related to #821

### DIFF
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -430,8 +430,60 @@ func SetHelmImage(app *v1alpha1.Application, newImage *image.ContainerImage) err
 		return fmt.Errorf("cannot set Helm params on non-Helm application")
 	}
 
-	appName := app.GetName()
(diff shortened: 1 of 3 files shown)


chengfang:
@etienne-napoleone thanks for the contribution. Can you address the CI failures while we're reviewing it?

etienne-napoleone:
Sorry, I might later when I have more bandwidth but might be in a while. If someone has the time to check if the approach seems good and make a good PR with it, it would be good for the project.

It's working as expected using a personal image, for both multi helm sources (with the helm-release param) and single helm source (without the param).

PR #1044, opened 2025-02-26
fix: fix KustomizeImage Match function to pass added unit tests

chengfang:
Sync up registry-scanner/pkg/image/kustomize.go with argo-cd


ishitasequeira (approved):
LGTM!!

PR #1055, opened 2025-03-03
Move to yaml.v3 and preserve comments, multiline strings, anchors in helm values

bewinsnw:
Similar to #1039 (and conflicts with it). In our codefresh install developers have complained that various parts of the original yaml aren't preserved:
- comments
- multiline string formatting for parameters like scripts
- anchors (the current code resolves all anchors and writes back with no anchors, resulting in duplicative config)

In particular we have cases where multiple jobs are defined in a chart which use the same images, which should all be updated via a single reference. yaml.v2 doesn't support this, but yaml.v3 does; the Node structure preserves pre-, post- and end-of-line comments, keeps the string formatting marker with the string for marshalling, and allows us to navigate aliases without losing them.

In this PR, I first switched to yaml.v3, then, like #1039, simplified the test cases for this code to not use the internal structures we unmarshal to but instead do yaml-to-yaml comparisons; this is easier to read and maintain and less code. Finally I added tests for the features described above.

I've only supported one level of aliasing; you can have multiple aliases through the path as you navigate the document but you can't have a mapping which is an alias for a mapping which is an alias for a mapping. Supporting that may mean the code could get stuck in a loop, and it doesn't seem like a sensible document structure anyway, just make the direct reference?

I don't currently support or set tags in the yaml, eg !!str. This would be an issue if you try to set  a field which is not tagged as a string.

Unlike #1039, this does not also update the path syntax used for updates.

### DIFF
--- a/pkg/argocd/update_test.go
+++ b/pkg/argocd/update_test.go
@@ -2335,6 +2335,112 @@ image:
 		assert.Error(t, err)
 		assert.Equal(t, "unexpected type ScalarNode for key attributes", err.Error())
 	})
+
+	t.Run("Aliases, comments, and multiline strings are preserved", func(t *testing.T) {
+		expected := `
+image:
+    attributes:
+        name: &repo repo-name
+        tag: v2.0.0
+        # this is a comment
+        multiline: |
+            one
+            two
+            three
+        alias: *repo
+`
+
+		inputData := []byte(`
+image:
+    attributes:
+        name: &repo repo-name
+        tag: v1.0.0
+        # this is a comment
+        multiline: |
+            one
+            two
+            three
+        alias: *repo
+`)
+		input := yaml.Node{}
+		err := yaml.Unmarshal(inputData, &input)
+		require.NoError(t, err)
+
+		key := "image.attributes.tag"
+		value := "v2.0.0"
+
+		err = setHelmValue(&input, key, value)
+		require.NoError(t, err)
+
+		output, err := yaml.Marshal(&input)
+		require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


jannfis:
Given this PR seems simpler than #1039, doesn't introduce a new dependency (just a version bump) and that there was no feedback on the review comments for about two weeks, I'd be fine to go with this one instead of #1039 for now.

@chengfang WDYT?

chengfang:
Looks promising. I'll take a closer look.

pkg/argocd/update_test.go
```diff
@@ -1179,10 +1179,10 @@ func Test_MarshalParamsOverride(t *testing.T) {
...
 		expected := `
 kustomize:
-  images:
-  - baz
-  - foo
-  - bar
+    images:
+        - baz
+        - foo
+        - bar
```
chengfang, line 1185: Looks like yaml.v3 always uses 4-space indentation. For existing helm values files using 2-space indentation, it will rewrite to 4 spaces. This adds some unnecessary space-only changes. Is there a way to configure it to honor the existing indentation?

bewinsnw: See the latest changes. I've set it up to marshal with 2 spaces; this still has some differences from v2 which I can't control:
```
this:
- sequence
```
becomes:
```
this:
  - sequence
```

I've also added a change to guess an preserve the indent where possible (ie only for Helm, since you have to Unmarshal to a Node to get told the columns), so eg:
```
this:
   has: 3 spaces and will be preserved
```
It's best effort, so I only guess for the 2 easiest cases (top level is a mapping inside a mapping, or top level is a sequence inside a mapping and the sequence is indented)

chengfang (approved):

Issue #932, opened 2024-11-11
Wrong default write branch is used in multi-source app


chengfang:
**Describe the bug**
In a multi-source application, when a custom write-back git repo is specified, but the git branch is not specified, image-updater selects the branch from the primary source, instead of the targetRevison of the chosen source.

**To Reproduce**
```yaml
    argocd-image-updater.argoproj.io/nginx.helm.image-name: image.repository
    argocd-image-updater.argoproj.io/nginx.helm.image-tag: image.tag
    argocd-image-updater.argoproj.io/git-repository: https://github.com/chengfang/image-updater-examples.git
    # argocd-image-updater.argoproj.io/git-branch: main
    argocd-image-updater.argoproj.io/write-back-method: git:secret:argocd/git-creds
    argocd-image-updater.argoproj.io/write-back-target: helmvalues:/write-helmvalues/source2/values.yaml
```

**Expected behavior**
the branch of the specified source repo is selected as the default write-back git branch, unless overridden with `git-branch` annotation.

**Additional context**
The workaround is to always specify `git-branch` annotation. 

**Version**
v0.15.1

**Logs**
```text
INFO[0001] Committing 1 parameter update(s) for application write-helmvalues  application=write-helmvalues
INFO[0001] Initializing https://github.com/chengfang/image-updater-examples.git to /var/folders/sc/bblzy4t15995h2zs1s4bztc00000gn/T/git-write-helmvalues2791939777
INFO[0001] git fetch origin 18.2.3 --force --prune --depth 1  dir=/var/folders/sc/bblzy4t15995h2zs1s4bztc00000gn/T/git-write-helmvalues2791939777 execID=a6d31
ERRO[0002] `git fetch origin 18.2.3 --force --prune --depth 1` failed exit status 128: fatal: couldn't find remote ref 18.2.3  execID=a6d31
INFO[0002] Trace                                         args="[git fetch origin 18.2.3 --force --prune --depth 1]" dir=/var/folders/sc/bblzy4t15995h2zs1s4bztc00000gn/T/git-write-helmvalues2791939777 operation_name="exec git" time_ms=284.635333
ERRO[0002] Could not update application spec: `git fetch origin 18.2.3 --force --prune --depth 1` failed exit status 128: fatal: couldn't find remote ref 18.2.3  application=write-helmvalues
INFO[0002] Processing results: applications=1 images_considered=1 images_skipped=0 images_updated=0 errors=1
```

[...]

PR #1056, opened 2025-03-05
fix: use correct default branch in multi-source applications

aali309:
This PR fixes: #932


pkg/argocd/argocd.go
```diff
@@ -675,17 +675,34 @@ func getApplicationSourceType(app *v1alpha1.Application) v1alpha1.ApplicationSou
 	return app.Status.SourceType
 }
 
-// getApplicationSource returns the main source of a Helm or Kustomize type of the application
+// getApplicationSource returns the source matching git-repository annotation
+// or the Helm/Kustomize source, or the primary source if no match is found
 func getApplicationSource(app *v1alpha1.Application) *v1alpha1.ApplicationSource {
```
chengfang, line 680: Changing this func will impact many places, and in some of these places, we do want to get the primary source, not the one for write-back. I think we should avoid changing this func, and instead modify git.go

pkg/argocd/argocd.go
```diff
@@ -675,17 +675,34 @@ func getApplicationSourceType(app *v1alpha1.Application) v1alpha1.ApplicationSou
...
+	// Get the git repository from annotations for write-back
+	gitRepo := app.Annotations["argocd-image-updater.argoproj.io/git-repository"]
+	logCtx := log.WithContext().AddField("application", app.GetName())
```
chengfang, line 683: when getting annotation values, we should follow the pattern used in other places, such as:
```
// Check whether application has our annotation set
		annotations := app.GetAnnotations()
		if _, ok := annotations[common.ImageUpdaterAnnotation]; !ok {

```

chengfang:
the e2e test is passing because it's using the default argocd writeback, not the git write-back. I think we can skip e2e test for this issue, and some unit tests should suffice.

pkg/argocd/git.go
```diff
@@ -128,6 +129,36 @@ func TemplateBranchName(branchName string, changeList []ChangeEntry) string {
...
 
+// getWriteBackBranch returns the branch to use for write-back operations.
+// It first checks for a branch specified in annotations, then uses the
+// targetRevision from the matching git source, falling back to getApplicationSource.
+func getWriteBackBranch(app *v1alpha1.Application) string {
+	// First check if branch is explicitly set in annotations
+	annotations := app.GetAnnotations()
+	if branch, ok := annotations[common.GitBranchAnnotation]; ok {
+		return branch
+	}
```
chengfang, line 140: we cannot take the annotation value as is, since it can be a composite form like "main:changes". See https://argocd-image-updater.readthedocs.io/en/latest/basics/update-methods/#specifying-a-branch-to-commit-to

I guess we dont' see the above block. Line 200 already handles this case. See next comment.

pkg/argocd/git.go
```diff
@@ -164,7 +195,7 @@ func commitChangesGit(app *v1alpha1.Application, wbc *WriteBackConfig, changeLis
 	// config, or taken from the application spec's targetRevision. If the
 	// target revision is set to the special value HEAD, or is the empty
 	// string, we'll try to resolve it to a branch name.
-	checkOutBranch := getApplicationSource(app).TargetRevision
+	checkOutBranch := getWriteBackBranch(app)
 	if wbc.GitBranch != "" {
 		checkOutBranch = wbc.GitBranch
 	}
```
chengfang, line 201: if the gitbranch is set, we always take it. So I think we only need to call the func `getWriteBackBranch` if wbc.GitBranch is empty (in the else clause).

chengfang (approved):

PR #1058, opened 2025-03-06
fix: determine the suitable github api base url from repo url

chengfang:
Fix the issue raised in discussion https://github.com/argoproj-labs/argocd-image-updater/discussions/1038

When the github enterprise api base url is not set, try to determine it based on repo url.

### DIFF
--- a/ext/git/creds.go
+++ b/ext/git/creds.go
@@ -42,7 +42,6 @@ const (
 	// githubAccessTokenUsername is a username that is used to with the github access token
 	githubAccessTokenUsername = "x-access-token"
 	forceBasicAuthHeaderEnv   = "ARGOCD_GIT_AUTH_HEADER"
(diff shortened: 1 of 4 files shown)


jannfis:
Since we tend to copy over the Git client's code from Argo CD in a more or less verbatim way, and then enrich it with code in writer.go, is there any possibility that we could outsource these changes into a separate file?

chengfang:
Shall we move the added func to a new file like `creds2.go` and test to `creds2_test.go`, or move it to the existing `writer.go` to avoid adding 2 short new files?

chengfang:
Okay, I moved `g.getBaseUrl()` to a separate file creds2.go and test to creds2_test.go to minimize changing creds.go

chengfang:
Will look into how to fix it in argo-cd instead.

PR #1062, opened 2025-03-12
fix: use compact indent for sequences

bewinsnw:
We had some grumbles from developers after the goyaml.v3 upgrade that this changed the compact sequence indent used by kustomize, kubectl etc.

There was a whole thread about this when the goyaml.v3 upgrade happened in kustomize itself: https://github.com/kubernetes-sigs/kustomize/issues/3946 and the result was that kubesigs forked goyaml.

In order to more closely match what the kubernetes tools do, this commit switches to the forked version, which has a compact option on the encoder:
https://github.com/kubernetes-sigs/yaml/blob/56d672052dcff7362af5b66f6424976539cddd78/goyaml.v3/patch.go#L25-L28

When I first tried this I also changed the code to guess if compact was the preferred style from the yaml. However, it turns out that the yaml formatter misbehaves when told to use "compact" and an indent other than 2; if set to 3 and compact it produces code like this:

    mapping:
     - list item

Instead of:

    mapping:
    - list item

This meant I couldn't get the "guess indentation" code to work, so I've removed that and its test to avoid raising a false expectation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,12 +24,12 @@ require (
 	golang.org/x/oauth2 v0.27.0
 	golang.org/x/sync v0.11.0
 	google.golang.org/grpc v1.70.0
-	gopkg.in/yaml.v3 v3.0.1
 	k8s.io/api v0.31.0
 	k8s.io/apimachinery v0.31.0
 	k8s.io/client-go v0.31.0
 	sigs.k8s.io/kustomize/api v0.17.2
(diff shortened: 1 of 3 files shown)


chengfang:
This is a follow-up to previous PR https://github.com/argoproj-labs/argocd-image-updater/pull/1055

chengfang (approved):

PR #1066, opened 2025-03-13
Fix multi images e2e test.

dkarpele:
1. Use public images `nginx` and `memcached` to test update for the pod with 2 containers. 

We can't use local docker registry for images named `<ip>:<port>/<image_name>:<tag>` because ArgoCD evaluates all images that starts with `<ip>:<port>` as the same images regardless `<image_name>`.

The [function](https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/types.go#L615C1-L621C2) in ArgoCD that works wrong in this case is:
```go
// Match returns true if the image name matches (i.e. up to the first delimiter)
func (i KustomizeImage) Match(j KustomizeImage) bool {
	delim := j.delim()
	imageName, _, _ := strings.Cut(string(i), delim)
	otherImageName, _, _ := strings.Cut(string(j), delim)
	return imageName == otherImageName
}
```

The question whether to fix the problem above or not is still open.


chengfang:
can we configure the internal registry as the default, and omit the url in deployment.yaml?

PR #1068, opened 2025-03-17
[GITOPS-5120] Add e2e test for write-back GIT repository as annotation.

dkarpele:
Add e2e test for write-back GIT repository as annotation.
1. 007-git-repository: use local kustomize app with local image. Annotations:
- `write-back-method: git`
- `git-repository:local_repo`

`03-run-updater-revert.yaml` reverts the changes made to local git repo to make all tests consistent. Without it the next run of this test or any other test that uses `001-simple-kustomize-app` will fail.

2. 008-git-repository-helm: use local helm app with public image. Annotations:
- `write-back-method: git`
- `git-repository:local_repo`.


test/e2e/suite/008-git-repository-helm/README.md
```diff
@@ -0,0 +1,195 @@
+This test case verifies [possibility to specify write-back GIT Helm repository as annotation](https://argocd-image-updater.readthedocs.io/en/latest/basics/update-methods/#specifying-a-repository-when-using-a-helm-repository-in-repourl).
+
+This test case performs the following steps:
+* install ArgoCD Helm Application to the test cluster with GIT write-back method
+* run argocd-image-updater
+* assert that the expected image was installed
+* uninstall Application from the test cluster
```
chengfang, line 7: any reason this test does not undo the changes to `008-simple-helm-app/.argocd-source-image-updater-008.yaml`, similar to 007-git-repository?

dkarpele: The difference between test 7 and 8 is in `argocd-image-updater.argoproj.io/write-back-target: kustomization` in test 7. Test 7 changes `001-simple-kustomize-app/kustomization.yaml` file itself while test 8 adds file `.argocd-source-image-updater-008.yaml`. 

Test 8 was originally copied from Test 2. Test 2 also doesn't have revert git status back step. Because of the file name (with 008) it doesn't break other tests in the test suite. I can work on the implementation of this step for Test 8 and Test 2 in this PR.

dkarpele: So, I added a revert step only to Test 8. 
Test 2 doesn't check image name equality that's why Test 2 passes always - even after retry. So reverting commit for Test 2 is just not necessary.

PR #1078, opened 2025-03-29
chore(deps): Upgrade argo-cd version to 2.13.6

aali309:
This change Upgrade argo-cd version from 2.13.5 to 2.13.6

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/argoproj-labs/argocd-image-updater
 go 1.23.5
 
 require (
-	github.com/argoproj-labs/argocd-image-updater/registry-scanner v0.0.0-20250114144021-10a01f381aa7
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -3,8 +3,8 @@ module github.com/argoproj-labs/argocd-image-updater
 go 1.23.5
 
 require (
-	github.com/argoproj-labs/argocd-image-updater/registry-scanner v0.1.0
-	github.com/argoproj/argo-cd/v2 v2.13.5
+	github.com/argoproj-labs/argocd-image-updater/registry-scanner v0.0.0-20250114144021-10a01f381aa7
```
chengfang, line 6: we should use registry-scanner v0.1.0, which is the real release, instead of the pseudo one.

aali309: Yes, this change was brought in after running `go mod tidy`

chengfang (approved):

Issue #1076, opened 2025-03-27
Support setting `--interval` via env variable


perlun:
**Is your feature request related to a problem? Please describe.**
At the moment, setting the check interval requires editing the `spec.containers.args` setting in the Deployment for `argocd-image-updater`. It would be more convenient to be able to update this using the `ConfigMap`

**Describe the solution you'd like**
Many settings are currently settable via environment variables:

https://github.com/argoproj-labs/argocd-image-updater/blob/b2472651c1a75399499d60d9d69e0b05d9f1074c/cmd/run.go#L221-L247

_However_, this particular setting is only available as a command line parameter to the argocd-image-updater binary:

https://github.com/argoproj-labs/argocd-image-updater/blob/b2472651c1a75399499d60d9d69e0b05d9f1074c/cmd/run.go#L228

The reason why it would be good to have this be settable using an environment variable is that we can then use a `ConfigMap` to edit it, like we do with other environment variables: https://github.com/argoproj-labs/argocd-image-updater/blob/master/manifests/install.yaml. This would make it much easier to customize the interval when needed.

**Describe alternatives you've considered**
None than I can think of, apart from editing the deployment manually - perhaps using Kustomize to modify the manifest when installing argocd-image-updater or something?

chengfang:
This does sound like a great enhancement.

chengfang:
@dkarpele please take a look.

perlun:
Sorry for the lack of replies from my end @chengfang, great to see that this has now been added! 🎉🙇

PR #1107, opened 2025-04-18
feat: Support setting `run --interval` via env variable.

dkarpele:
This commit introduces new env variable `IMAGE_UPDATER_INTERVAL` for `run --interval` option. It will make it easier for users to configure update cycles, along with many other parameters that can already be set via either environment variable or command line arguments.

Closes: #1076


dkarpele:
Checks fail because they still use [registry-scanner@v0.1.0](https://pkg.go.dev/github.com/argoproj-labs/argocd-image-updater/registry-scanner@v0.1.0/pkg/env) which doesn't have updated `env.go` and `env_test.go`

docs/install/reference.md
```diff
@@ -94,6 +94,8 @@ is given, milliseconds will be assumed. If set to `0`, ArgoCD Image Updater
 will exit after the first run, effectively disabling the interval. Default
 value is `2m0s`.
 
+Can also be set using the *IMAGE_UPDATER_INTERVAL* environment variable.
+
```
chengfang, line 98: I think we should descriibe the order the precedence here: if the param is not set, then check the env var, and if neither the param or env var is set, then fall back to the default value 2m.

registry-scanner/pkg/env/env.go
```diff
@@ -35,6 +36,18 @@ func GetStringVal(envVar string, defaultValue string) string {
...
 }
 
+// GetDurationVal retrieves a time.Duration value from given environment envVar
+// Returns default value if envVar is not set.
+func GetDurationVal(envVar string, defaultValue time.Duration) time.Duration {
+	if val := os.Getenv(envVar); val != "" {
+		duration, err := time.ParseDuration(val)
+		if err == nil {
+			return duration
+		}
```
chengfang, line 46: maybe we should log a warning for invalid value?

chengfang:
> Checks fail because they still use [registry-scanner@v0.1.0](https://pkg.go.dev/github.com/argoproj-labs/argocd-image-updater/registry-scanner@v0.1.0/pkg/env) which doesn't have updated `env.go` and `env_test.go` with `GetDurationVal`

We can separate registry-scanner changs into a separate PR, and once it is merged, change image-updater go.mod to reference a psudo-version of registry-scanner pointing to the commit sha of the first PR.

dkarpele:
> > Checks fail because they still use [registry-scanner@v0.1.0](https://pkg.go.dev/github.com/argoproj-labs/argocd-image-updater/registry-scanner@v0.1.0/pkg/env) which doesn't have updated `env.go` and `env_test.go` with `GetDurationVal`
> 
> We can separate registry-scanner changs into a separate PR, and once it is merged, change image-updater go.mod to reference a psudo-version of registry-scanner pointing to the commit sha of the first PR.

Separate PR: https://github.com/argoproj-labs/argocd-image-updater/pull/1110

registry-scanner/pkg/env/env.go
dkarpele: Yep, please see #1110

chengfang (approved):

Issue #902, opened 2024-10-30
Fails to list applications when installed by helm-chart and createClusterRoles not enabled


nabadger:
**Describe the bug**
I've upgraded from 0.12.0 to 0.15.0.

Previously no cluster roles were required.

Upon startup, i get the following errors

```
ttime="2024-10-30T12:53:25Z" level=error msg="error while communicating with ArgoCD" argocd_server=argocd-server.argocd grpc_web=true grpc_webroot= insecure=false plaintext=false
time="2024-10-30T12:53:25Z" level=error msg="Error: error listing applications: applications.argoproj.io is forbidden: User \"system:serviceaccount:argocd:argocd-image-updater\"
```

This is when running in kubernetes-api mode (so the initial complaints about ArgoCD are also strange...)

**To Reproduce**

Run the latest image against official helm chart and ensure `createClusterRoles: false`.

**Expected behavior**

I would expect the role-bindings shipped with the helm chart should work. I'm also curious to know why the clusterrole needs update/patch? 

**Additional context**

I'm not sure whether this is a helm-chart issue, or whether the actual code should support only handling applications from the deployed namespace (in our case `argocd`). All of our application resources are in this namespace. 

Ideally it would be nice to have the option here. That still might be covered by the RBAC though rather than app-code, so could still be a helm issue.

I believe the error comes from https://github.com/argoproj-labs/argocd-image-updater/blob/a5acd25853dd19a5186c1cc3801bf9e5cfa052dc/pkg/argocd/argocd.go#L35
and is a result of https://github.com/argoproj-labs/argocd-image-updater/pull/854

**Version**
v.015.0  - was not present in v0.12.0

borja00:
I'm experiencing the same problem, our current setup does not allowed to create Cluster scope roles and bindings. It would be great if cross namespace application support could be optional.

mysiki:
+1

fad3t:
Same issue here, we're deploying instances of image-updater per namespace, and we'd like to keep that level of isolation between namespaces.

If this is related to #854, then we might need a flag that tells image-updater whether it's namespaced or not.

Issue #1111, opened 2025-04-22
func (client *k8sClient) ListApplications() in v1.NamespaceAll:  kube-api CPU usage on big installations (~8K applications)


amorozkin:
Hi!

1. Since v0.15.0 calling [GetApplication()](https://github.com/argoproj-labs/argocd-image-updater/blob/release-0.15/pkg/argocd/argocd.go#L31)  triggers in turn [ListApplications()](https://github.com/argoproj-labs/argocd-image-updater/blob/release-0.15/pkg/argocd/argocd.go#L62) in v1.NamespaceAll.
2. GetApplication() is called multiple times in row.

These all caues very high resource (CPU mainly) usage on k8s masters as a result.
In our case (Muilt-cluster Argocd instance v2.14.10 + 8K argоcd apps) it looks like that on 6 CPU core masters (dedicated argocd k8s cluster, v1.29):
 
<img width="408" alt="Image" src="https://github.com/user-attachments/assets/def96d4a-d3bf-4635-bdd7-ea0d0ff24e60" />

K8S API logs:
<img width="823" alt="Image" src="https://github.com/user-attachments/assets/2dbacb52-777c-4f52-9386-83652c710a16" />

We are not using "Application in any namespase" feature currently so it's ok for us to switch back to using v0.14.0.

As of 0.15.0+ versions, may be there is a need for optimization - say adding redis cache on application side (with 1 min TTL for ListApplications() for example) ?

chengfang:
https://github.com/argoproj-labs/argocd-image-updater/pull/1112 tries to fix this issue by introducing a new param `AppNamespace` so that image-updater will only searches apps in that namespace. Will that help mitigate the performance issue your'are seeing?

amorozkin:
No, I think it wouldn't in the way it's going to be implemented.
Root cause of performance issue is here - GetApplication() now uses List() instead of  Get(.., appName ) as version 0.14- did.
Getting full applications list every time when you have to get just a particlular one is a very expensive operation on "big" argocd installations.

https://github.com/argoproj-labs/argocd-image-updater/pull/1112 still uses List().

If there is no way to roll back to using Get() app by its name - i see only two optimization options:

1. Adding cache for ListApplications() function as already been propoused
2. Refactor the code - add HahdleApplicationList() function that would be called periodically and would get apps list returned by ApplicationList() as its argument and then in the loop would process every app. In this way there would be no need to make List() applications every time you have to get just one of them.

[...]

PR #1112, opened 2025-04-24
feat: Add configuration option to scope application listing / getting to particular namespace

grahamalama:
In https://github.com/argoproj-labs/argocd-image-updater/pull/854, AIU made changes so that it scanned applications across all namespaces. This introduced a few regressions for some use cases, namely https://github.com/argoproj-labs/argocd-image-updater/issues/902.

In this PR, I introduce a config option so that users can optionally scope the K8s client to a specific namespace (as was the case before https://github.com/argoproj-labs/argocd-image-updater/pull/854). Since searching across all namespaces was recently introduced as the default, I opted to keep that the default.


grahamalama:
I was unsure how to add tests for this change. I saw [these tests](https://github.com/argoproj-labs/argocd-image-updater/blob/61ea007f2eb8b245270a5fde7e758c994b8ccdc3/pkg/argocd/argocd_test.go#L1013), but I'm not sure how to set up something similar to assert that "if the client is configured with an `appNamespace`, it will _only_ search for that namespace". Maybe asserting the fake client was called with certain parameters? Otherwise I think we'd end up testing the mock.

chengfang:
@grahamalama thanks for the contribution. Can you please fix the DCO check failure? Is this PR also intended to address issue https://github.com/argoproj-labs/argocd-image-updater/issues/1111 ?

grahamalama:
@chengfang just pushed the DCO fix.

> Is this PR also intended to address issue https://github.com/argoproj-labs/argocd-image-updater/issues/1111?

It could be I suppose! For me, my goal was to fix https://github.com/argoproj-labs/argocd-image-updater/issues/902 specifically. But I think it'd address both.

chengfang:
> I was unsure how to add tests for this change. I saw [these tests](https://github.com/argoproj-labs/argocd-image-updater/blob/61ea007f2eb8b245270a5fde7e758c994b8ccdc3/pkg/argocd/argocd_test.go#L1013), but I'm not sure how to set up something similar to assert that "if the client is configured with an `appNamespace`, it will _only_ search for that namespace". Maybe asserting the fake client was called with certain parameters? Otherwise I think we'd end up testing the mock.

can we add tests with non-nil appNamespace, just for some basic sanity check?

chengfang:
@grahamalama can you also update docs/install/reference.md to document the new flag? (https://argocd-image-updater.readthedocs.io/en/latest/install/reference/

chengfang:
@grahamalama can you also take a look at the comment on the related issue https://github.com/argoproj-labs/argocd-image-updater/issues/1111#issuecomment-2840828900. Although it covers a different issue (still related to this one), I think the comment about using listApp vs getApp is also revelant here. Some namespace may contain large number of apps and list operation in this namespace could still be a performance bottleneck. If we know image-updater is configured to 1 namespace, we should be able to just use getApp.

grahamalama:
> @grahamalama can you also take a look at the comment on the related issue [#1111 (comment)](https://github.com/argoproj-labs/argocd-image-updater/issues/1111#issuecomment-2840828900). Although it covers a different issue (still related to this one), I think the comment about using listApp vs getApp is also revelant here. Some namespace may contain large number of apps and list operation in this namespace could still be a performance bottleneck. If we know image-updater is configured to 1 namespace, we should be able to just use getApp.

Commented in #1111, and depending on the outcome of that conversation I'll update this PR accordingly. I will say though if the decision is to add some sort of caching option, I think that would be better suited for a separate PR.

chengfang:
@grahamalama from the discussion here and in #1111, we are in agreement to improve this PR by using getApp instead of listApp when single namespace is configured in image-updater.

grahamalama:
@chengfang, made the changes we discussed in #1111 and added the requested documentation

pkg/argocd/argocd.go
```diff
@@ -25,13 +25,21 @@ import (
...
-	// List all applications across all namespaces (using empty labelSelector)
-	appList, err := client.ListApplications(v1.NamespaceAll)
+	// List all applications across configured namespace or all namespaces (using empty labelSelector)
+	if *client.appNamespace != v1.NamespaceAll {
+		return client.kubeClient.ApplicationsClientset.ArgoprojV1alpha1().Applications(*client.appNamespace).Get(ctx, appName, v1.GetOptions{})
+	}
+	return client.getApplicationInAllNamespaces(ctx, appName)
+}
+
+func (client *k8sClient) getApplicationInAllNamespaces(ctx context.Context, appName string) (*v1alpha1.Application, error) {
```
chengfang, line 41: the param `ctx` is not used in this new function. I think we should remove it.

docs/install/reference.md
```diff
@@ -26,6 +26,10 @@ Runs the Argo CD Image Updater, possibly in an endless loop.
 
 ### Flags
 
+**--application-namespace *namespace***
+
+Specifies the Kubernetes namespace in which Argo Image Updater will manage Argo CD Applications when using the Kubernetes-based Application API. By default, applications in all namespaces are considered. This flag can be used to limit scope to a single namespace for performance, security, or organizational reasons.
```
chengfang, line 31:
```suggestion
Specifies the Kubernetes namespace in which Argo CD Image Updater will manage Argo CD Applications when using the Kubernetes-based Application API. By default, applications in all namespaces are considered. This flag can be used to limit scope to a single namespace for performance, security, or organizational reasons.
```

grahamalama: [3acfb8d](https://github.com/argoproj-labs/argocd-image-updater/pull/1112/commits/3acfb8d7a7558d7cf3a6874a87a8d9a0292f15fb)

pkg/argocd/argocd.go
grahamalama: [eab19bd](https://github.com/argoproj-labs/argocd-image-updater/pull/1112/commits/eab19bdef06599e643a5f0613c537bf50c6c88ff)

chengfang (approved):

PR #1129, opened 2025-05-10
Fix dependencies after applying operator sdk

dkarpele:
Also fix Dockerfile to make it build argoproj.io/image-updater image.

Commands that work:
```
make manifest
make generate
make build
make install
make run
make docker-build
make deploy
make undeploy
```

To upload to local cluster after docker build I use:
```
docker save argoproj.io/image-updater:0.0.1 -o image-updater.tar
sudo k3s ctr images import image-updater.tar
```

Some flags are commented now because they are not implemented yet.


Dockerfile
```diff
@@ -12,16 +12,18 @@ COPY go.sum go.sum
...
+COPY ext/ ext/
+COPY pkg/ pkg/
 
 # Build
 # the GOARCH has not a default value to allow the binary be built according to the host where the command
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager ./cmd
```
chengfang, line 27: This Dockerfile is quite different from the one in master. I think we'll need to keep it close to the master as much as we can in order to make it easier to merge into master in the future.

dkarpele: To avoid misunderstanding, let's keep the original Dockerfile as is and I created a new controller.Dockerfile that can be used later to run a controller.

chengfang: now that we're not using the original Dockerfile for the controller, shall we restore `Dockerfile` to its original content prior to the start of crd work?

dkarpele: hm, the Dockerfile in this PR is exactly the same as Dockerfile in master https://github.com/argoproj-labs/argocd-image-updater/blob/master/Dockerfile. It differs from the current version in crd branch but the same as in master.

chengfang (approved):

PR #1157, opened 2025-06-03
chore: improve registry-scanner/hack/create-release.sh

cjcocokrisp:
### Overview 
The following PR improves the registry-scanner/hack/create-release.sh with the following:
- Switching to annotated tags instead of using lightweight tags
- Automatically push the changed `VERSION` file in addition to the new tag to remote
- Added some usage examples to the header
- Improved help messages for missing arguments

I also noticed that the `VERSION` file's contents was `0.111.0`. Not sure if this was intentional because the tag listed is `0.1.0`. I changed the file to match the tag. If it should stay `0.111.0`, I can revert the change. 

### Testing: 
The below screenshots show the testing that I did to validate that my changes functioned properly. 

The output of `git tag -n registry-scanner/*` after running `./hack/create-release.sh TEST HelloHello` and `./hack/create-release.sh TEST2 HelloHello2`:
Screenshot 2025-06-03 at 3 04 31 PM

The commits and tags in the remote after running those two commands above:
Screenshot 2025-06-03 at 3 05 52 PM
Screenshot 2025-06-03 at 3 06 28 PM

The error messages output if you are missing either the `TARGET_VERSION` or `MESSAGE` arguments:
Screenshot 2025-06-03 at 3 08 33 PM

### DIFF
--- a/registry-scanner/hack/create-release.sh
+++ b/registry-scanner/hack/create-release.sh
@@ -63,8 +63,7 @@ echo "${TARGET_VERSION}" > VERSION
 
 # Commit updated VERSION file
 git add VERSION
-git commit -s -m "Release ${NEW_TAG}" 
-git push "${REMOTE}" "${RELEASE_BRANCH}"
+git commit -s -m "${MESSAGE}" 
 
 # Create tag for registry-scanner
 git tag -a "${NEW_TAG}" -m "${MESSAGE}"


registry-scanner/VERSION
```diff
@@ -1 +1 @@
-0.111.0
+0.1.0
```
chengfang: This is intentionally kept an unreal value. The real value is set in a release-* branch. See 
https://github.com/argoproj-labs/argocd-image-updater/blob/release-0.16/VERSION
https://github.com/argoproj-labs/argocd-image-updater/blob/registry-scanner/release-0.1/registry-scanner/VERSION

registry-scanner/hack/create-release.sh
```diff
@@ -4,29 +4,48 @@
...
 # - merge the PR
+# Example Uses:
+# `./hack/create-release.sh 0.1.1 "This is an example message"` Would create a new tag "registry-scanner/v0.1.1" with
+#                                                               the message "This is an example message" and edit 
+#                                                               VERSION to be 0.1.1 which would be committed.
+#
+# `./hack/create-release.sh 0.1.X "TEST"` origin                Would create a new tag "registry-scanner/v0.1.X" with
+#                                                               the message "TEST" and edit VERSION to be 0.1.X which 
+#                                                               would be committed. The contents would be pushed to the
+#                                                               remote "origin."
```
chengfang: Having a message gives more control for users to custom a tag and commit message, but at the same time also makes it slightly harder to run this command. I think the tag and commit message are pretty standard, using the tag name should suffice.

registry-scanner/hack/create-release.sh
```diff
@@ -4,29 +4,48 @@
...
 SUBMODULE_NAME="registry-scanner"
 
-if [[ ! "${CURRENT_BRANCH}" == registry-scanner/release-* ]]; then
-	echo "!! Please checkout branch 'registry-scanner/release-X.Y' (currently in branch: '${CURRENT_BRANCH}')" >&2
-	exit 1
-fi
+#if [[ ! "${CURRENT_BRANCH}" == registry-scanner/release-* ]]; then
+#	echo "!! Please checkout branch 'registry-scanner/release-X.Y' (currently in branch: '${CURRENT_BRANCH}')" >&2
+#	exit 1
+#fi
```
chengfang: the above validation is useful and should be preserved.

registry-scanner/hack/create-release.sh
```diff
@@ -47,7 +66,11 @@ fi
 echo "Creating tag ${NEW_TAG}"
 echo "${TARGET_VERSION}" > VERSION
 
+# Commit updated VERSION file
+git add VERSION
+git commit -s -m "automated: bump VERSION file to version ${TARGET_VERSION}"
```
chengfang: I think we should use the conventional commit message: "Release ${NEW_TAG}"

registry-scanner/hack/create-release.sh
```diff
@@ -47,7 +66,11 @@ fi
...
 echo "${TARGET_VERSION}" > VERSION
 
+# Commit updated VERSION file
+git add VERSION
+git commit -s -m "automated: bump VERSION file to version ${TARGET_VERSION}" 
+
 # Create tag for registry-scanner
-git tag "${NEW_TAG}"
-git push "${REMOTE}" tag "${NEW_TAG}"
+git tag -a "${NEW_TAG}" -m "${MESSAGE}"
```
chengfang, line 69: message can be the same as ${NEW_TAG}

registry-scanner/hack/create-release.sh
```diff
@@ -47,7 +66,11 @@ fi
...
+git push "${REMOTE}" --follow-tags #tag "${NEW_TAG}"
```
chengfang: the trailing comment seems redundant.

since we're usually directly pushing to upstream, I'd like to be specific about which tags and branches, something like:

git push "${REMOTE}" ${RELEASE_BRANCH} ${NEW_TAG}

Will that achieve the same effect?

registry-scanner/hack/create-release.sh
cjcocokrisp: Oops forgot to uncomment when I was testing 😅

registry-scanner/hack/create-release.sh
cjcocokrisp: That flag prevents you from needing to do a git push for the VERSION file commit and then needing to do another git push to commit the tag. It can be split up no problem but I left the comment as a note of what was originally was there. Which would be preferred the two separate pushes or just the one?

PR #1161, opened 2025-06-06
fix: remove dummy digest and add comparison logic

chansuke:
Removed dummy digest for workaround add comparison logic


chengfang:
@chansuke Can you describe the problem or bug caused by the current code?

chansuke:
@chengfang 
sorry for the late reply :pray: 
the current code mutates image state by setting a "dummy" digest value as a workaround to force updates when using digest strategy with tag-based images, which is a hack that could cause unexpected side effects

chengfang (approved):

PR #1163, opened 2025-06-08
chore: use context.Background() instead of context.TODO()

chansuke:
Both functions yield the same empty, never-cancelled context, using `context.Background()` clarifies intent:

- `context.Background()` clearly conveys that the call is rooted at a top-level context that is never cancelled or timed-out
- new contributors no longer wonder whether cancellation was forgotten; they see an explicit, canonical root context everywhere

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/argoproj-labs/argocd-image-updater
 go 1.23.5
 
 require (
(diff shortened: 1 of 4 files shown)


chengfang:
@chansuke can you describe why the changes are needed in the description section?

chansuke:
@chengfang 
Thank you so much for your commend and sorry for the late reply. I updated the description.

PR #1166, opened 2025-06-14
Preserve formatting in Helm values files

grahamalama:
Welp, I spent quite some time writing this PR to preserve formatting in Helm values files only to discover that there already some activity in this area, namely https://github.com/argoproj-labs/argocd-image-updater/pull/1039 and [this branch](https://github.com/bewinsnw/argocd-image-updater/commits/bazzargh/yamlv3/) referenced in that PR.

You can consider this another implementation and maybe decide between the two approaches. There's still more polish I could add, but since I discovered #1039, I thought I'd get this PR up to see if I should keep working on it.

Whether this PR or #1039, it'd be nice to get this general feature of preserving the formatting of Helm values files in.


chengfang:
Thanks for the contribution, @grahamalama. PRs https://github.com/argoproj-labs/argocd-image-updater/pull/1055 and https://github.com/argoproj-labs/argocd-image-updater/pull/1062, based on #1039, were merged and included in v0.16.0.

grahamalama:
@chengfang ahh I see. So it seems the final version of the formatting changes you mentioned do not use `goccy/go-yaml`? One consequence of this is that [blank lines are not preserved](https://github.com/go-yaml/yaml/issues/709). I've also noticed that image updater (whether through the yaml library or our logic) enforces this opinionated formatting of array items:

```
array:
- item
```
vs something like:
```
array:
  - item
```

[goccy/go-yaml](https://github.com/goccy/go-yaml) seems to handle both of these things, so I see some benefit of adopting that library as it would lead to overall less noisy diffs when applying updates.

What should I do with this PR? I'm sure it's not desirable to have 2 different yaml libraries in use simultaneously in the project, but if using `goccy/go-yaml` is something you want to consider further, I can continue working on this PR so that we switch everything over to `goccy/go-yaml` (or maybe create a separate follow-on PR to handle that). What do you think?

bewinsnw:
@grahamalama just noticed this one, I wrote a couple of the PRs mentioned above. The indentation of sequences was a deliberate choice in goyaml v3: see https://github.com/go-yaml/yaml/issues/661 and the many, many related discussions. Partly this goes back to, IIRC, goyaml having jumped the _other_ way in v2, breaking the formatting used in k8s, which then led to the entire k8s ecosystem relying on a fork of goyaml to avoid - you guessed it, noisy diffs.

grahamalama:
I'm going to close this PR and reopen it where:
- I fix conflicts
- I go all in with github.com/goccy/go-yaml so we don't have multiple yaml libraries

PR #1134, opened 2025-05-16
Create full CRD structure

dkarpele:
Important highlights:
1. `metadata.namespace` is a namespace for ImageUpdater CR.
2. 'spec.namespace' is a namespace where the controller will look for Argo CD Applications matching the criteria in `ApplicationRefs`. 'spec.namespace' **is required**. It means that all Applications in each CR will be assigned to one and only one namespace. 
3. `ApplicationRefs`, `images` are required. 

Full CRD version is `config/crd/bases/argocd-image-updater.argoproj.io_imageupdaters.yaml`

Short version:

```yaml
apiVersion: argocd-image-updater.argoproj.io/v1alpha1 
kind: ImageUpdater 
metadata:
  name: "imageupdater-sample"
  namespace: "argocd"
spec:
  namespace: "argocd"
  updateStrategy: "semver" # Default strategy if not specified per image
  forceUpdate: true
  allowTags: "regexp:^[0-9a-f]{7}$"
  ignoreTags: ["latest", "master"]
  pullSecret: "secret:<namespace>/<secret_name>#<field>" 
  writeBackMethod: "git:secret:argocd-image-updater/git-creds" # e.g., "git", "argocd" (to update Application CR). Default argocd - optional.
  gitConfig: # Group git-related settings. Should be ignored when writeBackMethod: "argocd"
    repository: "git@github.com:example/example.git"
    branch: "main"
    writeBackTarget: "helmvalues:/helm/config/test-values.yaml" # OR "kustomization:/config/overlays/bar" OR "helmvalues:/{{ .app.path.path }}/values.yaml" - for ApplicationSet.
  applicationRefs: # It's a list of applications or 1 app.
    - namePattern: "name1*"
      ...
    - namePattern: "name2*"
    ...
    - namePattern: "name3*"
      labelSelectors:
        matchLabels:
          app: "true"

      updateStrategy: "semver" # strategy for name3*
      forceUpdate: true
      allowTags: "regexp:^[0-9a-f]{7}$"
      ignoreTags: ["latest", "master"]
      pullSecret: "secret:<namespace>/<secret_name>#<field>" 
      writeBackMethod: "git:secret:argocd-image-updater/git-creds" # e.g., "git", "argocd" (to update Application CR). Default argocd - optional.
      gitConfig: # Group git-related settings. Should be ignored when writeBackMethod: "argocd"
        repository: "git@github.com:example/example.git"
        branch: "main"
[...]

PR #1190, opened 2025-07-16
Refactor argocd and registry-scanner modules to use CRD Spec instead of Annotations.

dkarpele:
1. Use pointers instead of values in `CommonUpdateSettings` type.

To support cascading configuration model (global -> application -> image-specific), fields within `CommonUpdateSettings` have been changed from value types (e.g., string, bool) to pointer types (*string, *bool). 
This  allows the system to distinguish between a "zero value" (like an empty string or false) and nil. This enables inheritance model where image-specific settings can override application-level settings, which in turn override global defaults. See `newContainerImageFromCommonSettings`. 
I remember about https://github.com/argoproj-labs/argocd-image-updater/pull/1134#discussion_r2096466253 but I think ArgoCD had a reason to use values. In our case pointers are much more convenient. I expect to change other fields to pointers later.

2. Adopt context.Context for Logging and Control Flow.

Function signatures throughout the argocd and registry-scanner modules have been updated to accept a `context.Context` as the first parameter. It provides a contextual Logging: as the context is passed down through the call stack, functions can add their own fields (e.g., application_name, image_name). 

3. Expand `image.ContainerImage` structure.

The `image.ContainerImage` struct has been expanded to hold not only the identity of an image but also its complete update configuration (strategy, allow/ignore lists, platform constraints, etc.). This turns it into a rich domain object, simplifying data flow by reducing the need to pass around multiple, separate configuration variables.

4. Refactor `GetImagesFromApplication`, `GetImagesAndAliasesFromApplication`, `ParseMatch`, `GetPlatformOptions` according to new CRD structure.

- `GetImagesFromApplication` and `GetImagesAndAliasesFromApplication` now source their information from the `ApplicationImages` struct.
- `ParseMatch` and `GetPlatformOptions` no longer parse annotations; instead, they receive strongly-typed data (e.g., `AllowTags` string, `Platforms` slice) directly from the `ContainerImage` configuration.

5. `newContainerImageFromCommonSettings` creates a new `image.ContainerImage` and populates it by layering the given settings on top of a parent configuration.
6. Refactor test.go.
7. Refactor tag.go

`ImageTagList` uses pointer receivers to ensure that every method call operates on the exact same instance of the struct and its lock.
`SortableImageTagList` uses value receivers for methods that don't need to modify the slice header itself (i.e., its length or capacity). The Len() and Swap() methods follow this convention, but Tags() did not.

8. Use a local version of the `registry-scanner` module in go.mod.
New changes in registry-scanner are incompatible with master branch and we can't commit these changes into master. The only way to continue working in crd and to have green tests is to use local version of registry-scanner.


go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/argoproj-labs/argocd-image-updater
 go 1.23.5
 
 require (
-	github.com/argoproj-labs/argocd-image-updater/registry-scanner v0.0.0-20250624020913-398db53f47e4
+	//github.com/argoproj-labs/argocd-image-updater/registry-scanner v0.0.0-20250624020913-398db53f47e4
```
chengfang: Shall we keep the original line, since it will be replaced in line 197 any way, and remove line 39? Will that achieve the same effect?

registry-scanner/pkg/image/image.go
```diff
@@ -152,6 +168,49 @@ func (img *ContainerImage) DiffersFrom(other *ContainerImage, checkVersion bool)
...
+		PullSecret:            img.PullSecret,
+		HelmParamImageName:    img.HelmParamImageName,
+		HelmParamImageVersion: img.HelmParamImageVersion,
+		original:              img.original,
+	}
+
+	if img.ImageTag != nil {
+		clone.ImageTag = &tag.ImageTag{
+			TagName:   img.ImageTag.TagName,
+			TagDigest: img.ImageTag.TagDigest,
```
chengfang, line 176: how about the field `TagDate   *time.Time` in `ImageTag`?

registry-scanner/pkg/image/image.go
```diff
@@ -5,20 +5,36 @@ import (
...
+	Platforms      []string
+
+	// Manifest-specific fields
 	HelmParamImageName    string
 	HelmParamImageVersion string
 	KustomizeImage        *ContainerImage
-	original              string
+
+	original string
 }
```
chengfang, line 25: ContainerImage is part of registry-scanner, which is supposed to be a neutral library. The issue with adding image updater knowledge to this type is that it becomes tied to image updater, and no longer easily reusable outside of image updater.

go.mod
dkarpele: Thanks, it works.

registry-scanner/pkg/image/image.go
dkarpele: Great! Thank you

registry-scanner/pkg/image/image.go
dkarpele: I have created type Image in types.go. 
I moved `GetParameterPullSecret` from registry-scanner to pkg because an original used annotations and this function is not related with registry-scanner.

chengfang:
Can you resolve the conflicts caused by recent rebasing onto master? After that it should be good to go.

chengfang (approved):

PR #1196, opened 2025-07-29
feat: Introduce Update methods (writeBackConfig) in CRD.

dkarpele:


api/v1alpha1/imageupdater_types.go
```diff
@@ -178,12 +178,14 @@ type CommonUpdateSettings struct {
 // specific configurations for that method, like Git settings.
 type WriteBackConfig struct {
 	// Method defines the method for writing back updated image versions.
-	// This acts as the default if not overridden.
-	// +optional
+	// This acts as the default if not overridden. If not specified, defaults to "argocd".
+	// +kubebuilder:validation:Required
 	// +kubebuilder:default:="argocd"
```
chengfang, line 183: Do you mean, if `WriteBackConfig` field is present, it must contain `Method` field with valid method value; and if `WriteBackConfig` field is not present, the method defaults to `argocd`?

chengfang:
>All types together with their receivers were moved to types.go from update.go. Update.go now starts with UpdateApplication

Did you mean all types in update.go and pkg/types/types.go are moved to argocd/types.go, and pkg/types/ are deleted?

api/v1alpha1/imageupdater_types.go
dkarpele: Yes, that's correct. 
User can still omit `WriteBackConfig` and it will be considered as `argocd`.
The default config is described here and is `argocd`
```
		wbc = &WriteBackConfig{
			Method:                 WriteBackApplication,
```

dkarpele:
> Did you mean all types in update.go and pkg/types/types.go are moved to argocd/types.go, and pkg/types/ are deleted?

Correct. It helps to avoid circular dependency.

chengfang (approved):

PR #1197, opened 2025-07-29
feat(webhook): add webhook handler for quay.io and make improvements to the community PR

cjcocokrisp:
This PR aims to add a Quay webhook processor to the webhook feature for Image Updater. This is accomplished by using the handler model that is used in the implementation. The handler is added along with the unit tests for it. 

On top of adding the webhook handler for quay, there is also some other enhancements to the feature in this PR. The list is below.

Enhancements:
- Fixed issues with the Git Write Back method by passing Git credentials to the webhook server which was not done in the initial implementation. 
- Made the webhook command more in line with the run command by adding similar options that were not present before. 
- Fixed some issues regarding secrets and added a secrets mechanism for Docker and Quay (More on this below) 

Secrets Changes:
The original implementation has this scheme to validate the Docker webhooks:
```go
// Validate validates the Docker Hub webhook payload
func (d *DockerHubWebhook) Validate(r *http.Request) error {
	if r.Method != http.MethodPost {
		return fmt.Errorf("invalid HTTP method: %s", r.Method)
	}

	// If secret is configured, validate the signature
	if d.secret != "" {
		signature := r.Header.Get("X-Hub-Signature-256")
		if signature == "" {
			return fmt.Errorf("missing webhook signature")
		}

		body, err := io.ReadAll(r.Body)
		if err != nil {
			return fmt.Errorf("failed to read request body: %w", err)
		}

		// Reset body for later reading
		r.Body = io.NopCloser(strings.NewReader(string(body)))

```
This scheme does not actually happen with Docker. Docker does not have that sort of scheme and the only registry that I found that has that is GitHub. Docker and Quay both in fact do not have any type of secrets and signatures built into their webhooks. [This issue](https://github.com/docker/roadmap/issues/51) that is open in the Docker roadmap mentions this and it is still open and people are still asking with the last reply being in June of this year. 

One thing I did to combat this issue of Docker and Quay not having secrets is add in a way that secrets could kind of be added. The way this is done by using the query parameter `secret` to hold what would be the secret. This is not the most secure thing and is more or less just there to be there. When I go and write the docs I plan to denote how this is not the most secure thing so don't use a secret that is shared with something else for this scheme. Let me know what you think about this.

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -32,6 +32,7 @@ import (
 // newRunCommand implements "run" command
 func newRunCommand() *cobra.Command {
 	var cfg *ImageUpdaterConfig = &ImageUpdaterConfig{}
+	var webhookCfg *WebhookConfig = &WebhookConfig{}
 	var once bool
 	var kubeConfig string
 	var disableKubernetes bool
@@ -183,7 +184,7 @@ func newRunCommand() *cobra.Command {
 
 			// Start the webhook server if enabled
 			var webhookServer *webhook.WebhookServer
-			if cfg.EnableWebhook && cfg.WebhookPort > 0 {
+			if cfg.EnableWebhook && webhookCfg.Port > 0 {
 				// Initialize the ArgoCD client for webhook server
 				var argoClient argocd.ArgoCD
 				switch cfg.ApplicationsAPIKind {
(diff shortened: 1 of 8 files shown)


cmd/webhook.go
```diff
@@ -2,37 +2,47 @@ package main
...
-	Plaintext           bool
-	GRPCWeb             bool
-	AuthToken           string
+type WebhookConfig struct {
+	Port          int
+	DockerSecret  string
+	GHCRSecret    string
+	QuaySecret    string
+	HarborSecret  string
+	UpdateOnEvent bool
```
chengfang: We should probably just remove this unused field UpdateOnEvent.

cmd/run.go
```diff
@@ -323,9 +330,15 @@ func newRunCommand() *cobra.Command {
...
 	runCmd.Flags().BoolVar(&cfg.EnableWebhook, "enable-webhook", env.GetBoolVal("ENABLE_WEBHOOK", false), "Enable webhook server for receiving registry events")
 
+	runCmd.Flags().IntVar(&webhookCfg.Port, "webhook-port", env.ParseNumFromEnv("WEBHOOK_PORT", 8082, 0, 65535), "Port to listen on for webhook events")
+	runCmd.Flags().StringVar(&webhookCfg.DockerSecret, "docker-secret", env.GetStringVal("DOCKER_WEBHOOK_SECRET", ""), "Secret for validating Docker Hub webhooks")
+	runCmd.Flags().StringVar(&webhookCfg.GHCRSecret, "ghcr-secret", env.GetStringVal("GHCR_WEBHOOK_SECRET", ""), "Secret for validating GitHub Container Registry webhooks")
+	runCmd.Flags().StringVar(&webhookCfg.QuaySecret, "quay-secret", env.GetStringVal("QUAY_WEBHOOK_SECRET", ""), "Secret for validating Quay webhooks")
+	runCmd.Flags().StringVar(&webhookCfg.HarborSecret, "harbor-secret", env.GetStringVal("HARBOR_WEBHOOK_SECRET", ""), "Secret for validating Harbor webhooks")
+	runCmd.Flags().BoolVar(&webhookCfg.UpdateOnEvent, "update-on-event", true, "Whether to trigger image update checks when webhook events are received")
+
 	return runCmd
```
chengfang, line 341: For these common flags shared between `webhook` and `run` commands, we may want to use the same set of names for consistency, and also make environtment variable names and flag names as close as possible. For ex,
* use `webhook-port` for both commands, to make it clear this port is for webhook use, not for heath or metrics, etc. In webhook command, it may be already implied, but it doesn't hurt to be extra clear.
* how about specifying that these secrets are for webhook, with `docker-webhook-secret` and the like, this is also follow the naming pattern of the corresponding env var.
*

cmd/webhook.go
```diff
@@ -44,68 +54,172 @@ update check for the affected images.
...
+	webhookCmd.Flags().BoolVar(&cfg.DisableKubeEvents, "disable-kube-events", env.GetBoolVal("IMAGE_UPDATER_KUBE_EVENTS", false), "Disable kubernetes events")
+
+	webhookCmd.Flags().IntVar(&webhookCfg.Port, "port", env.ParseNumFromEnv("WEBHOOK_PORT", 8082, 0, 65535), "Port to listen on for webhook events")
+	webhookCmd.Flags().StringVar(&webhookCfg.DockerSecret, "docker-secret", env.GetStringVal("DOCKER_WEBHOOK_SECRET", ""), "Secret for validating Docker Hub webhooks")
+	webhookCmd.Flags().StringVar(&webhookCfg.GHCRSecret, "ghcr-secret", env.GetStringVal("GHCR_WEBHOOK_SECRET", ""), "Secret for validating GitHub Container Registry webhooks")
+	webhookCmd.Flags().StringVar(&webhookCfg.QuaySecret, "quay-secret", env.GetStringVal("QUAY_WEBHOOK_SECRET", ""), "Secret for validating Quay webhooks")
+	webhookCmd.Flags().StringVar(&webhookCfg.HarborSecret, "harbor-secret", env.GetStringVal("HARBOR_WEBHOOK_SECRET", ""), "Secret for validating Harbor webhooks")
+	webhookCmd.Flags().BoolVar(&webhookCfg.UpdateOnEvent, "update-on-event", true, "Whether to trigger image update checks when webhook events are received")
```
chengfang, line 198: same comment as in `run` command

cmd/webhook.go
```diff
@@ -2,37 +2,47 @@ package main
...
+	"github.com/argoproj-labs/argocd-image-updater/registry-scanner/pkg/env"
 	"github.com/argoproj-labs/argocd-image-updater/registry-scanner/pkg/log"
+	"github.com/argoproj-labs/argocd-image-updater/registry-scanner/pkg/registry"
 
+	"github.com/argoproj/argo-cd/v2/util/askpass"
 	"github.com/spf13/cobra"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // WebhookOptions holds the options for the webhook server
```
chengfang: WebhookOptions => WebhookConfig in comment

cjcocokrisp:
Resolved feedback and added a comment above the quay payload with link to Quay docs on webhook events.

chengfang (approved):

PR #1198, opened 2025-08-01
manifests(webhook): expose webhook feature

cjcocokrisp:
Adds the Kubernetes manifests to expose the webhook feature. Created a service and ingress to expose the webhook server. Environment variables for CLI command flags related to the feature have been added to the container spec in the deployment as well. 

One other thing I noticed while testing the manifests was that when using the webhook command the health port was different so it would show up as failed within the cluster. So in the webhook command I realized I never actually set up the health and metrics server due to the webhook server having a health endpoint. So I removed the command flags for that in the webhook command.

### DIFF
--- a/manifests/base/kustomization.yaml
+++ b/manifests/base/kustomization.yaml
@@ -8,5 +8,4 @@ images:
 resources:
 - ./config
 - ./deployment
-- ./networking
 - ./rbac
--- a/manifests/install.yaml
+++ b/manifests/install.yaml
@@ -108,20 +108,6 @@ metadata:
     app.kubernetes.io/part-of: argocd-image-updater
   name: argocd-image-updater-secret
 ---
-apiVersion: v1
-kind: Service
-metadata:
-  name: argocd-image-updater-service
-spec:
-  ports:
-  - name: server-port
-    port: 8080
-    protocol: TCP
-    targetPort: 8080
-  selector:
-    app.kubernetes.io/name: argocd-image-updater
-  type: NodePort
----
 apiVersion: apps/v1
 kind: Deployment
 metadata:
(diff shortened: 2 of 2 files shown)


chengfang:
so with this PR, the new image-updater and service and ingress will always be installed after applying the install manifest file, either if the users do not enable it? 

How about not including them in the install, but providing the service and ingress manifests as a reference?

manifests/base/deployment/argocd-image-updater-deployment.yaml
```diff
@@ -113,6 +113,42 @@ spec:
...
+        - name: GHCR_WEBHOOK_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: argocd-image-updater-config
+              key: webhook.ghcr-secret
+              optional: true
+        - name: HARBOR_WEBHOOK_SECRET
+          valueFrom:
+              secretKeyRef:
+                name: argocd-image-updater-config
```
chengfang: Should these secrets come from the secret `argocd-image-updater-secret` in base/config/argocd-image-updater-secret.yaml?

cjcocokrisp: Yes that is correct for to edit it. Just fixed.

cjcocokrisp:
> so with this PR, the new image-updater and service and ingress will always be installed after applying the install manifest file, either if the users do not enable it?
> 
> How about not including them in the install, but providing the service and ingress manifests as a reference?

Yeah I think this would be a good idea. So should they just be removed from the base `kustomization.yaml` file? Then mention in the docs that you can find them in the location?

chengfang:
yes.

chengfang (approved):

PR #1205, opened 2025-08-02
feat(crd): introduce `ManifestTarget`

dkarpele:
1. Use pointers instead of values in `HelmTarget` and `KustomizeTarget` type.

To support cascading configuration model (global -> application -> image-specific), fields within  `HelmTarget` and `KustomizeTarget` have been changed from value types (e.g., string) to pointer types (*string). 
This  allows the system to distinguish between a "zero value" (like an empty string or false) and nil. 
I remember about https://github.com/argoproj-labs/argocd-image-updater/pull/1134#discussion_r2096466253 but I think ArgoCD had a reason to use values. In our case pointers are much more convenient. 

2. `ManifestTarget` settings were added to the `Image` struct.
3.  `GetHelmImage`, `SetHelmImage`, `GetKustomizeImage`, `SetKustomizeImage` now use CRD fields instead of annotations.
4. New function `newImageFromManifestTargetSettings`. We use `ManifestTarget`s per image so we call this function only from `parseImageListIuCR`. 
5. `newImageFromCommonUpdateSettings` was replaced by `mergeCommonUpdateSettings` and `newImageFromCommonUpdateSettings` to simplify the code. We don't need to `parentImage.Clone()` anymore and we don't pass parent as a parameter anymore. 
6. `newWBCFromCommonWBCSettings` was replaced by `mergeWBCSettings` and `newWBCFromSettings`. `newWBCFromSettings` imitate `getWriteBackConfig` function from master.
7. `Clone`s receivers were removed because code was simplified.
8. `context.Context` was passed to all necessary functions in the stack.
9. `getWriteBackBranch`: refactoring from annotations to CRD WriteBackConfig field.

### DIFF
--- a/pkg/argocd/types.go
+++ b/pkg/argocd/types.go
@@ -82,34 +82,6 @@ func (wbc *WriteBackConfig) RequiresLocking() bool {
 	}
 }
 
-// Clone creates a deep copy of the WriteBackConfig object.
-func (wbc *WriteBackConfig) Clone() *WriteBackConfig {
-	if wbc == nil {
-		return nil
-	}
-
-	clone := &WriteBackConfig{
-		Method:                 wbc.Method,
-		ArgoClient:             wbc.ArgoClient,
-		GitClient:              wbc.GitClient,
-		GetCreds:               wbc.GetCreds,
-		GitBranch:              wbc.GitBranch,
-		GitWriteBranch:         wbc.GitWriteBranch,
-		GitCommitUser:          wbc.GitCommitUser,
-		GitCommitEmail:         wbc.GitCommitEmail,
-		GitCommitMessage:       wbc.GitCommitMessage,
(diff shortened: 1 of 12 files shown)


pkg/argocd/types.go
```diff
@@ -182,6 +154,12 @@ type Image struct {
 	IgnoreTags     []string
 	PullSecret     string
 	Platforms      []string
+
+	// ManifestTarget settings
+	HelmName      string
+	HelmTag       string
+	HelmSpec      string
+	KustomizeName string
```
chengfang: Shall we name them as `HelmImageName, HelmImageTag, HelmImageSpec, KustomizeImageName` just to be a bit more descriptive?

dkarpele: fixed

PR #1207, opened 2025-08-05
docs(webhook): add documentation for webhook feature

cjcocokrisp:
Adds documentation for the webhook feature. 

Changes:
- Updates the `run` page with new webhook related flags
- Adds the documentation for `webhook` command
- Adds a configuration page with information about configuring the webhook 
- Removed some flags that I realized were not used from the `run` command

### DIFF
--- a/docs/configuration/webhook.md
+++ b/docs/configuration/webhook.md
@@ -1,6 +1,7 @@
 # Webhook Configuration
 
-Image Updater can be configured to respond to webhook notifications from various container registries. 
+Image Updater can be configured to respond to webhook notifications from 
+various container registries. 
 
 Currently Supported Registries:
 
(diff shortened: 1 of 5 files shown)


docs/configuration/webhook.md
```diff
@@ -0,0 +1,134 @@
...
+
+Because some container registries do not support secrets, there is a method included to have secrets for registries. This is through the query parameters in the URL of the webhook. This is not the most secure method and is there for a small extra layer.
+
+!!!warning
+    This is not the most secure method, it is just here for a small extra layer. 
+    **Do not use a secret that is shared with other critical services for this method!**
+
+It can be applied to the URL as below:
+```
+https://imageupdater.yourdomain.com/webhook?type<YOUR_REGISTRY_TYPE>&secret?=<YOUR_SECRET>
```
chengfang: looks like a missing = after type, and an extra ? after secret.

the host part of the url can also vary, doesn't have to be imageupdater.xxxx.com, is that right?

cjcocokrisp: Yeah it doesn't need to be imageupdater.xxxx.com it can be anything. Should I denote that in another way then or just put a disclaimer near it?

chengfang: maybe something like app1.example.com so users know they should use their own host name?

cjcocokrisp:
Altered the example url to be app1.example.com.

chengfang:
all the new command flags can also be specified via env var, and we may want to document this capability too.

docs/install/cmd/run.md
```diff
@@ -83,6 +83,18 @@ for images can only be specified from an environment variable.
...
+
+Secret for validating Docker Hub webhooks.
+
+**--enable-webhook *enabled***
+
+Enable webhook server for receiving registry events.
+
+**--ghcr-webhook-secret *secret***
+
+Secret for validating GitHub container registryw webhooks.
```
chengfang: registryw => registry

docs/configuration/webhook.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Secrets
+
+To help secure the webhook server you can apply a secret that is used to validate the incoming notification. The secrets can be set by editing the `argocd-image-updater-secret` secret.
+
+```
+stringData:
+  webhook.docker-secret: <YOUR_SECRET>
+  webhook.ghcr-secret: <YOUR_SECRET>
+  webhook.harbor-secret: <YOUR SECRET>
```
chengfang: use <YOUR_SECRET> to be consistent.

chengfang:
How about also include a sample webhook event payload?  

During your testing, have you seen issues with not receiving expected events due to various configuration such as unmatched secret, incorrect port, etc? If so, can you also include them in a troubleshooting section?

cjcocokrisp:
> How about also include a sample webhook event payload?
> 
> During your testing, have you seen issues with not receiving expected events due to various configuration such as unmatched secret, incorrect port, etc? If so, can you also include them in a troubleshooting section?

Yeah sure I can add a sample webhook event payload. The payloads vary from each of the container registries so do you think having an example specifically shown would be good or just links to the documentation for each of the registries would be suffice? 

For troubleshooting, I can definitely add a section that shows errors that you might receive as well.

chengfang:
> having an example specifically shown would be good or just links to the documentation for each of the registries would be suffice?

links should suffice; if links not available, then include a sample payload from one of the registries will be good too.

cjcocokrisp:
Added those two sections. Let me know what you think.

chengfang (approved):

PR #1208, opened 2025-08-05
feat: Allow yaml.SequenceNode type

### LINKED ISSUE #1032: helm values.yaml updates do not work with lists
**Describe the bug**
when using the image updater in git write back mode the image updater does not update values in the values.yaml file when the image.tag/name have lists.

an example 
```yaml
images:
  - name: blah
    image: quay.io/animage
    tag: v0.0.1
```

annotations 
```yaml

argocd-image-updater.argoproj.io/image-list: test=quay.io/animage
argocd-image-updater.argoproj.io/test.helm.image-name: images.0.image
argocd-image-updater.argoproj.io/test.helm.image-tag: images.0.tag

```

ultimately it picks up the image change but instead of updating it in place within the list of images it will add to the bottom of the values.yaml which ultimately does not trigger the image to be updated properly. 

```yaml
images[0]:
  - tag: v1.0.0 
```

The issue seems to be caused by the `setHelmValue` function which incorrectly assumes that the structure of the values is a map. https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/argocd/update.go#L575

**To Reproduce**
See above. 

**Expected behavior**
an update is made to the values.yaml as shown below  

```yaml
images:
  - name: blah
    image: quay.io/animage
    tag: v1.0.0

```

**Additional context**

**Version**
0.15.2

Note feel free to assign this issue to me. I'm happy to open up a pr to fix this

### LINKED ISSUE #1206: Allow yaml.SequenceNode as valid path for helm.image-spec, helm.image-name and helm.image-tag annotations
I am trying to configure Image Updater to automatically update the [Pipelines Init-container image (OpenWebUI Helm Chart Dependency)](https://github.com/open-webui/helm-charts/blob/main/charts/pipelines/values.yaml#L132-L138). In the Helm chart, the extraInitContainers field is a YAML array (yaml.SequenceNode), which is why I get the error "Could not update application spec: failed to set image parameter name value: unexpected type SequenceNode for key extraInitContainers"

I would like the path specified in helm.image-spec, helm.image-name and helm.image-tag to be able to contain yaml.SequenceNode. This can be an index indication using the listField[i] notation, for example like this:
```yaml
argocd-image-updater.argoproj.io/openwebuidev.helm.image-spec: pipelines.extraInitContainers[0].image
```

where pipelines.extraInitContainers is a yaml list

I had to change the original helm chart so that I could specify the image as a target.

**Additional context**
After analyzing the code, it seemed to me that changes were needed in the setHelmValue function.

AntonShadrinNN:
Closes #1206
Closes #1032

Fixes a situation where it was impossible to update images that were in the yaml array.
The code can probably be improved, I will be glad to follow the recommendations.

### DIFF
--- a/docs/configuration/images.md
+++ b/docs/configuration/images.md
@@ -455,20 +455,20 @@ use the new image.
 
 ```yaml
 foo:
-- name: bar-1
+- name: foo-1
   image: busybox:latest
   command: ['sh', '-c', 'echo "Custom container running"']
-- name: bar-2
+- name: foo-2
   image: nginx:1.19
 ```
 
 *Solution:* Use the index in square brackets of the item that needs to be updated, i.e.
 
 ```yaml
-argocd-image-updater.argoproj.io/fooalias.helm.image-tag: foo[1].image
+argocd-image-updater.argoproj.io/fooalias.helm.image-spec: foo[1].image
 ```
 
-This works also for both annotation `<image_alias>.helm.image-name` and annotation `<image_alias>.helm.image-tag`.
+This works for annotations `<image_alias>.helm.image-name`, `<image_alias>.helm.image-tag` and `<image_alias>.helm.image-spec`.
 
 
 ## Appendix
--- a/docs/configuration/images.md
+++ b/docs/configuration/images.md
@@ -460,8 +460,7 @@ foo:
   command: ['sh', '-c', 'echo "Custom container running"']
 - name: bar-2
   image: nginx:1.19
-
-```.
+```
 
 *Solution:* Use the index in square brackets of the item that needs to be updated, i.e.
 
--- a/docs/configuration/images.md
+++ b/docs/configuration/images.md
@@ -358,6 +358,10 @@ If the `<image_alias>.helm.image-spec` annotation is set, the two other
 annotations `<image_alias>.helm.image-name` and `<image_alias>.helm.image-tag`
 will be ignored.
 
+If the image is in the yaml list, then the index can be specified
+in the annotations `<image_alias>.helm.image-spec`, `<image_alias>.helm.image-name`
+or `<image_alias>.helm.image-tag` in square brackets.
+
 ## Examples
 
 ### Following an image's patch branch
@@ -444,6 +448,29 @@ When there's a new build for `yourorg/yourimage:latest` found in the registry,
 Argo CD Image Updater will update your configuration to use the SHA256 sum of
 the image, and Kubernetes will restart your pods automatically to have them
 use the new image.
+
+### Updating the image in the yaml list
+
+*Scenario:* You want to automatically update the image `nginx:1.19` that is inside the yaml list, e.g.
+
+```yaml
+foo:
+- name: bar-1
+  image: busybox:latest
+  command: ['sh', '-c', 'echo "Custom container running"']
+- name: bar-2
+  image: nginx:1.19
+
+```.
+
+*Solution:* Use the index in square brackets of the item that needs to be updated, i.e.
+
+```yaml
+argocd-image-updater.argoproj.io/fooalias.helm.image-tag: foo[1].image
+```
+
+This works also for both annotation `<image_alias>.helm.image-name` and annotation `<image_alias>.helm.image-tag`.
+
 
 ## Appendix
 
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -65,7 +65,8 @@ const (
 
 const defaultIndent = 2
 
-// listElementPattern is a regular expression for searching for an element in a yaml array
+// listElementPattern is a regular expression for searching for an element in a yaml array.
+// example: any-string[1]
 const listElementPattern = `^(.*)\[(.*)\]$`
 
 var re = regexp.MustCompile(listElementPattern)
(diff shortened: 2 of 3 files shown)


AntonShadrinNN:
I decided to use the abcd[0] syntax, as it is simpler in terms of parsing and more intuitive in terms of appearance.

docs/configuration/images.md
```diff
@@ -445,6 +449,28 @@ Argo CD Image Updater will update your configuration to use the SHA256 sum of
...
+  image: nginx:1.19
+```
+
+*Solution:* Use the index in square brackets of the item that needs to be updated, i.e.
+
+```yaml
+argocd-image-updater.argoproj.io/fooalias.helm.image-tag: foo[1].image
+```
+
+This works also for both annotation `<image_alias>.helm.image-name` and annotation `<image_alias>.helm.image-tag`.
```
chengfang: earlier in this doc, it says all 3 annotations will support this index notation: `image-name, image-tag, image-spec`. So the above line should be updated to include all 3.

docs/configuration/images.md
```diff
@@ -445,6 +449,28 @@ Argo CD Image Updater will update your configuration to use the SHA256 sum of
...
+  image: busybox:latest
+  command: ['sh', '-c', 'echo "Custom container running"']
+- name: bar-2
+  image: nginx:1.19
+```
+
+*Solution:* Use the index in square brackets of the item that needs to be updated, i.e.
+
+```yaml
+argocd-image-updater.argoproj.io/fooalias.helm.image-tag: foo[1].image
```
chengfang: to make it easier to understand which list element is targetted, you may want to change to use image foo-1 and foo-2 (not using another bar* image), and use foo-alias and foo[1] to refer to the 2nd element.

pkg/argocd/update.go
```diff
@@ -632,15 +634,45 @@ func setHelmValue(currentValues *yaml.Node, key string, value interface{}) error
...
+		var idPtr *int
+		// by default, the search is based on the key without changes, but
+		// if string matches pattern, we consider it is an id in YAML list.
+		key := k
+		matches := re.FindStringSubmatch(k)
+		if matches != nil {
+			idStr := matches[2]
+			id, err := strconv.Atoi(idStr)
+			if err != nil {
+				return fmt.Errorf("id \"%s\" in yaml array must match pattern ^(\\D+)\\[(\\d+)\\]$", idStr)
```
chengfang: the declared pattern should be used in the error msg, to be consistent.

pkg/argocd/update.go
```diff
@@ -632,15 +634,45 @@ func setHelmValue(currentValues *yaml.Node, key string, value interface{}) error
 
 	var err error
 	keys := strings.Split(key, ".")
-
+	// any-string[1]
+	pattern := `^(.*)\[(.*)\]$`
```
chengfang: we may want to use a more specific pattern to avoid invalid values like `foo[x]`:
```golang
pattern := `^([^[]+)\[(\d+)\]$`
```

pkg/argocd/update.go
AntonShadrinNN: My bad. Forgot to change it.

pkg/argocd/update.go
AntonShadrinNN: The original version looked like this, but it makes it harder to return a clear error [like this](https://github.com/argoproj-labs/argocd-image-updater/pull/1208/files#diff-2ed79833b3f255d5071af0d2bee1f5a158de25b5d59ebd5c3a42496910970866R2620-R2639) . The problem is that the search is based on the key. If it doesn't fit the pattern, then the search will follow it entirely, which means it will fall into the [else block](https://github.com/argoproj-labs/argocd-image-updater/pull/1208/files#diff-8e693a5e571533d36392d0511babeaca835de796757d447dbe7fc59af570ac18R688-L683) and simply write this value.  Should I take a different approach, such as checking the validity of the strings and parsing them separately?

docs/configuration/images.md
AntonShadrinNN: Done

docs/configuration/images.md
AntonShadrinNN: Done, if I got the gist right.

pkg/argocd/update.go
AntonShadrinNN: Changed, including tests

pkg/argocd/update.go
chengfang: It's a small difference, in that case, let's just keep what you have now.

pkg/argocd/update.go
```diff
@@ -632,15 +634,45 @@ func setHelmValue(currentValues *yaml.Node, key string, value interface{}) error
...
+	re := regexp.MustCompile(pattern)
```
chengfang: How about moving the regexp compilation to the pkg level var? This func is called multiple times per image update and we should avoid recompiling the regexp repeatedly.

AntonShadrinNN: That's a good point. Fixed it.

AntonShadrinNN:
@chengfang First of all, thank you for your help.

Can I ask you to tag these changes? I would like to use it to solve my original problem.

PR #1210, opened 2025-08-07
feat(webhook): add rate limiting to webhook endpoint

cjcocokrisp:
This PR aims to add rate limiting to the webhook endpoint so it can not be spammed with requests. This is accomplished with a sliding window approach where timestamps of requests are stored and then when you make a request it is checked how many requests were made in the specified window from the time of your request. If the window was fixed you could burst the requests at the end of the window and then at the start of the window which could cause overloading problems. The sliding window approach makes it so this is not a problem. The downside to this approach however, is that it requires more memory to store each individual timestamp. Because container registries could have rotating IP addresses this can be a problem due to information of an IP that won't make any more requests will still be stored. So to fix this I added a clean up interval that can be set that will clean any clients that have not been seen in a while. 

I looked into some packages that were available and found the [rate](https://pkg.go.dev/golang.org/x/time/rate) and [uber/ratelimit](https://github.com/uber-go/ratelimit) package but after doing research into what strategies are used for rate limiting it seemed like it wasn't the most difficult to implement.

I will add the documentation for this in a separate PR.


manifests/base/deployment/argocd-image-updater-deployment.yaml
```diff
@@ -149,6 +149,27 @@ spec:
...
+        - name: WEBHOOK_RATELIMIT_WINDOW
+          valueFrom:
+            configMapKeyRef:
+                name: argocd-image-updater-config
+                key: webhook.ratelimit-window
+        - name: WEBHOOK_RATELIMIT_CLEANUP_INTERVAL
+          valueFrom:
+            configMapKeyRef:
+                name: argocd-image-updater-config
+                key: webhook.ratelimit-cleanup-interval
```
chengfang: should the above env vars all be optional: true? Do they all have good defaults?

cjcocokrisp: They are meant to be optional forgot to write it, fixed.

pkg/webhook/ratelimit.go
```diff
@@ -0,0 +1,91 @@
...
+type RateLimiter struct {
+	mu       sync.Mutex
+	clients  map[string][]int64
+	lastSeen map[string]time.Time
+	window   time.Duration
+	allowed  int
+	done     chan bool
+}
+
+func NewRateLimiter(numRequets int, window time.Duration, cleanUpInterval time.Duration) *RateLimiter {
```
chengfang: numRequets => numRequests

pkg/webhook/ratelimit.go
```diff
@@ -0,0 +1,91 @@
...
+	if _, ok := rl.clients[clientIP]; !ok {
+		rl.clients[clientIP] = []int64{}
+	}
+
+	allow := false
+	windowStart := now.Unix() - int64(rl.window.Seconds())
+	filtered := []int64{}
+	for _, ts := range rl.clients[clientIP] {
+		if ts > windowStart {
+			filtered = append(filtered, ts)
```
chengfang: how about a sub-slice operation, starting from the first found good timestamp, to get the new `filtered` timestamps? Since we're re-populating the clients map for every request, and if the map is large, it may take up spaces to re-create this slice. Just thinking out aloud, with no benchmark evidence.

pkg/webhook/ratelimit.go
```diff
@@ -0,0 +1,91 @@
...
+
+// RateLimiter implements a sliding window rate limiting algorithm
+type RateLimiter struct {
+	mu       sync.Mutex
+	clients  map[string][]int64
+	lastSeen map[string]time.Time
+	window   time.Duration
+	allowed  int
+	done     chan bool
+}
```
chengfang: Can you add docs to this new type, especially 2 maps fields, preferrably with some sample data to show their usage?

pkg/webhook/ratelimit_test.go
```diff
@@ -0,0 +1,295 @@
...
+		name        string
+		clientIP    string
+		numRequests int
+		window      time.Duration
+		sendAmount  int
+		allowed     bool
+	}{
+		{
+			name:        "One request",
+			clientIP:    "127.0.0.1",
```
chengfang: does same IP but different port still count as different clients?  Can you add a test for that too?

cjcocokrisp: Yes it will count as different clients. I can add a test for that.

pkg/webhook/ratelimit.go
cjcocokrisp: Yeah that sounds like a better approach. Thanks for the suggestion.

cmd/run.go
```diff
@@ -337,6 +346,10 @@ func newRunCommand() *cobra.Command {
 	runCmd.Flags().StringVar(&webhookCfg.GHCRSecret, "ghcr-webhook-secret", env.GetStringVal("GHCR_WEBHOOK_SECRET", ""), "Secret for validating GitHub Container Registry webhooks")
 	runCmd.Flags().StringVar(&webhookCfg.QuaySecret, "quay-webhook-secret", env.GetStringVal("QUAY_WEBHOOK_SECRET", ""), "Secret for validating Quay webhooks")
 	runCmd.Flags().StringVar(&webhookCfg.HarborSecret, "harbor-webhook-secret", env.GetStringVal("HARBOR_WEBHOOK_SECRET", ""), "Secret for validating Harbor webhooks")
+	runCmd.Flags().BoolVar(&webhookCfg.RateLimitEnabled, "enable-webhook-ratelimit", env.GetBoolVal("ENABLE_WEBHOOK_RATELIMIT", false), "Enable rate limiting for the webhook endpoint")
+	runCmd.Flags().IntVar(&webhookCfg.RateLimitNumAllowedRequests, "webhook-ratelimit-num-allowed", env.ParseNumFromEnv("WEBHOOK_RATELIMIT_NUM_ALLOWED_REQUESTS", 100, 0, math.MaxInt), "The number of allowed requests in a window for webhook rate limiting")
```
chengfang: `WEBHOOK_RATELIMIT_NUM_ALLOWED_REQUESTS` seems a bit inconsistent with the flag name. How about shortening it to be in line with the flag name?

cjcocokrisp: How does `WEBHOOK_RATELIMIT_ALLOWED` sound?

cjcocokrisp:
Resolved the things mentioned in the review.

docs/configuration/webhook.md
```diff
@@ -168,6 +191,10 @@ environment variables. Below is the list of which variables correspond to which
 |`GHCR_WEBHOOK_SECRET` |`--gchr-webhook-secret`|
 |`HARBOR_WEBHOOK_SECRET` |`--harbor-webhook-secret`|
 |`QUAY_WEBHOOK_SECRET` |`--quay-webhook-secret`|
+|`ENABLE_WEBHOOK_RATELIMIT`|`--enable-webhook-ratelimit`|
+|`WEBHOOK_RATELIMIT_ALLOWED_`|`--webhook-ratelimit-allowed`|
```
chengfang: extra _ at the end of the env variable name.

docs/configuration/webhook.md
```diff
@@ -155,6 +155,29 @@ to apply them yourself.
...
+The window and amount of requests are configurable and can be set through editing
+the `argocd-image-updater-config` ConfigMap. There is also an interval that can be
+set to clean up the storage to remove clients that might not have sent notifications
+in awhile. 
+```yaml
+data:
+  # Enable rate limiting for the webhook endpoint
+  webhook.enable-rate-limit: true
+  # Set the amount of requests that can be made in a window before getting limited
+  webhook.ratelimit-num-allowed-requests: <SOME_NUMBER>
```
chengfang: how about also rename the config map key to follow the corresponding env var and flag name?

docs/install/cmd/run.md
```diff
@@ -225,4 +241,24 @@ whether to perform a cache warm-up on startup (default true)
...
 
+Can also be set with the `WEBHOOK_PORT` environment variable.
+
+**--webhook-ratelimit-cleanup-interval *duration***
+
+How often the rate limiter cleans up stale clients (default 1h0m0s)
+
+Can also be set with the `WEBHOOK_RATELIMIT_CLEANUP_INTERVAL` environment variable.
+
+**--webhook-ratelimit-num-allowed *numRequests***
```
chengfang: need to update the above flag name.

manifests/base/deployment/argocd-image-updater-deployment.yaml
```diff
@@ -149,6 +149,30 @@ spec:
...
+            configMapKeyRef:
+                name: argocd-image-updater-config
+                key: webhook.enable-rate-limit
+                optional: true
+        - name: WEBHOOK_RATELIMIT_NUM_ALLOWED_REQUESTS
+          valueFrom:
+            configMapKeyRef:
+                name: argocd-image-updater-config
+                key: webhook.ratelimit-num-allowed-requests
+                optional: true
```
chengfang: the above env var name should be consistent with the one used in run.go and webook.go.

the config map key should also be consistent with the rest as much as possible.

manifests/install.yaml
```diff
@@ -257,6 +257,30 @@ spec:
...
+            configMapKeyRef:
+              key: webhook.enable-rate-limit
+              name: argocd-image-updater-config
+              optional: true
+        - name: WEBHOOK_RATELIMIT_NUM_ALLOWED_REQUESTS
+          valueFrom:
+            configMapKeyRef:
+              key: webhook.ratelimit-num-allowed-requests
+              name: argocd-image-updater-config
+              optional: true
```
chengfang: the above env var name should be consistent with the one used in run.go and webook.go.

the config map key should also be consistent with the rest as much as possible.

cjcocokrisp:
Sorry about that, tried to get that done quick before leaving the office fixed again.

jannfis:
I was wondering why you would roll up your own rate limiter, especially given the fact that we already use the rate limiter from uber [within image updater](https://github.com/argoproj-labs/argocd-image-updater/blob/18767eda47352aadbf049edb066db615890c2b11/registry-scanner/pkg/registry/client.go#L29)

cjcocokrisp:
> I was wondering why you would roll up your own rate limiter, especially given the fact that we already use the rate limiter from uber [within image updater](https://github.com/argoproj-labs/argocd-image-updater/blob/18767eda47352aadbf049edb066db615890c2b11/registry-scanner/pkg/registry/client.go#L29)

I did not realize this and haven't really touched the registry scanner. I'll refactor this PR tomorrow thanks for the advice I should definitely look more at the packages used next time 🤦‍♂️

chengfang:
Good catch, Jan. I should've noticed that earlier. Yes, let's stick to uber ratelimiter.

cjcocokrisp:
I updated the rate limiting functionality to use the uber package now. This one is slightly different and it allows for a certain amount of requests per second instead of a certain amount in a window. It also blocks until it is able to make the request so I set it to always be enabled as well because I can see this bringing benefits in not overloading resources if a CI pipeline pushed a bunch of images that had notifications set up for Image Updater.

cmd/webhook.go
```diff
@@ -190,6 +193,7 @@ Supported registries:
 	webhookCmd.Flags().StringVar(&webhookCfg.GHCRSecret, "ghcr-webhook-secret", env.GetStringVal("GHCR_WEBHOOK_SECRET", ""), "Secret for validating GitHub Container Registry webhooks")
 	webhookCmd.Flags().StringVar(&webhookCfg.QuaySecret, "quay-webhook-secret", env.GetStringVal("QUAY_WEBHOOK_SECRET", ""), "Secret for validating Quay webhooks")
 	webhookCmd.Flags().StringVar(&webhookCfg.HarborSecret, "harbor-webhook-secret", env.GetStringVal("HARBOR_WEBHOOK_SECRET", ""), "Secret for validating Harbor webhooks")
+	webhookCmd.Flags().IntVar(&webhookCfg.RateLimitNumAllowedRequests, "webhook-ratelimit-allowed", env.ParseNumFromEnv("WEBHOOK_RATELIMIT_ALLOWED", 20, 0, math.MaxInt), "The number of allowed requests in a window for webhook rate limiting")
```
chengfang: the description needs update once we decide on the time unit.

cmd/run.go
```diff
@@ -337,6 +342,7 @@ func newRunCommand() *cobra.Command {
 	runCmd.Flags().StringVar(&webhookCfg.GHCRSecret, "ghcr-webhook-secret", env.GetStringVal("GHCR_WEBHOOK_SECRET", ""), "Secret for validating GitHub Container Registry webhooks")
 	runCmd.Flags().StringVar(&webhookCfg.QuaySecret, "quay-webhook-secret", env.GetStringVal("QUAY_WEBHOOK_SECRET", ""), "Secret for validating Quay webhooks")
 	runCmd.Flags().StringVar(&webhookCfg.HarborSecret, "harbor-webhook-secret", env.GetStringVal("HARBOR_WEBHOOK_SECRET", ""), "Secret for validating Harbor webhooks")
+	runCmd.Flags().IntVar(&webhookCfg.RateLimitNumAllowedRequests, "webhook-ratelimit-allowed", env.ParseNumFromEnv("WEBHOOK_RATELIMIT_ALLOWED", 20, 0, math.MaxInt), "The number of allowed requests in a window for webhook rate limiting")
```
chengfang, line 348: the description needs update once we decide on the time unit.

cmd/run.go
```diff
@@ -216,6 +219,8 @@ func newRunCommand() *cobra.Command {
 				log.Infof("Starting webhook server on port %d", webhookCfg.Port)
 				webhookServer = webhook.NewWebhookServer(webhookCfg.Port, handler, cfg.KubeClient, argoClient)
 
+				webhookServer.RateLimiter = ratelimit.New(webhookCfg.RateLimitNumAllowedRequests)
```
chengfang: with the default time unit being second, the lowest one can get is 1. For many users, I guess 1 request per second is still very frequent for image updates. I'm wondering if we could use a larger time unit, like hour or day? What do you think?

How about make 0 mean disabled, and make it the default, so users will need to explicitly opt in to ratelimiter?

cjcocokrisp: After some quick searches we can use a longer unit of time. One thing that I'm thinking is that the request will be blocked until the rate limiter says it can proceed. So I think having a really long interval might make it so the request will be blocked until the interval of the rate limit passes even if nothing is being updated which IMO will make it a bit too close to being like polling and at that point just use polling.

To combat this maybe we can have it add one request back in when an update is finished so the rate limiting kind of controls how many requests can be done at one time which I would assume the limit to that would be how much resources a user wants to dedicate to image updater. Not sure if I'm thinking about this right but let me know what you think about this?

 I do like the idea of it being opt in which could combat the issue I mentioned above. If a user opts in to the rate limiter then they would accept the fact that things might have to wait until the interval is up if that makes sense.

chengfang: Right, really longer interval can take long to reset. So day is probably too long, and hour looks like a good middle ground.

I'm hestitant to customize the limiting logic in the ratelimiter lib. The built-in slack option in uber ratelimiter should address above concerns to some degree. And this slack option is on by default and defaults to 10 allowances.

cjcocokrisp: I agree with the hour interval I think that would work great. 

I'll add in the enable flags and slack option by the end of the week. Thanks for the feedback!

chengfang: Sound great! I think we don't need to expose the slack option as a flag. Just using the default slack mechanism in the lib should suffice.

cjcocokrisp:
I made the adjustments to the PR.

One thing I also made an adjustment to is where the rate limit happens. I moved it to right before the updates happen because if the request gets held up on the rate limit then there may be a chance the container registry times it out and then disables the webhook. 

I know that Quay has it so if the webhook notification fails to send 3 time it will disable the webhook and that could cause some annoyance on the user. LMK what you think about this.

cmd/run.go
```diff
@@ -337,6 +344,8 @@ func newRunCommand() *cobra.Command {
 	runCmd.Flags().StringVar(&webhookCfg.GHCRSecret, "ghcr-webhook-secret", env.GetStringVal("GHCR_WEBHOOK_SECRET", ""), "Secret for validating GitHub Container Registry webhooks")
 	runCmd.Flags().StringVar(&webhookCfg.QuaySecret, "quay-webhook-secret", env.GetStringVal("QUAY_WEBHOOK_SECRET", ""), "Secret for validating Quay webhooks")
 	runCmd.Flags().StringVar(&webhookCfg.HarborSecret, "harbor-webhook-secret", env.GetStringVal("HARBOR_WEBHOOK_SECRET", ""), "Secret for validating Harbor webhooks")
+	runCmd.Flags().BoolVar(&webhookCfg.RateLimitEnabled, "enable-webhook-ratelimit", env.GetBoolVal("ENABLE_WEBHOOK_RATELIMIT", false), "Enable ratelimit for the webhook endpoint")
+	runCmd.Flags().IntVar(&webhookCfg.RateLimitNumAllowedRequests, "webhook-ratelimit-allowed", env.ParseNumFromEnv("WEBHOOK_RATELIMIT_ALLOWED", 1, 1, math.MaxInt), "The number of allowed requests in an hour for webhook rate limiting")
```
chengfang: I thought we decided to use this flag `webhook-ratelimit-allowed` for dual purpose: for enable/disable ratelimit (0 value is to disable it, also as the default value), and for setting the number of allowed requests (positive int). So we don't need to add another flag just for enabling/disabling ratelimit. WDYT?

cjcocokrisp: I completely misread that sorry, I will fix it right now. I think thats an easier way to do it.

cmd/webhook.go
```diff
@@ -21,16 +23,19 @@ import (
...
-	DockerSecret string
-	GHCRSecret   string
-	QuaySecret   string
-	HarborSecret string
+	Port                        int
+	DockerSecret                string
+	GHCRSecret                  string
+	QuaySecret                  string
+	HarborSecret                string
+	RateLimitEnabled            bool
```
chengfang: is `RateLimitEnabled` not used, and should be removed?

cjcocokrisp: Yes, fixed.

PR #1211, opened 2025-08-09
Feature/allow empty helm.image-name annotation

will4j:
In case when only image tag is required to update, like:
```yaml
kind: Application
metadata:
...
  annotations:
    argocd-image-updater.argoproj.io/image-list: app=harbor.example.com/nginx
    argocd-image-updater.argoproj.io/app.helm.image-tag: image.tag
    argocd-image-updater.argoproj.io/app.update-strategy: newest-build
    argocd-image-updater.argoproj.io/app.allow-tags: regexp:^sha-[0-9a-f]{7}$
    argocd-image-updater.argoproj.io/write-back-target: helmvalues
```
It make no sense to let `argocd-image-updater.argoproj.io/app.helm.image-name` annotation mandatorily, either I have to fill a fake image name value, still it produces meaningless configs in target `.argocd-source` or helm value files.

By this pr, I Supposed to get `values.yaml` file like
```yaml
image:
  registry: harbor.example.com
  repository: nginx
  tag: "sha-xxxx"
```
instead of
```yaml
image:
  registry: harbor.example.com
  name: harbor.example.com/nginx # which actually not used
  repository: nginx
  tag: "sha-xxxx"
```

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -516,7 +516,18 @@ func marshalParamsOverride(app *v1alpha1.Application, originalData []byte) ([]by
 				helmAnnotationParamName, helmAnnotationParamVersion := getHelmParamNamesFromAnnotation(app.Annotations, c)
 
 				if helmAnnotationParamName == "" {
-					return nil, fmt.Errorf("could not find an image-name annotation for image %s", c.ImageName)
+					// allow empty image-name
+					fmt.Sprintf("no image-name annotation found for image %s", c.ImageName)
+				} else {
+					helmParamName := getHelmParam(appSource.Helm.Parameters, helmAnnotationParamName)
+					if helmParamName == nil {
+						return nil, fmt.Errorf("%s parameter not found", helmAnnotationParamName)
+					}
+
+					err = setHelmValue(&helmNewValues, helmAnnotationParamName, helmParamName.Value)
+					if err != nil {
+						return nil, fmt.Errorf("failed to set image parameter name value: %v", err)
+					}
 				}
 				// for image-spec annotation, helmAnnotationParamName holds image-spec annotation value,
 				// and helmAnnotationParamVersion is empty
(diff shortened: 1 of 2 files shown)


chengfang:
when `image-name` annotation is absent, it defaults to `image.name`. That's why there is an extra `image.name` fields generated in your values file. But if you specify it with the following, the expected fields will be updated.
```yaml
argocd-image-updater.argoproj.io/app.helm.image-name: image.repository
```
The above tells image updater to use `image.repository` field for image name.

See https://argocd-image-updater.readthedocs.io/en/stable/configuration/images/#specifying-helm-parameter-names

chengfang:
see similar discussion: https://github.com/argoproj-labs/argocd-image-updater/discussions/1189

will4j:
> see similar discussion: #1189

The point is `app.helm.image-name` annotation is not required in my situation, even if I make it with `argocd-image-updater.argoproj.io/app.helm.image-name: image.repository`, it will not work as expected, it will generate values file like:

```yaml
image:
  registry: harbor.example.com
  repository: harbor.example.com/nginx
  tag: "sha-xxxx"
```

exactly the same problem as mentioned in discussion: #1189

chengfang:
Can you update the relevant sections of docs to reflect the new behavior?

Can you also add tests to verify the content of helm values when missing `image-name` annotation?

chengfang:
Can you also update the docs file: docs/configuration/images.md, e.g., around line 359, to describe the new behavior, that if the `<image_alias>.helm.image-tag` annotation is set but `<image_alias>.helm.image-name` annotation is not set, image updater will update the image tag value in the target helm values file, without updating the image name value.

will4j:
@chengfang I'm reconsidering the situation that `image.name` is the default value of anntotation `helm.image-name`， which used in both UpdateApplication flow and marshalParamsOverride flow. currently, this pr only handle marshalParamsOverride flow with helmvalues but marshalParamsOverride with default target (which is `.argocd-source-*.yaml` file) keeps unchanged. I'm not sure if this inconsistent will cause new problem, in addition, I have to confirm that this pr will not affect the UpdateApplication flow. After solving these uncertainty, I will update the doc files.

Issue #1108, opened 2025-04-18
Image updater unable to update image tags through ApplicationSet


eric09240:
**Describe the bug**
Using [write-back-target](https://argocd-image-updater.readthedocs.io/en/latest/basics/update-methods/#specifying-a-repository-when-using-a-helm-repository-in-repourl:~:text=helmvalues%20in%20config/base-,...or%20absolute%20with%20respect%20to%20the%20repository%3A,-%23%20absolute%20paths%20start%20with) to deploy an application through an ApplicationSet fails to update the image parameters. This works when deploying as an Application. 

Also referenced this [example](https://github.com/chengfang/image-updater-examples/tree/main/write-helmvalues) from a contributor. 
```
 level=info msg="git checkout --force main" dir=/tmp/git-fake-webserver execID=6f9b7
 level=info msg=Trace args="[git checkout --force main]" dir=/tmp/git-fake-webserver operation_name="exec git" time_ms=412.68863500000003
level=info msg="git clean -ffdx" dir=/tmp/git-fake-webserver execID=2ba29
level=info msg=Trace args="[git clean -ffdx]" dir=/tmp/git-fake-webserver operation_name="exec git" time_ms=1.492445
level=error msg="Could not update application spec: failed to set image parameter version value: unexpected type  for root" application=fake-webserver
```

**To Reproduce**
```
---
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: fake-webserver
  namespace: argo-cd
spec:
  goTemplate: true
  goTemplateOptions: ["missingkey=error"]
  generators:
    - git:
        repoURL: "https://foo.com/foo/argo-tools"
        revision: main
        directories:
        - path: "apps/fake-webserver/*/*"
        pathParamPrefix: app
        values:
          region: "{{index .path.segments 2}}"
          environment: "{{index .path.segments 3}}" 
  template:
    metadata:
      name: "fake-webserver-{{.values.region}}-{{.values.environment}}"
      namespace: argo-cd   
      labels:
        environment: "{{.values.environment}}"
        region: "{{.values.region}}"
      annotations:
        argocd-image-updater.argoproj.io/image-list: fake-webserver=xxxxxxxx.dkr.ecr.us-east-1.amazonaws.com/argo-tools/fake-webserver
        argocd-image-updater.argoproj.io/fake-webserver.update-strategy: newest-build
[...]

PR #1212, opened 2025-08-09
Feature/allow no exists helmvalues file

will4j:
I have a multi sources Application like below

```yaml
kind: Application
metadata:
  annotations:
    argocd-image-updater.argoproj.io/write-back-method: git
    argocd-image-updater.argoproj.io/write-back-target: helmvalues:/image-updater/.argocd-source-xyz-app-test.yaml
    argocd-image-updater.argoproj.io/git-repository: https://github.com/abc/xyz
    argocd-image-updater.argoproj.io/git-branch: release/prod
    argocd-image-updater.argoproj.io/image-list: client=harbor.abc.com/xyz-client,job=harbor.abc.com/xyz-job
    argocd-image-updater.argoproj.io/job.allow-tags: regexp:^test-sha-[0-9a-f]{7}$
    argocd-image-updater.argoproj.io/job.helm.image-tag: xyzJob.image.tag
    argocd-image-updater.argoproj.io/job.update-strategy: newest-build
    argocd-image-updater.argoproj.io/client.allow-tags: regexp:^test-sha-[0-9a-f]{7}$
    argocd-image-updater.argoproj.io/client.helm.image-tag: xyzClient.image.tag
    argocd-image-updater.argoproj.io/client.update-strategy: newest-build
  name: xyz-app-test
  namespace: argocd
spec:
  sources:
  - chart: xyz-app
    helm:
      ignoreMissingValueFiles: true
      releaseName: xyz-app
      valueFiles:
      - $values/apps-base/xyz-app/values-test.yaml
      - $values/image-updater/.argocd-source-xyz-app-test.yaml
    repoURL: harbor.abc.com/charts
    targetRevision: 1.0.0
  - ref: values
    repoURL: https://github.com/abc/abc-deploy
    targetRevision: release/prod
```

after deploy this app, argocd image updater will show failed logs
```logs
level=info msg="git checkout --force release/prod" dir=/tmp/git-xyz-app-test execID=49aae
level=info msg=Trace args="[git checkout --force release/prod]" dir=/tmp/git-xyz-app-test operation_name="exec git" time_ms=38.132636
level=info msg="git clean -ffdx" dir=/tmp/git-xyz-app-test execID=9f728
level=info msg=Trace args="[git clean -ffdx]" dir=/tmp/git-xyz-app-test operation_name="exec git" time_ms=2.081065
level=error msg="Could not update application spec: failed to set image parameter version value: unexpected type  for root" application=git-xyz-app-test
```

I found this is because when target file not exists, it will use `nil` originalData to marshal params override, which will failed when try to set helm value to wrong kind of yaml Node, like it explained [here](https://github.com/argoproj-labs/argocd-image-updater/issues/1108#issuecomment-2819538241).

and the call entry is like below:

https://github.com/argoproj-labs/argocd-image-updater/blob/c6167ecf2ca4e11759a49bcb9ae5ed06561018e8/pkg/argocd/git.go#L319-L337

This pr is intend to solve the problem by create a empy yaml.DocumentNode when originalData is nil, Therefore commit a new file to git when target file is not exists.

a new files likes:
```yaml
# auto generated by argocd image updater

xyzClient:
  image:
    tag: test-sha-c103907
xyzJob:
  image:
    tag: test-sha-f19def5

```

This may solve #1108 .


chengfang:
I noticed your target helm values file is named ".argocd-source-xyz-app-test.yaml". Do you still see the problem when the name does not start with `.argocd-source-*`? This prefix is used as the default target for git write-back.

will4j:
> I noticed your target helm values file is named ".argocd-source-xyz-app-test.yaml". Do you still see the problem when the name does not start with `.argocd-source-*`? This prefix is used as the default target for git write-back.

actually the pattern of filename is not limited here, I just named with prefix `.argocd-source` to identity these files auto generated by argo, which means argocd image updater here.

---

Further more, as I understand, `.argocd-source-*.yaml` file is the extra parameter file auto recognized by argocd, and it should placed at the source root dir, which is the dir where the chart is pulled and extracted to in my multiple sources application. so none of `.argocd-source-*.yaml` files in the ref repo(which is invisible when generate manifests of the helm chart) will take effect, it takes me lots of time to find out. 

So in another case, if remove `argocd-image-updater.argoproj.io/write-back-target` annotation and argocd image updater do successfully generate a `.argocd-source-*.yaml` file to the ref repo, which is the default behavior, it will not work as the file not used when rendering helm chart, currently i don't see a way to let `.argocd-source-*.yaml` file to work with remote helm repo source without extra user config, that's why I add `.argocd-source-xyz-app-test.yaml` to the valuesFiles of helm chart repo in my application.

pkg/argocd/update.go
```diff
@@ -485,10 +485,27 @@ func marshalParamsOverride(app *v1alpha1.Application, originalData []byte) ([]by
 		if strings.HasPrefix(app.Annotations[common.WriteBackTargetAnnotation], common.HelmPrefix) {
 			images := GetImagesAndAliasesFromApplication(app)
 
-			helmNewValues := yaml.Node{}
-			err = yaml.Unmarshal(originalData, &helmNewValues)
-			if err != nil {
-				return nil, err
+			var helmNewValues yaml.Node
+			if originalData == nil {
```
chengfang: how about changing the above condition to: `if len(originalData) == 0` to handle both nil and empty content?

PR #1216, opened 2025-08-13
chore(crd): remove all functions that have no remaining usages

dkarpele:


cmd/run.go
```diff
@@ -340,19 +343,21 @@ This enables a CRD-driven approach to automated image updates with Argo CD.
 	controllerCmd.Flags().StringVar(&cfg.RegistriesConf, "registries-conf-path", common.DefaultRegistriesConfPath, "path to registries configuration file")
 	controllerCmd.Flags().BoolVar(&disableKubernetes, "disable-kubernetes", false, "do not create and use a Kubernetes client")
 	controllerCmd.Flags().IntVar(&cfg.MaxConcurrency, "max-concurrency", 10, "maximum number of update threads to run concurrently")
+	controllerCmd.Flags().IntVar(&MaxConcurrentReconciles, "max-concurrent-reconciles", 1, "maximum number of concurrent Reconciles which can be run")
```
chengfang: How is `max-concurrent-reconciles` different from the existing `max-concurrency`? Is it possbile to merge them into one single flag?

chengfang: Shall we also support reading this flag from env var?

dkarpele: Great catch :)
I have renamed `MaxConcurrency` to `MaxConcurrentApps` to be more specific. 
we use  `MaxConcurrentApps` and `MaxConcurrentReconciles` in different parts of the code.

```
**--max-concurrent-apps *number***

Process a maximum of *number* applications concurrently. To disable concurrent
application processing, specify a number of `1`.

**--max-concurrent-reconciles *number***

Process a maximum of *number* ImageUpdater custom resources concurrently. 
This controls how many ImageUpdater CRs can be reconciled simultaneously by the controller. 
To disable concurrent reconciliation processing, specify a number of 1. 
Higher values may improve throughput but could increase resource usage and API load.
```

dkarpele: > Shall we also support reading this flag from env var?

We don't read `MaxConcurrentApps` from env vars. We can add both of them to env vars or none of them.
I don't have arguments what is better.

cmd/webhook.go
```diff
@@ -61,8 +61,8 @@ Supported registries:
 				return err
 			}
 
-			if cfg.MaxConcurrency < 1 {
-				return fmt.Errorf("--max-concurrency must be greater than 1")
+			if cfg.MaxConcurrentApps < 1 {
+				return fmt.Errorf("--max-concurrent-apps must be greater than 1")
```
chengfang: Can we enforce the value range in controllerCmd.Flags().IntVar(...)?

cmd/run.go
```diff
@@ -83,9 +84,9 @@ This enables a CRD-driven approach to automated image updates with Argo CD.
 				cfg.HealthPort = 0
 			}
 
-			// Enforce sane --max-concurrency values
-			if cfg.MaxConcurrency < 1 {
-				return fmt.Errorf("--max-concurrency must be greater than 1")
+			// Enforce sane --max-concurrent-apps values
+			if cfg.MaxConcurrentApps < 1 {
+				return fmt.Errorf("--max-concurrent-apps must be greater than 1")
```
chengfang: Can we enforce the value range in controllerCmd.Flags().IntVar(...)?

cmd/run.go
chengfang: ok, let's not add these env vars for now, since the defaults should work in most cases. If users do need to modify them, they can edit the deployment yaml adding these flags.

cmd/webhook.go
dkarpele: We can add it to `cobra.Command.PreRunE` function...

chengfang: I meant to say something like Flags().IntVar(..., min, max, ...), with min and max as parameters to IntVar func to use the built-in checking in IntVar func.

dkarpele: hm, but IntVar doesn't have these settings
`func (f *FlagSet) IntVar(p *int, name string, value int, usage string)`
https://pkg.go.dev/github.com/spf13/pflag#FlagSet.IntVar

dkarpele: Ah, I think I got it. I can replace this with
```
controllerCmd.Flags().IntVar(&cfg.MaxConcurrentApps, "max-concurrency", env.ParseNumFromEnv("MAX_CONCURRENT_APPS", 10, 1, 100), "maximum number of update threads to run concurrently")
```
so using `env.ParseNumFromEnv` but in this case we have to use env vars for MaxConcurrentApps and MaxConcurrentReconciles

dkarpele: I made patch, should be better now

PR #1223, opened 2025-08-25
chore(crd): add deprecation warnings for CRD branch removed functionality

dkarpele:
Add deprecation comments to functions, structs, constants, and flags that were removed in the CRD branch to notify users of upcoming breaking changes.

- Add deprecation notices to annotation-based functions in options.go
- Add deprecation notices to annotation-related constants in constants.go  and helper.go
- Add deprecation notices to native ArgoCD client struct and methods in argocd.go
- Add deprecation notices to removed CLI flags in run.go

These deprecation warnings inform users that the CRD branch introduces a
new architecture that eliminates the need for annotation-based approaches
and native ArgoCD client configurations.


cmd/run.go
```diff
@@ -309,12 +309,16 @@ func newRunCommand() *cobra.Command {
 		},
 	}
 
+	// DEPRECATED: These flags have been removed in the CRD branch and will be deprecated in a future release.
```
chengfang: shall we say '.. will be deprecated and removed in a future release' (to emphasize removal)?

chengfang: for affected cli flags, can we also add a short warning message to the description field, something like: https://github.com/argoproj/argo-cd/blob/master/server/cache/cache.go#L40 ?

chengfang: Do you also plan to add similar deprecation message to affected docs files?

dkarpele: Thanks,
all comments were addressed

chengfang (approved):

PR #1225, opened 2025-08-27
feat(crd): implement basic finalizer for graceful deletion

dkarpele:
See GITOPS-7467.
This is a basic finalizer structure currently without any logic. We can add logic later when we implement `status` for example or if we manipulate with other resources, or labels, etc.

### DIFF
--- a/internal/controller/imageupdater_controller.go
+++ b/internal/controller/imageupdater_controller.go
@@ -76,8 +76,8 @@ type ImageUpdaterReconciler struct {
 }
 
 const (
-	// imageUpdaterFinalizer is the name of the finalizer used by the ImageUpdater controller.
-	imageUpdaterFinalizer = "argocd-image-updater.argoproj.io/finalizer"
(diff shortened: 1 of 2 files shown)


internal/controller/imageupdater_controller.go
```diff
@@ -74,6 +75,11 @@ type ImageUpdaterReconciler struct {
 	Wg sync.WaitGroup
 }
 
+const (
+	// imageUpdaterFinalizer is the name of the finalizer used by the ImageUpdater controller.
+	imageUpdaterFinalizer = "argocd-image-updater.argoproj.io/finalizer"
```
chengfang: Shall we follow argo-cd finalizers' naming, like https://github.com/argoproj/argo-cd/blob/master/pkg/apis/application/v1alpha1/application_defaults.go#L10 ? It makes it easier to have more granular finalizers if needed.

dkarpele: sure, let's use the same naming convention

PR #1226, opened 2025-08-28
fix(crd): adjust deployment.yaml, fix build target in controller.Makefile, and bump go from 1.23.5 to 1.24 in controller.Dockerfile

chengfang:
* bump golang from 1.23.5 to 1.24 in controller.Dockerfile to be consistent with the version in go.mod
* fix build target in controller.Makefile to build the binary
* adjust container image in deployment.yaml. 
* add env vars and volumes to deployment.yaml
* copy configmap and secret files to config/default, adapt to the new naming and annotations, add them as kustomization resources

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -3,7 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
-    app.kubernetes.io/name: image-updater
+    app.kubernetes.io/name: image-updater-system
     app.kubernetes.io/managed-by: kustomize
   name: system
 ---
@@ -14,7 +14,7 @@ metadata:
   namespace: system
   labels:
     control-plane: controller-manager
-    app.kubernetes.io/name: image-updater
+    app.kubernetes.io/name: image-updater-controller-manager
     app.kubernetes.io/managed-by: kustomize
 spec:
   selector:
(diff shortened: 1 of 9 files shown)


chengfang:
@dkarpele can you please review?

controller.Makefile
```diff
@@ -131,7 +131,7 @@ lint-fix: golangci-lint ## Run golangci-lint linter and perform fixes
 
 .PHONY: build
 build: manifests generate fmt vet ## Build manager binary.
-	go build -o bin/manager cmd/main.go
```
dkarpele, line 134: In Makefile there is a target that builds binary:
```
.PHONY: controller
controller:
	CGO_ENABLED=0 GOOS=${OS} GOARCH=${ARCH} go build -ldflags '${LDFLAGS}' -o ${OUTDIR}/${BINNAME} cmd/*.go
```

Do you think we can add these env vars from `controller` target to `build` target?

config/manager/manager.yaml
```diff
@@ -59,17 +59,131 @@ spec:
...
+                key: git.commit-sign-off
+                name: image-updater-config
+                optional: true
+          - name: IMAGE_UPDATER_KUBE_EVENTS
+            valueFrom:
+              configMapKeyRef:
+                name: image-updater-config
+                key: kube.events
+                optional: true
+          - name: ARGOCD_LOGLEVEL
```
dkarpele: Not related to this PR but `ARGOCD_LOGLEVEL` seems legacy var. It's not used anywhere. I think it can be safely removed from all places.

dkarpele (commented):
There is a github action [Ensure kubernetes manifests conform to their schema](https://github.com/argoproj-labs/argocd-image-updater/actions/runs/17306700119/job/49131354771?pr=1226#logs)

that runs against `manifests` dir. Do you think we should run it against `config` dir also?

chengfang:
> There is a github action [Ensure kubernetes manifests conform to their schema](https://github.com/argoproj-labs/argocd-image-updater/actions/runs/17306700119/job/49131354771?pr=1226#logs)
> 
> that runs against `manifests` dir. Do you think we should run it against `config` dir also?

That will be good. But we'll need to provide the image updater CRD schema to teh validator in JSON openapi format, or configure it to skip all image updater CR.

config/manager/manager.yaml
chengfang: Good catch. I'll remove them.

controller.Makefile
chengfang: yes, will update it to be consistent.

dkarpele:
> That will be good. But we'll need to provide the image updater CRD schema to teh validator in JSON openapi format, or configure it to skip all image updater CR.

Yes, I see. Let's do it separately.

dkarpele (approved):

Issue #1227, opened 2025-08-30
Updating empty values file in git fails unexpectedly


MenD32:
**Describe the bug**
Trying to update an empty values-example.yaml file in a git repository using ArgoCD Image Updater will cause an error in the controller instead of updating the image successfully

**To Reproduce**
- Create a git repo with a helm chart
- Create an file named values-example.yaml with only whitespace characters (newline f.e.)
- Create an application with the following annotations:
```yaml
argocd-image-updater.argoproj.io/force-update: "true"
argocd-image-updater.argoproj.io/write-back-method: git
argocd-image-updater.argoproj.io/write-back-target: 'helmvalues:values-example.yaml'
argocd-image-updater.argoproj.io/example.helm.image-tag: example.image.tag
argocd-image-updater.argoproj.io/example.helm.image-name: example.image.repository
argocd-image-updater.argoproj.io/image-list: example=<YOUR_OCI_REPOSITORY>
argocd-image-updater.argoproj.io/update-strategy: newest-build
```
- Create a new tag in the image repository

ArgoCD Image Updater will fail to update the values-example.yaml 

**Expected behavior**
ArgoCD should add to the values-example.yaml the following text:
```yaml
example:
  image:
    repository: ***
    tag: ***
```

**Additional context**

**Version**
v16.0.0

**Logs**
time="2025-08-30T12:11:23Z" level=error msg="Could not update application spec: failed to set image parameter version value: unexpected type  for root" application=***

zentavr:
I have the same issue. My Application file is the next:
```yaml
---
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: "example-io"
  namespace: "sys-argocd"
  labels:
    'argocd.example.io/deployed-by': "argo-cd"
  annotations:
    notifications.argoproj.io/subscribe.slack: website-deployment-staging
    # ArgoCD Image Updater setup
    #argocd-image-updater.argoproj.io/write-back-method: "argocd"
    #argocd-image-updater.argoproj.io/write-back-method: "git"
    argocd-image-updater.argoproj.io/write-back-method: git:secret:sys-argocd/argocd-repo-argocd-git-shared
    argocd-image-updater.argoproj.io/git-repository: "git@github.com:example/argocd-staging-shared.git"
    argocd-image-updater.argoproj.io/git-branch: "main"
[...]

PR #1228, opened 2025-08-30
fix: whitespace only values.yaml files fail parsing

MenD32:
- whitespace values.yaml files fail parsing even though they are valid yaml files
- fixes https://github.com/argoproj-labs/argocd-image-updater/issues/1227


pkg/argocd/update.go
```diff
@@ -874,3 +876,17 @@ func commitChanges(app *v1alpha1.Application, wbc *WriteBackConfig, changeList [
...
+		return true
+	}
+	for i := 0; i < len(data); {
+		r, size := utf8.DecodeRune(data[i:])
+		if !unicode.IsSpace(r) {
+			return false
+		}
+		i += size
+	}
+	return len(data) > 0
```
chengfang: since we already know from the first line, that the data is not empty, should we just return true at the end?

pkg/argocd/update_test.go
```diff
@@ -2241,6 +2241,72 @@ replicas: 1
 		_, err := marshalParamsOverride(&app, nil)
 		assert.Error(t, err)
 	})
+
+	t.Run("test", func(t *testing.T) {
```
chengfang: can you change to a more descriptive name instead of "test"?

MenD32: oops! my bad, I'll fix it

pkg/argocd/update.go
MenD32: sounds good

MenD32:
Hi, I added a fix to the comments in the CR, ready for review again

PR #1231, opened 2025-09-05
Add support for JSON log formatting

mikepilat:
- Add log setter for text or JSON log formatting
- Add CLI option "--log-format", accepts "text" or "json"
- Defaults to current "text" logging for compatibility

Additionally:
- Configure `cmd` go.mod to reference registry-scanner locally to support concurrent changes in both modules


docs/install/cmd/run.md
```diff
@@ -195,6 +195,12 @@ Set the log level to *level*, where *level* can be one of `trace`, `debug`,
 
 Can also be set using the *IMAGE_UPDATER_LOGLEVEL* environment variable.
 
+**--log-format *format***
+
+Set the log format to *format*, where *level* can be one of `text` or `json`
```
jannfis:
```suggestion
Set the log format to *format*, where *format* can be one of `text` or `json`
```

Dockerfile
```diff
@@ -4,6 +4,7 @@ RUN mkdir -p /src/argocd-image-updater
 WORKDIR /src/argocd-image-updater
 # cache dependencies as a layer for faster rebuilds
 COPY go.mod go.sum ./
+COPY registry-scanner ./
```
jannfis: I think this belongs to a different PR?

go.mod
```diff
@@ -37,7 +37,7 @@ require (
 	dario.cat/mergo v1.0.1 // indirect
 	github.com/Azure/go-ansiterm v0.0.0-20230124172434-306776ec8161 // indirect
 	github.com/MakeNowJust/heredoc v1.0.0 // indirect
-	github.com/Masterminds/semver/v3 v3.3.1 // indirect
+	github.com/Masterminds/semver/v3 v3.4.0 // indirect
```
jannfis: I think this belongs to a different PR?

go.mod
```diff
@@ -211,3 +211,5 @@ replace (
 	k8s.io/sample-cli-plugin => k8s.io/sample-cli-plugin v0.32.2
 	k8s.io/sample-controller => k8s.io/sample-controller v0.32.2
 )
+
+replace github.com/argoproj-labs/argocd-image-updater/registry-scanner => ./registry-scanner
```
jannfis: I think this belongs to a different PR?

chengfang:
According to https://argo-cd.readthedocs.io/en/stable/user-guide/commands/argocd_admin/#options, argocd uses `logformat` instead of `log-format`. I think we should probaby follow the existing naming pattern of `loglevel` (used in both argo-cd and image-updater), and `logformat` from argocd.  The same for the corresponding environmental variable.

chengfang:
@mikepilat I moved the changes related to registry-scanner replacement into a seprate #1234 since that could also benefit other PRs in the same concurrent update situation. I also took the liberty to resolve the conflicts and pushed to your branch.

Can you please review the remaining comments in this PR so we can include this in the upcoming release?

docs/install/cmd/run.md
mikepilat: Fixed in 7f3a65b7132eb5857559eee04ef4fa13c5828f59

mikepilat:
@chengfang Thanks for bringing that upstream! Other PR feedback has been incorporated.

docs/install/cmd/run.md
mikepilat: Thanks, fixed in 7f3a65b

PR #1232, opened 2025-09-06
fix(crd): refactor webhook receiver to align with new CRD architecture

dkarpele:


cmd/common.go
```diff
@@ -0,0 +1,124 @@
...
+	// error when the file does not exist, but we emit a warning.
+	if cfg.RegistriesConf != "" {
+		st, err := os.Stat(cfg.RegistriesConf)
+		if err != nil || st.IsDir() {
+			setupLogger.Info("Registry configuration not found or is a directory, using default configuration", "path", cfg.RegistriesConf, "error", err)
+		} else {
+			err = registry.LoadRegistryConfiguration(ctx, cfg.RegistriesConf, false)
+			if err != nil {
+				setupLogger.Error(err, "could not load registry configuration", "path", cfg.RegistriesConf)
+				return nil
```
chengfang: with the above `return nil`, we will skip the rest of the setup steps in this function upon errors loading registry conf file. Is this intended?

dkarpele: Good catch! It's not intended, I just copy pasted an original implementation of this function. It makes sense to return err because other parts of this function have 
```
		setupLogger.Error(err, "could not create K8s client")
		return err
```
```
		setupLogger.Error(err, "could not parse commit message template")
		return err
```
```
				setupLogger.Error(err, "could not read commit message template", "path", commitMessagePath)
				return err
``` 
etc. Let me create patch.

PR #1236, opened 2025-09-12
docs(crd): update community documentation

dkarpele:
- Refactor legacy annotation based documentation
- Add migration guide
- Rename vars in webhook.go

### DIFF
--- a/docs/configuration/migration.md
+++ b/docs/configuration/migration.md
@@ -1,0 +1,296 @@
+# Migration Guide: From Annotations to CRD
+
+This guide helps you migrate from the legacy annotation-based configuration to the new CRD-based approach in Argo CD Image Updater.
+
+## Overview
+
+The new CRD-based approach provides several advantages over the annotation-based system:
+
+- **Better structure**: Configuration is now organized in a clear, hierarchical structure
+- **Type safety**: CRD validation ensures correct field types and required values
+- **Easier management**: No need to manage complex annotation strings
+- **Better tooling support**: IDEs and YAML editors provide better autocomplete and validation
+- **Scalability**: Easier to manage multiple applications and images
+
+## Migration Steps
+
+### Step 1: Create an ImageUpdater Resource
+
+Instead of annotating individual Argo CD Applications, you now create dedicated `ImageUpdater` resources.
+
+**Before (Annotations):**
+```yaml
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: my-app
+  annotations:
+    argocd-image-updater.argoproj.io/image-list: nginx=nginx:1.20
+    argocd-image-updater.argoproj.io/nginx.update-strategy: semver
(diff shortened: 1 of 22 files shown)


docs/configuration/migration.md
```diff
@@ -0,0 +1,296 @@
...
+    argocd-image-updater.argoproj.io/nginx.update-strategy: latest
+```
+
+### After (CRD-based)
+
+```yaml
+apiVersion: argocd-image-updater.argoproj.io/v1alpha1
+kind: ImageUpdater
+metadata:
+  name: production-image-updater
```
chengfang: this CR covers both production and staging apps so the name should reflect that.

docs/configuration/migration.md
```diff
@@ -0,0 +1,296 @@
...
+      images:
+        - alias: "nginx"
+          imageName: "nginx:1.20"
+          commonUpdateSettings:
+            allowTags: "regexp:^[0-9]+\\.[0-9]+$"
+            ignoreTags: ["latest", "dev"]
+        - alias: "redis"
+          imageName: "redis:6.2"
+          commonUpdateSettings:
+            updateStrategy: "latest"
```
chengfang: We should use `newest-build` to avoid the deprecated strategy name `latest`

With `newest-build` strategy, we should be able to omit the tag in `imageName` field.

docs/configuration/migration.md
```diff
@@ -0,0 +1,296 @@
...
+    - namePattern: "staging-app"
+      writeBackConfig:
+        method: "git"
+        gitConfig:
+          branch: "develop"
+      images:
+        - alias: "nginx"
+          imageName: "nginx:latest"
+          commonUpdateSettings:
+            updateStrategy: "latest"
```
chengfang: We should use `newest-build` to avoid the deprecated strategy name `latest`

With `newest-build` strategy, we should be able to omit the tag in `imageName` field.

docs/configuration/migration.md
```diff
@@ -0,0 +1,296 @@
...
+      repository: "git@github.com:myorg/myrepo.git"
+      branch: "main"
+      writeBackTarget: "helmvalues:./values.yaml"
+  applicationRefs:
+    - namePattern: "production-app"
+      images:
+        - alias: "nginx"
+          imageName: "nginx:1.20"
+          commonUpdateSettings:
+            allowTags: "regexp:^[0-9]+\\.[0-9]+$"
```
chengfang, line 266: why the double backslash in some regexp values?

dkarpele: YAML treats `\.` as an escape for `.`, so it remains a dot, but depending on the parser, sometimes this could cause an error (e.g., if YAML tries to interpret `\.` as a special character).
https://yaml.org/spec/1.2.2/#escaped-characters
In annotations we can use one backslash because annotations treat as a text

chengfang (approved):

PR #1243, opened 2025-09-19
feat: Refactor prometheus metrics 

dkarpele:
Prometheus metrics were successfully exported after refactoring
```
dkarpele@dkarpele-mac /tmp % curl --insecure --header "Authorization: Bearer ${TOKEN}" https://127.0.0.1:8443/metrics | grep argocd
argocd_image_updater_applications_watched_total The total number of applications watched by Argo CD Image Updater CR
# TYPE argocd_image_updater_applications_watched_total gauge
argocd_image_updater_applications_watched_total{image_updater_cr_name="image-updater-001",image_updater_cr_namespace="argocd-image-updater-system"} 1
# HELP argocd_image_updater_k8s_api_errors_total The total number of K8S API requests resulting in error
# TYPE argocd_image_updater_k8s_api_errors_total counter
argocd_image_updater_k8s_api_errors_total 0
# HELP argocd_image_updater_k8s_api_requests_total The total number of K8S API requests performed by the Argo CD Image Updater
# TYPE argocd_image_updater_k8s_api_requests_total counter
argocd_image_updater_k8s_api_requests_total 0
```

- Remove legacy flags `health-port` and `metrics-port` and replace them with `health-probe-bind-address` and `metrics-bind-address`. Update docs. 
- `health` module was removed in favor of controller-runtime.
- Add RBAC rules for metrics. 
- Images metrics will be implemented in GITOPS-8068


chengfang:
For the metric port and health port flags, shall we also support setting them via env var? If so, also update `manager.yaml`?

The new flag names, health-probe-bind-address, metrics-bind-address, look like support both hostname or ip plus port number. But I think in practice, we only use the port number part. What are the accepted format, are values like :9999 or 9999 both acceptable?

I'm wondering if we should rename them from the old health-port and metrics-port, or just stick to the old names. The old names are consistent with argo-cd (for example, https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-server/commands/argocd_server.go#L314), and a bit more readable. If we only care about the port part, I feel the old names are fine.

cmd/run.go
```diff
@@ -267,16 +257,11 @@ This enables a CRD-driven approach to automated image updates with Argo CD.
 		},
 	}
 
-	// TODO: flags below are not documented yet and don't have env vars yet. Metrics and health checks will be implemented in GITOPS-7113
 	controllerCmd.Flags().StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
-	controllerCmd.Flags().StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	controllerCmd.Flags().StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to. Leave as 0 to disable the probe service.")
```
chengfang: the default is :8081, so 'Leave as 0' can be confusing.

chengfang: How does the `health-probe-bind-address` work with the liveness setting in `manager.yaml`? Are there overlapping, or should they be kept in sync?
```yaml
        livenessProbe:
          httpGet:
            path: /healthz
            port: 8081
```

dkarpele:
> For the metric port and health port flags, shall we also support setting them via env var? If so, also update manager.yaml?

We didn't have env vars before for these flags, argocd doesn't have env vars for metrics port (see link you provided), metrics are optional and disabled by default (it's `0`). So I think we shouldn't add them as env vars at least before customers request.

dkarpele:
> The new flag names, health-probe-bind-address, metrics-bind-address, look like support both hostname or ip plus port number. But I think in practice, we only use the port number part. What are the accepted format, are values like :9999 or 9999 both acceptable?
I'm wondering if we should rename them from the old health-port and metrics-port, or just stick to the old names. The old names are consistent with argo-cd (for example, https://github.com/argoproj/argo-cd/blob/master/cmd/argocd-server/commands/argocd_server.go#L314), and a bit more readable. If we only care about the port part, I feel the old names are fine.

The accepted formats are:
`":8081"`
`"host.com:8081"`
`"10.20.30.40:8081"`
not recommended is `"8081"`

The difference between argocd flags and image updater flags is that argocd flag is `int` but we use `string`. Yes, in practice we use only port numbers but library supports hostnames and ip addresses, there are options `BindAddress` and `HealthProbeBindAddress`. If we stick to the old names it makes things even more confusing: the field name `metrics-port` accept hostname/ip + port.

 health-probe-bind-address, metrics-bind-address flag names were autogenerated by kubebuilder.

cmd/run.go
dkarpele: They should be kept in sync. If these settings are not synchronized, the health check will fail, leading to the pod being terminated and restarted. `health-probe-bind-address` tells the controller which port to listen on for incoming health check requests. `livenessProbe` is a k8s setting that tells the kubelet where to send requests to check if the container is healthy. If they don't match the health check fails, connection is refused and k8s kills the pod.

chengfang (approved):

PR #1276, opened 2025-10-06
chore(crd): prepare for release

dkarpele:


config/samples/argocd-image-updater_v1alpha1_imageupdater.yaml
```diff
@@ -2,9 +2,9 @@ apiVersion: argocd-image-updater.argoproj.io/v1alpha1
 kind: ImageUpdater
 metadata:
   labels:
-    app.kubernetes.io/name: image-updater
+    app.kubernetes.io/name: argocd-image-updater
     app.kubernetes.io/managed-by: kustomize
-  name: imageupdater-sample
+  name: argocd-image-updater-sample
```
chengfang, line 7: a few lines below, 2 `namePatterns` are both `image-updater-001`. Should they be different?

chengfang:
Previously some users experience slowness when searching apps across all namespaces. That was the reason we added a `--application-namespace` flag to limit the search and update to apps in specific namespace in legacy releases. With crd and controller, without this flag, is this still a concern with cluster wide operations?

config/samples/argocd-image-updater_v1alpha1_imageupdater.yaml
dkarpele: Of course, makes sense

dkarpele:
> Previously some users experience slowness when searching apps across all namespaces. That was the reason we added a `--application-namespace` flag to limit the search and update to apps in specific namespace in legacy releases. With crd and controller, without this flag, is this still a concern with cluster wide operations?

I want to be brave and say no it's not a concern :)
With new architecture we _"always apply `--application-namespace`"._ In CRD there is required field `spec.namespace` that is used to restict argocd apps to a single namespace. Reconcile function works with one CR at a time, so with one namespace.

In 
https://github.com/argoproj-labs/argocd-image-updater/blob/24f5e51fed964b1138960aa2b1b8526851062e61/pkg/argocd/argocd.go#L222-L249
we use  
```
    listOpts := []ctrlclient.ListOption{
    	ctrlclient.InNamespace(cr.Spec.Namespace),
    }
```
The ctrlclient.InNamespace() option tells the client library to make a List request to the Kubernetes API server specifically for the cr.Spec.Namespace. The API server does the filtering, and the controller only ever receives the list of applications from that one namespace. And it's efficient because we make 1 List API call, no multiple Get calls.

config/manager/manager.yaml
```diff
@@ -62,115 +63,115 @@ spec:
...
           - name: WEBHOOK_PORT
             valueFrom:
               configMapKeyRef:
-                name: image-updater-config
+                name: argocd-image-updater-config
                 key: webhook.port
                 optional: true
           - name: QUAY_WEBHOOK_SECRET
             valueFrom:
               configMapKeyRef:
```
chengfang: this should be a secretKeyRef, just like the next few other secret refs.

chengfang:
Now that the CR is cluster-scoped, and have pretty liberal rbac, but it only watches one namespace, which seems a bit wasteful. I'm wondering if we can replace spec.namespace with either spec.namespaces, or spec.namespaceSelector, so we  might as well support updating apps in any namespace with a single CR. WDYT? This is not a change request for this PR.

config/manager/manager.yaml
dkarpele: Thanks fixed. Btw the same is in master

chengfang: Yes, we'll need to fix it in master too. Is install.yaml re-generated?

dkarpele:
> Now that the CR is cluster-scoped, and have pretty liberal rbac, but it only watches one namespace, which seems a bit wasteful. I'm wondering if we can replace spec.namespace with either spec.namespaces, or spec.namespaceSelector, so we might as well support updating apps in any namespace with a single CR. WDYT? This is not a change request for this PR.

I tested controller again and it seems I was too fast applying this change. Thanks for the thorough review. 
We can release with Namespaced scope and wait for users feedback and make next decisions afterwards.
So, I'm reverting these changes.
I had error below that I can't reproduce now:
```
time="2025-10-02T14:44:54Z" level=error msg="Failed to watch" error="failed to list *v1alpha1.ImageUpdater: imageupdaters.argocd-image-updater.argoproj.io is forbidden: User \"system:serviceaccount:argocd-image-updater-system:argocd-image-updater-controller\" cannot list resource \"imageupdaters\" in API group \"argocd-image-updater.argoproj.io\" at the cluster scope" logger=controller-runtime.cache.UnhandledError reflector="pkg/mod/k8s.io/client-go@v0.33.1/tools/cache/reflector.go:285" type="*v1alpha1.ImageUpdater"
```

I tested these scenarios (ImageUpdater is IU): 
1. 
- ArgoCD server is in `argocd-image-updater-e2e` ns
- ArgoCD app is in `argocd-image-updater-e2e` ns
- ArgoCD apps deployment (pod) is in `image-updater-e2e-001` ns. 
```
spec:
  destination:
    server: https://kubernetes.default.svc
    namespace: image-updater-e2e-001
```
- IU CRD is **Cluster** scoped
- IU controller runs in `argocd-image-updater-system` ns
- IU CR is in `argocd-image-updater-system` ns

**Result**: Successful update, no errors

2. Same but IU CRD is in **`namespaced`** scope 
- ArgoCD server is in `argocd-image-updater-e2e` ns
- ArgoCD app is in `argocd-image-updater-e2e` ns
- ArgoCD apps deployment (pod) is in `image-updater-e2e-001` ns. 
```
spec:
  destination:
    server: https://kubernetes.default.svc
    namespace: image-updater-e2e-001
```
- IU CRD is **Namespaced** scoped
- IU controller runs in `argocd-image-updater-system` ns
- IU CR is in `argocd-image-updater-system` ns

Result: Successful update, no errors

config/manager/manager.yaml
dkarpele: Yes, I regenrated it with make bundle-install

chengfang:
@dkarpele See #1279 for fixing env var configMapKeyRef in master branch, and adding the same resource request and limit to master.

PR #1279, opened 2025-10-10
fix(manifests): use secretKeyRef instead of configMapKeyRef for QUAY_WEBHOOK_SECRET; add resource request and limit

chengfang:
I copied the resource quest and limit values from crd branch. They feel pretty small, especially the request. For comparison, this is the default resources request and limit for GitOps components: https://docs.redhat.com/en/documentation/red_hat_openshift_gitops/1.18/html-single/installing_gitops/index#sizing-requirements-for-gitops_preparing-gitops-install

If we follow dex, that would be:

```yaml
  resources:
    requests:
      cpu: 250m
      memory: 128Mi
    limits:
      cpu: 500m
      memory: 256Mi
```
@dkarpele WDYT?

### DIFF
--- a/manifests/base/deployment/argocd-image-updater-deployment.yaml
+++ b/manifests/base/deployment/argocd-image-updater-deployment.yaml
@@ -127,7 +127,7 @@ spec:
                 optional: true
         - name: QUAY_WEBHOOK_SECRET
           valueFrom:
-            configMapKeyRef:
+            secretKeyRef:
                 name: argocd-image-updater-secret
                 key: webhook.quay-secret
                 optional: true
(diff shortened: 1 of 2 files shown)


dkarpele:
CPU seems ok, but memory is too small. I think we can follow one of controllers: openshift-gitops-application-controller
 or applicationset-controller.

chengfang:
I updated to below, making limit double the amount of request:
```yaml
        resources:
          limits:
            cpu: 500m
            memory: 1024Mi
          requests:
            cpu: 250m
            memory: 512Mi
```

dkarpele:
LGTM

PR #1280, opened 2025-10-13
Fix GitHub Webhook response parsing

### LINKED ISSUE #1278: Github webhook fails to parse response, unsupported package type: CONTAINER
**Describe the bug**
We just did a fresh setup of Argocd Image Updater and noticed that the webhook payload of Github seems to have changed, the value of the `"package_type"` property is now uppercase  `"CONTAINER"` which breaks the parser in https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/webhook/ghcr.go#L111-L113 and results in an `unsupported package type: CONTAINER` error message.

If I change it manually to lowercase and execute the command via curl, all works as expected: `Webhook received and processing`

I don't know if Github changed this from lower to upper case, maybe we should make it an equals ignore case check.

**To Reproduce**
Setup webhook with Github

**Expected behavior**
Webhook requests are successfully evaluated

**Additional context**
```
{
  "action": "published",
  "package": {
    "id": xxx,
    "name": "xxx",
    "namespace": "xxx",
    "description": "",
    "ecosystem": "CONTAINER",
    "package_type": "CONTAINER",
    "html_url": "https://github.com/xxx/packages/yyy",
    "created_at": "2025-09-04T11:38:53Z",
    "updated_at": "2025-10-10T13:42:54Z",
 ...
 ```

**Version**
0.17.0

**Logs**
│ time="2025-10-10T14:36:38Z" level=error msg="Failed to process webhook: unsupported package type: CONTAINER" remote="xxx"

SailReal:
Fixes #1278

### DIFF
--- a/pkg/webhook/ghcr_test.go
+++ b/pkg/webhook/ghcr_test.go
@@ -148,12 +148,60 @@ func TestGHCRWebhook_Parse(t *testing.T) {
 		expectError  bool
 	}{
 		{
-			name: "valid registry package published event",
+			name: "valid registry package published event with lower case package_type",
 			payload: `{
 				"action": "published",
 				"package": {
 					"name": "myapp",
 					"package_type": "container",
+					"owner": {
+						"login": "myuser"
+					},
+					"package_version": {
+						"name": "v1.0.0",
+						"container_metadata": {
+							"tag": {
+								"name": "v1.0.0"
+							}
+						}
+					}
+				}
+			}`,
+			expectedRepo: "myuser/myapp",
+			expectedTag:  "v1.0.0",
+			expectError:  false,
+		},
+		{
+			name: "another valid registry package published event with upper case package_type",
+			payload: `{
+				"action": "published",
+				"package": {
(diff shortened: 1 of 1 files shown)


chengfang:
can you also update some tests in pkg/webhook/ghcr_test.go to verify package_type values of different cases?

SailReal:
Oh sure, I searched for tests but couldn't find this file :sweat_smile:

chengfang:
> Oh sure, I searched for tests but couldn't find this file 😅

https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/webhook/ghcr_test.go

SailReal:
> > Oh sure, I searched for tests but couldn't find this file 😅
> 
> https://github.com/argoproj-labs/argocd-image-updater/blob/master/pkg/webhook/ghcr_test.go

Haha yes, sure, the first hint was already enough, I meant before you pointed me into the direction, I did only looked in `/test` for tests but thanks and sorry for the time :sweat_smile:

PR #1287, opened 2025-10-21
feat: Preserve original tag when using digest strategy

zc149:
# Preserve original tag when using digest strategy

## Summary
When using the **digest** strategy, Argo CD Image Updater replaces the image tag with the digest value (e.g. `nginx:1.14.2` → `nginx@sha256:xxxx`).  
This behavior ensures immutability but makes it difficult to know which tag was originally deployed.

This PR adds support for preserving the **original tag** by recording it as an annotation on the Application.

---

## Motivation
- Digest strategy is useful for reproducibility, but it hides the original tag.
- Many users (myself included) found it inconvenient when debugging or reviewing Applications, since the human-friendly tag is lost.
- By storing the original tag as an annotation, users can:
  - Keep digest for immutability
  - See the original tag for readability and traceability

---

## Implementation Details
- Added `RecordOriginalTag` helper in `pkg/argocd/argocd.go`
- Updated `UpdateApplication` in `pkg/update/update.go` to call `RecordOriginalTag` when strategy = **digest**
- Annotation key format:

- Example:
```yaml
metadata:
  annotations:
    argocd-image-updater.argoproj.io/original-tag.nginx: "1.14.2"

### DIFF
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -785,6 +785,24 @@ func SetHelmImage(ctx context.Context, app *argocdapi.Application, newImage *ima
 	return nil
 }
 
+// recordOriginalTag saves the original tag as an annotation when using digest strategy.
+// This allows users to see which human-readable tag was originally deployed while
+// maintaining the immutability benefit of using digests.
+func recordOriginalTag(app *argocdapi.Application, img *image.ContainerImage, originalTag string) {
+	if app.Annotations == nil {
+		app.Annotations = make(map[string]string)
+	}
+
+	// Use ImageAlias if available, otherwise use ImageName
+	identifier := img.ImageAlias
+	if identifier == "" {
+		identifier = img.ImageName
+	}
+
+	key := fmt.Sprintf("%s/original-tag.%s", ImageUpdaterAnnotationPrefix, identifier)
+	app.Annotations[key] = originalTag
+}
+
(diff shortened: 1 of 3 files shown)


chengfang:
How does this approach work with `git` write-back-method?

zc149:
> How does this approach work with `git` write-back-method?

Thank you for raising this question — that’s a great point.

@chengfang 
To clarify: the main motivation for this change is specifically for the annotation write-back method (e.g. when using the ECR poll approach). In this mode, when the digest strategy is used, the original image tag gets replaced by the digest and the tag information is lost from the Application spec. Adding it back as an annotation helps preserve that information for better readability and troubleshooting.

For the git write-back method, I agree this isn’t really necessary. Since the tag remains in the Git repository manifests themselves, there is already a reliable source of truth and history for the tag. So in practice, this change is mostly relevant and useful only for the annotation-based flow.

jankosecki:
Do we think this could still be added but maybe with an additional condition that it should be done only with `method: "argocd"`

I've been looking for something similar but instead of original tag, I was considering annotating deployment with tag/tags of the pulled digest (so one doesn't need to check registry to cross-check digest and actual tag of image to ensure we are running what is expected to run)

zc149:
@jankosecki 
Thanks for the feedback! 
The previous PR got broken, so I’ve applied your suggestions and recreated it here.
PR #1476

PR #1304, opened 2025-10-27
fix: include reference string in invalid reference format error (#1303)

### LINKED ISSUE #1303: Improve error message for invalid reference format
**Describe the bug**
When argocd-image-updater encounters an invalid image reference format, it logs an error like:

```go
time="2025-10-27T19:37:36Z" level=error msg="Could not get tags from registry: invalid reference format" alias= application=workflows-prod image_name= image_tag= registry=
```

The error message doesn't include the actual image reference string that failed validation, making it difficult to debug the issue.
The image information fields (`image_name`, `image_tag`, `registry`) are also empty in these cases.

**To Reproduce**
1. Have an invalid reference format

**Expected behavior**

The error message should include the invalid reference string so users can identify what image reference caused the problem, for example:

```go
Failed to create repository for image 'invalid@reference': invalid reference format
```

**Additional context**

The error originates from `reference.WithName(nameInRegistry)` in the registry client when parsing an invalid image reference. The error bubbles up through `GetTags()` but loses the context of what reference string caused the failure.

I recommend a log message in `registry-scanner/pkg/registry/registry.go` when `NewRepository()` fails to include the invalid reference string in the logs, enabling users to identify and fix the problematic image reference.

**Version**

**Logs**

ellisvalentiner:
When NewRepository fails to parse an image reference, log the actual reference string that failed validation. This makes it possible to identify which image reference is causing the "invalid reference format" error.

Previously, users would see:
  Could not get tags from registry: invalid reference format

Now users will see:
  Failed to create repository for image 'invalid@reference': invalid reference format

This provides the necessary context to debug configuration issues with malformed image references.

Fixes #1303

### DIFF
--- a/registry-scanner/pkg/registry/registry.go
+++ b/registry-scanner/pkg/registry/registry.go
@@ -44,6 +44,7 @@ func (ep *RegistryEndpoint) GetTags(ctx context.Context, img *image.ContainerIma
 	}
 	err = regClient.NewRepository(nameInRegistry)
 	if err != nil {
+		logCtx.Errorf("Failed to create repository for image '%s': %v", nameInRegistry, err)
 		return nil, err
 	}
 	tTags, err := regClient.Tags(ctx)


chengfang:
Thanks for the fix. Can you also apply the same to master-annotation-based branch?

Issue #1335, opened 2025-11-12
helmvalues not working properly


diamanat:
**Describe the bug**
With v1.0.0, when using writeBackConfig.gitConfig.writeBackTarget set to helmvalues, Argo CD Image Updater commits a file containing Argo CD Helm parameters (helm.parameters with forcestring fields) instead of a plain Helm values YAML file. This is a regression: older versions produced a regular values.yaml consumable via Application.spec.sources[n].helm.valueFiles.
Concretely, the committed file looks like this (actual output):

`helm:
  parameters:
  - name: image-updater/api.image.name
    value: api
    forcestring: true
  - name: Module.api.image.tag
    value: main@sha256:fa066ecd309f64f6c821e0342bd1f05b92e23a4e5a640d5bfedc4e618ff9acde
    forcestring: true@argocd-image-updater`

This is not a Helm values file and cannot be referenced from Argo CD as a values file.

**To Reproduce**
Deploy Argo CD Image Updater v1.0.0.
Create an ImageUpdater CR with:
spec.writeBackConfig.method: git
spec.writeBackConfig.gitConfig.writeBackTarget: helmvalues:/abc/versions1.yaml
A Helm application with ManifestTargets (helm name/tag) such as:
name: image-updater/api.image.name
tag: Module.api.image.tag
Ensure a new image is available to trigger an update (e.g., update a digest).
Observe the commit in the target repo: the file contains helm.parameters instead of a plain values.yaml.

**Expected behavior**
When writeBackTarget starts with helmvalues:, the updater should commit a plain Helm values YAML file setting the configured keys, e.g.:
```yaml
image-updater:
  api:
    image:
      name: api
Module:
  api:
    image:
      tag: main@sha256:fa066ecd309f64f6c821e0342bd1f05b92e23a4e5a640d5bfedc4e618ff9acde
```

**Version**
1.0.0

chengfang:
[...]

PR #1339, opened 2025-11-13
fix(helmvalues): git write-back to helmvalues file produces wrong structure

chengfang:
Fixes https://github.com/argoproj-labs/argocd-image-updater/issues/1335

## Summary by CodeRabbit

* **Refactor**
  * Improved detection of application source types to more reliably identify kustomize and helm sources.
  * Adjusted parameter-handling conditions to better handle overridden chart and values paths, reducing misclassification and improving update accuracy.
* **Tests**
  * Updated test inputs and setups to reflect the new configuration field usage; no behavioral changes.

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -379,7 +379,7 @@ func marshalParamsOverride(ctx context.Context, applicationImages *ApplicationIm
 			return []byte{}, nil
 		}
 
-		if wbc != nil && strings.HasPrefix(wbc.Target, common.HelmPrefix) {
+		if wbc != nil && !strings.HasPrefix(filepath.Base(wbc.Target), common.DefaultTargetFilePrefix) {
 			images := GetImagesAndAliasesFromApplication(applicationImages)
 
 			var helmNewValues yaml.Node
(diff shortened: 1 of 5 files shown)


pkg/argocd/update.go
```diff
@@ -379,7 +379,7 @@ func marshalParamsOverride(ctx context.Context, applicationImages *ApplicationIm
 			return []byte{}, nil
 		}
 
-		if wbc != nil && strings.HasPrefix(wbc.Target, common.HelmPrefix) {
+		if wbc != nil && !strings.HasPrefix(filepath.Base(wbc.Target), ".argocd-source-") {
```
dkarpele: I think it makes sense to add new var to constants.go. For example 
`DefaultWriteBackTargetPrefix               = ".argocd-source-"` and use it here.

dkarpele (changes requested):

dkarpele:
I tested with an empty `writeBackTarget`, like this:
```
  writeBackConfig:
    method: "git:secret:argocd-image-updater-e2e/e2e-git-repo"
    gitConfig:
      branch: master
      repository: "https://127.0.0.1:30003/testdata.git"
```
and `.argocd-source-write-helmvalues.yaml` was written to the repo but the content is
```
helm:
  parameters:
  - name: image.repository
    value: docker.io/bitnamilegacy/nginx
    forcestring: true
  - name: image.tag
    value: 1.27.5
    forcestring: true
```
Is it expected? Just want to check

chengfang:
yes, the above sample content from `argocd-source-*.yaml` looks correct.

pkg/argocd/update.go
chengfang: fixed.

PR #1379, opened 2025-12-02
fix: preserve existing Helm tag parameter when image has no tag (#1351)

### LINKED ISSUE #1351: forceUpdate: true clears Helm parameters when there is no tag in registry, causing invalid image reference
**Describe the bug**

When `forceUpdate: true` is set in the ImageUpdater CRD, argocd-image-updater sets Helm parameter (`global.image.tag`) to empty when there is no image tag specified in the Artifact Registry's `imageName` field. This causes a syntax error because the resulting image reference becomes invalid (e.g., `registry/image:` with an empty tag). Additionally, argocd-image-updater removes the `global.image.tag` Helm parameter that was explicitly set in the ArgoCD Application manifest.

Previously, the image updater would preserve the existing `global.image.tag` value even when there was no specified image tag in Artifact Registry.

**To Reproduce**

1. Create an ArgoCD Application manifest with `global.image.tag` set as a Helm parameter:
   ```yaml
   kind: Application
   metadata:
     name: example-app
   spec:
     project: preview
     source:
       repoURL: https://github.com/org/repo.git
       path: infrastructure/helm/frontend
       targetRevision: test
       helm:
         valueFiles:
           - ../envs/frontend-partners/values-preview.yaml
         parameters:
           - name: global.image.tag
             value: "mq@sha256:123456"
             forceString: true
   ```

2. Create an ImageUpdater CRD with `forceUpdate: true` and an `imageName` that doesn't include a tag:
   ```yaml
   kind: ImageUpdater
   metadata:
     name: example-image-updater
   spec:
     namespace: argocd
     commonUpdateSettings:
       forceUpdate: true
       updateStrategy: "digest"
     writeBackConfig:
       method: "argocd"
     applicationRefs:
       - namePattern: "example-app"
         images:
           - alias: frontend-partners
             imageName: "example-registry/docker/frontend-partners"
             manifestTargets:
               helm:
                 name: "global.image.name"
                 tag: "global.image.tag"
   ```

3. Apply both resources to the cluster
4. Observe that argocd-image-updater removes the `global.image.tag` parameter from the Application manifest
5. Observe the error when ArgoCD tries to deploy the application

**Expected behavior**

The image updater should preserve the existing `global.image.tag` value when:
- `forceUpdate: true` is set
- The `global.image.tag` is explicitly set in the Application manifest as a Helm parameter

The image updater should not remove Helm parameters from the Application manifest, and should not set `global.image.tag` to empty, which results in invalid image references.

**Additional context**

- This issue only occurs when `forceUpdate: true` is set
- When `forceUpdate: false` (or omitted), the image updater works correctly and preserves the tag value
- This appears to be a regression from previous behavior where the tag value was preserved
- Related to the fix for issue #1344
- The error occurs because Kubernetes/Docker cannot parse image references with empty tags (e.g., `registry/image:`)

**Version**

1.0.1

**Logs**

```
Failed to apply default image tag "example-registry/docker/frontend-partners:": couldn't parse image name "example-registry/docker/frontend-partners:": invalid reference format
```

chengfang:
When forceUpdate is enabled with an imageName that has no tag specified, SetHelmImage was setting the Helm tag parameter to an empty string. This caused "invalid reference format" errors when the image reference was later parsed (e.g., "example-registry/docker/frontend-partners:").

This fix modifies SetHelmImage to skip setting the tag parameter when the new image has no tag, preserving the existing tag value instead.

Fixes #1351

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where Helm image tag parameters could be overwritten with empty values during updates.

* **Tests**
  * Added test cases validating that existing image tag parameters are preserved when updating Helm images without a specified tag.

### DIFF
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -706,8 +706,15 @@ func SetHelmImage(ctx context.Context, app *argocdapi.Application, newImage *ima
 	} else {
 		mergeParams = append(mergeParams,
 			argocdapi.HelmParameter{Name: hpImageName, Value: newImage.GetFullNameWithoutTag(), ForceString: true},
-			argocdapi.HelmParameter{Name: hpImageTag, Value: newImage.GetTagWithDigest(), ForceString: true},
 		)
+		// Only set the tag parameter if we have a non-empty tag value.
+		// When forceUpdate is enabled and no tag is specified, the tag can be empty.
+		// Setting an empty tag would overwrite existing tag values and cause invalid image references.
+		if tagValue := newImage.GetTagWithDigest(); tagValue != "" {
+			mergeParams = append(mergeParams,
+				argocdapi.HelmParameter{Name: hpImageTag, Value: tagValue, ForceString: true},
+			)
+		}
 	}
 
 	appSource := getApplicationSource(ctx, app)
--- a/pkg/argocd/argocd_test.go
+++ b/pkg/argocd/argocd_test.go
@@ -1019,6 +1019,81 @@ func Test_SetHelmImage(t *testing.T) {
 		require.Error(t, err)
 	})
 
+	t.Run("Test set Helm image with empty tag preserves existing tag (issue #1351)", func(t *testing.T) {
+		// This test verifies the fix for issue #1351:
+		// https://github.com/argoproj-labs/argocd-image-updater/issues/1351
+		//
+		// Scenario from the issue:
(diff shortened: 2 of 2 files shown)


dkarpele:
Thanks @chengfang 
Will you create the same fix for annotation based version?

PR #1398, opened 2025-12-15
chore: update github.com/argoproj-labs/argocd-image-updater/registry-scanner from v1.0.0 to v1.0.2

dkarpele:
For master-annotation-based branch


dkarpele:
We want to keep legacy version of registry-scanner in master-annotation-based

PR #1417, opened 2025-12-27
force app type to helm when write back target is set to helmvalues

Jyrno42:
Previously when in a multi source application that included both helm and kustomize the app type was detected as kustomize even when i explicitly set the writeback target to helmvalues.

This PR automatically detects helmvalues: prefix from writeback target and sets WriteBackConfig.HelmChart to true if its found.

Similarly to wbc.KustomizeBase the app type is returned as "helm" when wbc.HelmChart is true

Previous implementation of multiple sources is naive and returns the first app type argocd finds. If it would be somehow possible to tell image updater what source index to prefer then this change would not be needed for me atleast. But for now this method seems to be okay given such logic laready exists for kustomize.

## Summary by CodeRabbit

* **New Features**
  * Write-back configuration now recognizes Helm chart write-back targets and treats them as Helm application sources when detected, improving multi-source resolution and ensuring Helm sources take precedence where applicable.
  * Helm-target detection supports conventional Helm target prefixes.

* **Tests**
  * Added unit and integration tests covering Helm vs. Kustomize write-back parsing and source-type resolution to prevent regressions.

### DIFF
--- a/pkg/argocd/argocd_test.go
+++ b/pkg/argocd/argocd_test.go
@@ -387,6 +387,30 @@ func Test_GetApplicationSourceType(t *testing.T) {
 		appType := GetApplicationSourceType(application, wbc)
 		assert.Equal(t, v1alpha1.ApplicationSourceTypeKustomize, appType)
 	})
+
+	t.Run("Get application Source type with helmvalues target", func(t *testing.T) {
+		application := &v1alpha1.Application{
+			ObjectMeta: v1.ObjectMeta{
+				Name:      "test-app",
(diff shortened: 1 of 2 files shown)


chengfang:
@Jyrno42 thanks for the contribution. We'll need to evaluate if it's necessary to add a new field to types.go. We should be able to know the type of write-back target from the current WriteBackConfig.

Jyrno42:
@chengfang yeah i am happy if there's another way to do this, it was just a blocker for me personally so I decided to make a solution that resolves my personal issue.

Issue #1096, opened 2025-04-03
Multi-source `ApplicationSet` write-back error


rvandernoort:
**Describe the bug**

Hi, I'm trying to add multiple sources to an `ApplicationSet` from which I only want to update the versions of a single source using git-write back and kustomize.  Overall my setup works perfectly as long as I do not put the additional helm chart souces in the `ApplicationSet`, however I would like to join them. Is this currently possible or should I implement a workaround in which I put the redis instance in a seperate `ApplicationSet`?

**To Reproduce**

```yaml
apiVersion: argoproj.io/v1alpha1
kind: ApplicationSet
metadata:
  name: rover-test-1
spec:
  goTemplate: true
  goTemplateOptions: ['missingkey=error']
  generators:
    - list:
        elements:
          - environment: staging
          - environment: production
  template:
    metadata:
      name: rover-test-1-{{.environment}}
      annotations:
        argocd-image-updater.argoproj.io/image-list: rover-test=private-registry/rover-test-1
        argocd-image-updater.argoproj.io/rover-test.update-strategy: semver
        argocd-image-updater.argoproj.io/rover-test.allow-tags: regexp:^v[0-9]+.[0-9]+.[0-9]+$|^[a-f0-9]{40}$
        argocd-image-updater.argoproj.io/rover-test.ignore-tags: '^[a-f0-9]{40}$'
        argocd-image-updater.argoproj.io/write-back-method: git
        argocd-image-updater.argoproj.io/write-back-target: kustomization
    spec:
      project: hw-{{.environment}}
      destination:
        namespace: rover-ns-{{.environment}}
        name: cluster-name

      sources:
        - repoURL: 'git@github.com:organisation/gitops.git'
          path: ./deployments/rover-test/overlays/{{.environment}}
          targetRevision: main

        - repoURL: 'registry-1.docker.io/bitnamicharts'
          chart: redis
          targetRevision: 20.11.*
          helm:
            valueFiles:
              - $values/deployments/rover-test/base/redis/values.yaml
        - repoURL: 'https://github.com/orginsation/gitops.git'
          ref: values
          targetRevision: main
```

which eventually results in the following error message:

```
[...]

PR #1443, opened 2026-01-16
fix(write-back): support multi-sources containing both kustomize and helm

chengfang:
Fixes https://github.com/argoproj-labs/argocd-image-updater/pull/1417
Fixes https://github.com/argoproj-labs/argocd-image-updater/issues/1096

## Summary by CodeRabbit

* **New Features**
  * Write-back context awareness added across source selection and image operations for more accurate behavior.

* **Bug Fixes**
  * Improved multi-source selection: write-back targets now influence Helm vs Kustomize choice and fallback resolution.
  * Safer Helm parameter handling: missing params now fall back to available image name/tag data instead of failing.

* **Tests**
  * Expanded test coverage for multi-source detection, write-back targets, and fallback behaviors.


dkarpele:
I think you should update the description and add
"Fixes #1096"

pkg/argocd/argocd.go
```diff
@@ -924,12 +932,29 @@ func getApplicationSourceType(app *argocdapi.Application, wbc *WriteBackConfig)
...
+			}
+			// If source type is Kustomize, look for Kustomize source
+			if sourceType == argocdapi.ApplicationSourceTypeKustomize && s.Kustomize != nil {
+				return s
+			}
+		}
+
+		// Fallback: look for any Helm or Kustomize source
+		for i := range app.Spec.Sources {
+			s := &app.Spec.Sources[i]
```
dkarpele, line 955: Just a question. Why don't you use the original loop? It's shorter:
```
		for _, s := range app.Spec.Sources {
			if s.Helm != nil || s.Kustomize != nil {
				return &s
			}
		}
```

pkg/argocd/argocd.go
```diff
@@ -924,12 +932,29 @@ func getApplicationSourceType(app *argocdapi.Application, wbc *WriteBackConfig)
...
 // getApplicationSource returns the main source of a Helm or Kustomize type of the application
-func getApplicationSource(ctx context.Context, app *argocdapi.Application) *argocdapi.ApplicationSource {
+func getApplicationSource(ctx context.Context, app *argocdapi.Application, wbc *WriteBackConfig) *argocdapi.ApplicationSource {
 	log := log.LoggerFromContext(ctx)
 	if app.Spec.HasMultipleSources() {
-		for _, s := range app.Spec.Sources {
+		// Determine the source type from WriteBackConfig
+		sourceType := getApplicationSourceType(app, wbc)
+
+		for i := range app.Spec.Sources {
```
dkarpele, line 941: Just a question. Why don't you use the original loop? It's a bit shorter:
```
		for _, s := range app.Spec.Sources {
			if sourceType == argocdapi.ApplicationSourceTypeHelm && s.Helm != nil {
				return &s
			}
		}
```

pkg/argocd/argocd.go
```diff
@@ -924,12 +932,29 @@ func getApplicationSourceType(app *argocdapi.Application, wbc *WriteBackConfig)
...
+				return s
+			}
+			// If source type is Kustomize, look for Kustomize source
+			if sourceType == argocdapi.ApplicationSourceTypeKustomize && s.Kustomize != nil {
+				return s
+			}
+		}
+
+		// Fallback: look for any Helm or Kustomize source
+		for i := range app.Spec.Sources {
```
dkarpele, line 954: Why it's necessary to iterate again on `app.Spec.Sources`?

chengfang:
> Could you extend test coverage for `Test_GetApplicationSourceType` and `Test_GetApplicationSource` in `argocd_test.go` with the new test cases?

I can move Test_MultiSourceHelmAndKustomize_SourceTypeDetection from update_test.go to argocd_test.go. This test specifically verify GetApplicationSourceType and GetApplicationSource, and so argocd_test.go is a better place.

pkg/argocd/argocd.go
chengfang: Right, I'll change to shorter loop form.

pkg/argocd/argocd.go
chengfang: the first loop gets the source based on sourceType being helm or kustomize. If there is no wbc, and the 1st source element is neither helm nor kustomize, then the sourceType will be something else (directory or plugin). And the first loop will not give us a source, since the sourceType is not helm or kustomize. So the second loop will try again to get the first element that is helm or kustomize.

pkg/argocd/argocd.go
chengfang: Right, I'll change to shorter loop form.

chengfang:
> **Actionable comments posted: 0**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > pkg/argocd/argocd.go (1)> `905-916`: **Avoid misclassifying custom argocd-source YAML targets as Helm values.**
> > > The new `.yaml/.yml` heuristic treats any non-default YAML target as Helm. This can flip single-source Kustomize apps (or multi-source apps using a custom argocd-source override filename) to Helm, causing `marshalParamsOverride` to emit Helm values or empty output and skip Kustomize write-back. Consider gating the heuristic to multi-source apps and/or tracking an explicit `helmvalues:` origin flag in `WriteBackConfig`.
> > > 
> > > 🛠️ Suggested guard to reduce unintended reclassification
> > > ```diff
> > > -		if wbc.Target != "" && (strings.HasSuffix(wbc.Target, ".yaml") || strings.HasSuffix(wbc.Target, ".yml")) {
> > > +		if app.Spec.HasMultipleSources() && wbc.Target != "" &&
> > > +			(strings.HasSuffix(wbc.Target, ".yaml") || strings.HasSuffix(wbc.Target, ".yml")) {
> > >  			targetBase := filepath.Base(wbc.Target)
> > >  			if !strings.HasPrefix(targetBase, common.DefaultTargetFilePrefix) {
> > >  				return argocdapi.ApplicationSourceTypeHelm
> > >  			}
> > >  		}
> > > ```
> 
> 📜 Review details

The current condition should be safe, since wbc.Target is for helm values file target only. If the target of write-back is kustomize directory, wbc.Target will be "", and this condition will be false.

pkg/argocd/argocd.go
chengfang: WRT short loop form, if we use the short form like `for _, s := range app.Spec.Sources`, it would create a copy of the underlying element and assign to `s`, and `&s` would be a pointer to that copy, which is not what we want. So I think the original form is better.

pkg/argocd/argocd.go
chengfang: WRT short loop form, if we use the short form like `for _, s := range app.Spec.Sources`, it would create a copy of the underlying element and assign to `s`, and `&s` would be a pointer to that copy, which is not what we want. So I think the original form is better.

Issue #1446, opened 2026-01-22
Helm Manifest Targets name/tag required even when spec supplied


michaelfedell:
**Describe the bug**
When supplying `manifestTargets.helm.spec` for helm charts which take a single, canonical image name name for the values file input, the CRD validator throws an error indicating that `manifestTargets.helm.tag` and `manifestTargets.helm.name` are required.

**To Reproduce**

```yaml
apiVersion: argocd-image-updater.argoproj.io/v1alpha1
kind: ImageUpdater
metadata:
  annotations:
  name: image-updater-test
  namespace: argocd
spec:
  applicationRefs:
  - images:
    - alias: hello-world
      imageName: hello-world:latest
      manifestTargets:
        helm:
          spec: "image"
      commonUpdateSettings:
        forceUpdate: false
        updateStrategy: digest
    namePattern: hello-world
  namespace: argocd
  writeBackConfig:
    gitConfig:
      repository: https://github.com/org/image-updater-test
      writeBackTarget: helmvalues:/manifests/hello-world/prd/images.values.yaml
    method: git
```

**Expected behavior**

`manifestTargets.helm.spec` should be sufficient to identify the helm parameter to update as suggested by the Docs:
https://argocd-image-updater.readthedocs.io/en/stable/configuration/images/#specifying-helm-parameter-names

**Additional context**

**Version**

`1.0.2`

**Regression**
Does this issue apply to (check all that apply):

- [ ] Annotation-based configuration (versions starting with `0`, e.g., `0.17.0`)
- [X] CRD-based configuration (versions starting with `1`, e.g., `1.0.0`)

**Latest working version (if applicable)**
If you are reporting a regression, please specify the last version where this feature worked as expected.

**Logs**
Please paste any relevant logs here

```
imageupdaters.argocd-image-updater.argoproj.io "image-updater-test" was not valid:
* spec.applicationRefs[0].images[0].manifestTargets.helm.name: Required value
* spec.applicationRefs[0].images[0].manifestTargets.helm.tag: Required value
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
```

chengfang:
@michaelfedell thanks for reporting the bug. Yes, helm spec param should be supported without image-name and image-tag.

PR #1452, opened 2026-01-28
fix: Helm Manifest Targets name/tag should not be required when spec is set

dkarpele:
Closes: #1446

HelmTarget is valid if:
- spec is set, OR
- Both name and tag are set, OR
- Neither name nor tag are set (defaults will be used)

The HelmTarget is invalid if:
- Only name is set (without tag and without spec)
- Only tag is set (without name and without spec)

## Summary by CodeRabbit

* **Improvements**
  * Helm target Name and Tag are now optional; validation requires either Spec be set, or both Name and Tag together (or neither). When Spec is set, Name and Tag are ignored; defaults apply when neither path is used.
* **Documentation**
  * Updated field descriptions and CRD/install schema wording to reflect new conditional validation and defaulting behavior.
* **Tests**
  * Added validation tests covering allowed and rejected Helm target combinations.

### DIFF
--- a/api/v1alpha1/imageupdater_types.go
+++ b/api/v1alpha1/imageupdater_types.go
@@ -222,18 +222,20 @@ type ManifestTarget struct {
 }
 
 // HelmTarget defines parameters for updating image references within Helm values.
-// +kubebuilder:validation:XValidation:rule="has(self.spec) ? true : (has(self.name) && has(self.tag))",message="Either spec must be set, or both name and tag must be set together"
+// +kubebuilder:validation:XValidation:rule="has(self.spec) || (has(self.name) == has(self.tag))",message="Either spec must be set, both name and tag must be set together, or neither name nor tag must be set (defaults will be used)"
 type HelmTarget struct {
 	// Name is the dot-separated path to the Helm key for the image repository/name part.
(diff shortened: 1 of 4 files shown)


api/v1alpha1/imageupdater_types.go
```diff
@@ -222,22 +222,26 @@ type ManifestTarget struct {
 }
 
 // HelmTarget defines parameters for updating image references within Helm values.
+// +kubebuilder:validation:XValidation:rule="has(self.spec) ? true : (has(self.name) && has(self.tag))",message="Either spec must be set, or both name and tag must be set together"
```
chengfang: this new validation may cause errors for apps relying in default values of image name and tag. The current defaults: https://github.com/argoproj-labs/argocd-image-updater/blob/77f92be2551775e5337f508288fce9c5984d3bc4/pkg/common/constants.go#L17

dkarpele: Thanks, fixed. This config is now valid:
```
    - alias: hello-world
      imageName: hello-world:latest
      manifestTargets:
        helm: {}  # Empty object - defaults to image.name and image.tag
```

chengfang (approved):

PR #1458, opened 2026-01-31
fix: improve webhook processing with non-existing or empty type

### LINKED ISSUE #1389: Incorrect error message when sending webhook request with non-existing type
**Describe the bug**
When user sends a webhook request with non-existing type it returns incorrect error message. 

**To Reproduce**
1. Set up webhook server (in my example on localhost)
2. Send a request with non-existing type
```
curl -X POST \
  -H "Content-Type: application/json" \
  -d '{
  "specversion": "1.0",
  "id": "1",
  "type": "image",
...
  }
}' \
  "http://localhost:8080/webhook?type=doesntexist"    
repository name not found in webhook payload
```
3. Send the same request again and see error message
```
no tags in the payload
```
4. and again
```
unsupported event type: image
```
5. again - the worst case
```
Webhook received and processing
```

**Expected behavior**
Show a clear error message and stop execution.

**Additional context**

**Version**
1.0.1
0.17

**Regression**
Does this issue apply to (check all that apply):

- [x] Annotation-based configuration (versions starting with `0`, e.g., `0.17.0`)

dkarpele:
Closes #1389 

I have removed the for loop below because it's buggy and never used. If type was empty or handler doesn't exist this loop always fails on the `Validate` or `Parse` function or always passes only for `docker.io` handler. `h.handlers` map is not ordered that's way the first handler is always random. Moreover doc requires `type` in the [webhook url](https://argocd-image-updater.readthedocs.io/en/latest/configuration/webhook/#setting-up-a-webhook-notification). So the config without `type` is wrong.
```
	for _, handler := range h.handlers {
		if err := handler.Validate(r); err == nil {
			return handler.Parse(r)
		}
	}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer error responses when the registry type is missing or unsupported.
  * Webhook type identifier standardized from "docker" to "docker.io".

* **Improvements**
  * Validation now enumerates supported registry types in error messages for deterministic feedback.

* **Tests**
  * Stabilized webhook tests to avoid parallel-test port conflicts and updated expectations to match the new validation and webhook type.

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -54,16 +54,12 @@ func (h *WebhookHandler) ProcessWebhook(r *http.Request) (*argocd.WebhookEvent, 
 		return nil, fmt.Errorf("invalid registry type: %s. Supported types: %s", registryType, strings.Join(registryTypes, ", "))
 	}
 
-	// If handler exists for the specified type, use it
-	if handler, exists := h.handlers[registryType]; exists {
-		if err := handler.Validate(r); err != nil {
-			return nil, err
-		}
-		return handler.Parse(r)
+	// Handler is guaranteed to exist after the validation above
+	handler := h.handlers[registryType]
+	if err := handler.Validate(r); err != nil {
+		return nil, err
 	}
-
-	// Valid type but no handler registered - return clear error
-	return nil, fmt.Errorf("no handler registered for registry type: %s", registryType)
+	return handler.Parse(r)
 }
 
 // getSupportedRegistryTypes returns a list of all supported registry types from registered handlers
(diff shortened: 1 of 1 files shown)


pkg/webhook/webhook.go
```diff
@@ -39,21 +41,38 @@ func (h *WebhookHandler) ProcessWebhook(r *http.Request) (*argocd.WebhookEvent,
...
+	registryTypes := h.getSupportedRegistryTypes()
+
+	// Registry type is required
+	if registryType == "" {
+		return nil, fmt.Errorf("missing registry type parameter. Supported types: %s", strings.Join(registryTypes, ", "))
+	}
+
+	// Validate the registry type
+	if !slices.Contains(registryTypes, registryType) {
+		return nil, fmt.Errorf("invalid registry type: %s. Should be one of %s", registryType, strings.Join(registryTypes, ", "))
```
chengfang: how about making the above 2 errors have similar messages, as close as possible for easy search? `supported types...` seems better.

dkarpele: fixed

dkarpele:
image

PR #1473, opened 2026-02-06
fix: Do not validate the current image tag against SemVer

christian-schlichtherle:
In our Helm charts, we use `latest` as a reasonable default value for most image tags. We want to use CD with SemVer constraints for these images however. Without this change, that's not possible because the Argo CD Image Updater errors with "Unable to find newest version from available tags: invalid semantic version" if it finds the current version to be violating the semantic version constraint.

I think validating the current image tag against SemVer doesn't add any value, it just adds an impediment, so this PR removes it for good. With this change, you can now upgrade from `latest` (or whatever) to `1.2.3` etc.

## Summary by CodeRabbit

* **Refactor**
  * Simplified version-constraint handling so semver constraints are applied only when explicitly provided, preserving previous behavior when no constraint exists.
* **Tests**
  * Added a test verifying the newest semver-matching version is chosen even if the current tag is non-semver.
* **Documentation**
  * Updated docs and changelog: current tags no longer must be semver for semver-based updates; examples updated.

### DIFF
--- a/ext/git/client.go
+++ b/ext/git/client.go
@@ -718,7 +718,7 @@ func (m *nativeGitClient) VerifyCommitSignature(ctx context.Context, revision st
 
 // IsAnnotatedTag returns true if the revision points to an annotated tag
 func (m *nativeGitClient) IsAnnotatedTag(ctx context.Context, revision string) bool {
-	cmd := exec.Command("git", "describe", "--exact-match", revision)
+	cmd := exec.CommandContext(ctx, "git", "describe", "--exact-match", revision)
 	out, err := m.runCmdOutput(ctx, cmd, runOpts{SkipErrorLogging: true})
 	if out != "" && err == nil {
 		return true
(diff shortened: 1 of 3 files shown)


christian-schlichtherle:
I'm using my fork in our clusters and it works like a charm: We have dozens of images which we simply tag as `latest` and now they get properly upgraded.

christian-schlichtherle:
I've updated the docs and the CHANGELOG.md, too.

registry-scanner/pkg/image/version.go
```diff
@@ -106,24 +106,11 @@ func (img *ContainerImage) GetNewestVersionFromTags(ctx context.Context, vc *Ver
...
-
-		if vc.Constraint != "" {
-			if vc.Strategy == StrategySemVer {
-				semverConstraint, err = semver.NewConstraint(vc.Constraint)
-				if err != nil {
-					logCtx.Errorf("invalid constraint '%s' given: '%v'", vc, err)
-					return nil, err
-				}
-			}
+	if vc.Strategy == StrategySemVer && vc.Constraint != "" {
```
chengfang, line 109: Can you also add a unit test for the case where the version constraint is empty (`vc.Constraint == ""`), and current tag is a non-semver? Similar to the existing subtest 'Find the latest version without any constraint' around line version_test.go:32

christian-schlichtherle: Done - please review.

Issue #1445, opened 2026-01-22
Git Process ArgoCD image exhaustion


larivierec-psl:
**Describe the bug**

After the image-updater runs for some time, git operations fail with errors like:

error: cannot fork() for remote-https: Resource temporarily unavailable
error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500\nfatal: expected flush after ref listing

Observed inside the container there are 8,141 git processes, and further git commands fail with exit status 255 or 128. This appears to be either process/fork() exhaustion or leaking/unreaped git helper processes.

**To Reproduce**

- Configure Argo CD Image Updater to use a GitHub App for repository authentication (the GitHubAppCreds flow).
- Run the updater against many applications or let it run for extended time with default concurrency.
- Monitor processes inside the pod: `ps aux | grep git | wc -l`
- Eventually you will see process count grow (observed 8141) and git fetch/push will start failing with the errors above.

```bash
~ $ ps aux | grep git | wc -l
8239
```

**Expected behavior**

Git fetch/push operations should succeed and not exhaust container process limits or leak helper processes. The updater should manage concurrency and clean up child processes reliably.

**Additional context**
N/A

**Version**

v1.0.2

**Regression**
Does this issue apply to (check all that apply):

- [X] CRD-based configuration (versions starting with `1`, e.g., `1.0.0`)

**Latest working version (if applicable)**
If you are reporting a regression, please specify the last version where this feature worked as expected.

**Logs**

```bash
time="2026-01-21T17:08:57Z" level=error msg="Could not update application spec: `git fetch origin main --force --prune --depth 1` failed exit status 255: error: cannot fork() for remote-https: Resource temporarily unavailable" application=argocd/ami-beiditive-app-staging controller=imageupdater controllerGroup=argocd-image-updater.argoproj.io controllerKind=ImageUpdater imageUpdater_name=beiditive-app-staging-updater-app-template-image-updater imageUpdater_namespace=argocd logger=reconcile
```

and a different log most likely similar or related

```
time="2026-01-22T14:14:27Z" level=error msg="`git fetch origin main --force --prune --depth 1` failed exit status 128: error: RPC failed; HTTP 500 curl 22 The requested URL returned error: 500\nfatal: expected flush after ref listing" execID=b47bb
[...]

PR #1477, opened 2026-02-09
fix: implement git command process cleanup and use context-aware git command execution

chengfang:
Fixes https://github.com/argoproj-labs/argocd-image-updater/issues/1445

## Summary by CodeRabbit

* **Bug Fixes**
  * Git commands now honor cancellation requests, reducing hangs and improving timeout handling.
  * Improved cleanup of subprocesses by ensuring entire command groups are terminated to prevent orphaned processes and PID buildup.
  * Credential-invoked Git operations also support context-aware cancellation for more reliable remote interactions.

### DIFF
--- a/ext/git/client.go
+++ b/ext/git/client.go
@@ -718,7 +718,7 @@ func (m *nativeGitClient) VerifyCommitSignature(ctx context.Context, revision st
 
 // IsAnnotatedTag returns true if the revision points to an annotated tag
 func (m *nativeGitClient) IsAnnotatedTag(ctx context.Context, revision string) bool {
-	cmd := exec.Command("git", "describe", "--exact-match", revision)
+	cmd := exec.CommandContext(ctx, "git", "describe", "--exact-match", revision)
 	out, err := m.runCmdOutput(ctx, cmd, runOpts{SkipErrorLogging: true})
 	if out != "" && err == nil {
 		return true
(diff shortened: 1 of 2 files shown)


ext/git/client.go
```diff
@@ -828,5 +833,15 @@ func (m *nativeGitClient) runCmdOutput(ctx context.Context, cmd *exec.Cmd, ropts
...
 	}
-	return executil.RunWithExecRunOpts(cmd, opts)
+	output, err := executil.RunWithExecRunOpts(cmd, opts)
+
+	// After the git command finishes (normally or via timeout/context cancellation),
+	// kill the entire process group to clean up any orphaned child processes such as
+	// git-remote-https. Without this, child processes accumulate over time and
+	// eventually exhaust the container's PID limit, causing "cannot fork()" errors.
+	if cmd.Process != nil {
+		_ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
```
olivergondza, line 847: Sorry if I missed something obvious, is the PID negated on a purpose here?

olivergondza: This idiom what works for me decently: https://github.com/olivergondza/argo-dev-tools/commit/6099bc7f161a6d6401802a8636af75bbe1e40f79#diff-e1ef05b601cdc463a7dfa9f237ecc178ce3040ef21f29bfc45677020aa4f8b6bR47

It favors sigterm over sigkill to give processes the chance for responsible cleanup

chengfang: this is to send signal to all processes in the process group by using a process group id that is < -1.

In our case, the current process in question is already terminated, since the above block is already after the command execution call.  The kill is to terminate any potentially remenant, orphaned child processes.

ext/git/client.go
dkarpele: Does it makes sense to replace it with 
`	cmd := exec.CommandContext(ctx, wrapper, args...)`?

ext/git/client.go
dkarpele: Could you add this to the comment section above?

olivergondza: Ah. Thanks

PR #1490, opened 2026-02-11
fix: deprecate spec.namespace

### LINKED ISSUE #1381: Prevent users from referencing Applications in any namespace
We are running ArgoCD in a multi-tenant setup with the App-of-Apps pattern allowing users to create their own ArgoCD Application resources in their own namespaces so they can take full ownership. We limit users access by setting up ArgoCD AppProjects so they can only create Applications which deploy in the user's namespaces.

The problem with our setup and argocd-image-updater is that users can reference any application in any namespace via the `.spec.namespace` field as far as I am concerned. This means that a user could set up an ImageUpdater resource pointing to an Application which is not in their allowed namespaces and trigger an image update of one of the resources managed by said ArgoCD Application resource. Of course, this probably requires harmful intentions and a little bit of insider knowledge about the resources in other namespaces, but nevertheless this could be a risk.

Enable admins to set a flag in the argocd-image-updater-controller to only allow ImageUpdate resources which point to Applications which live in the same namespace as the ImageUpdater resource itself.

If I am correct, this should be enough to close this issue.

I already tried to set up an Application in a user's namespace which is not allowed to reconcile based on the ArgoCD RBAC. When I set up an ImageUpdater resource in the same namespace then the Image Updater should not consider that application as it has no managed resources resulting in `Image 'nginx/nginx-unprivileged' seems not to be live in this application, skipping`. So this means to me that a simple check if `.metadata.namespace` and `.spec.namespace` are in sync should be sufficient.

- The platform team could manage the ImageUpdater resources. But we don't like that solution as we don't want to be the bottleneck for all the devs on our platform. This defeats the purpose of enabling users.
- Having a dedicated image-updater running in every users namespace with limited access (I did not test that). I would like to prevent that.
- Running a policy engine which only allows to reference Applications in the same namespace as the ImageUpdater resource (again, I did not test that). Introducing a policy engine for that use case seems to be a little far fetched.

**Version**
1.0.1

dkarpele:
Closes #1381

It was a bug in `argocd.go`. Application namespace should go from Application not from ImageUpdater. I replaced
`appNSName := fmt.Sprintf("%s/%s", cr.Spec.Namespace, app.Name)`
with
`appNSName := fmt.Sprintf("%s/%s", app.Namespace, app.Name)`.
Though they are always the same.

## Summary by CodeRabbit

* **Breaking Changes**
  * spec.namespace is deprecated/ignored; the controller now scopes applications by the ImageUpdater CR's metadata.namespace — configs relying on a differing spec.namespace will break.

* **Documentation**
  * Guides, examples, migration steps and field descriptions updated to reflect metadata.namespace usage and deprecation guidance.

* **Bug Fixes**
  * CRD schema and sample manifests updated to remove required spec.namespace and mark it deprecated; sample YAML structure issues addressed.

* **Tests**
  * Tests and e2e manifests updated to use the CR's metadata.namespace and stricter namespace filtering in the test harness.


internal/controller/reconcile_test.go
```diff
@@ -29,6 +29,42 @@ import (
...
+			return nil
+		}
+		// For other list types, we could use reflection, but since we primarily test
+		// with ApplicationList, this explicit handling is sufficient.
+		// If other types are needed, we can extend this.
+	}
+
+	// Delegate to underlying client for non-empty namespaces
+	return c.Client.List(ctx, list, opts...)
+}
```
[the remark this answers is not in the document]
dkarpele: The comment is clear enough for test.

chengfang:
Shall we add tests for the case where spec.namespace is present and equal to metadata.namespace (should still work), or differ from metadata.namespace (should fail)?

dkarpele:
> Shall we add tests for the case where spec.namespace is present and equal to metadata.namespace (should still work), or differ from metadata.namespace (should fail)?

Of course, I will add them. Later, when we remove `spec.namespace` we must delete these tests. Otherwise the code will not compile.

dkarpele:
@coderabbitai review

docs/index.md
```diff
@@ -4,6 +4,9 @@ A tool to automatically update the container images of Kubernetes workloads
 that are managed by
 [Argo CD](https://github.com/argoproj/argo-cd).
 
+!!!warning "Breaking Change: spec.namespace Field Deprecated"
+    The `spec.namespace` field is now optional and deprecated. The controller now uses the ImageUpdater CR's `metadata.namespace` to determine which namespace to search for applications. This is a breaking change for configurations where `spec.namespace` differs from `metadata.namespace`. For details and migration examples, see [Namespace Field Deprecation](#namespace-field-deprecation) below.
```
chengfang: nit: Should we also emphasize that it will be removed in a future release, just as in imageupdater_types.go comment?

dkarpele: fixed

chengfang (approved):

PR #1491, opened 2026-02-13
fix: deprecate spec.namespace

dkarpele:
(cherry picked from commit 1c19a38d986ab7ffd1c74c592bda7cb46261ca59)

## Summary by CodeRabbit

* **Breaking Changes**
  * The `spec.namespace` field in ImageUpdater specs has been deprecated and is now optional and ignored. The controller now uses the ImageUpdater resource's metadata.namespace for application scoping. Existing manifests with `spec.namespace` should have this field removed.

* **Documentation**
  * Updated with migration guidance for the namespace field deprecation and new namespace scoping behavior.

### DIFF
--- a/test/e2e/suite/006-apps-in-any-namespace/01-install.yaml
+++ b/test/e2e/suite/006-apps-in-any-namespace/01-install.yaml
@@ -93,7 +93,7 @@ metadata:
     app.kubernetes.io/name: image-updater
     app.kubernetes.io/managed-by: kustomize
   name: image-updater-006
-  namespace: argocd-image-updater-e2e
+  namespace: image-updater-e2e-006-01
 spec:
   commonUpdateSettings:
     updateStrategy: "semver"


dkarpele:
@coderabbitai review

test/e2e/suite/006-apps-in-any-namespace/01-install.yaml
```diff
@@ -95,7 +95,6 @@ metadata:
   name: image-updater-006
   namespace: argocd-image-updater-e2e
 spec:
-  namespace: image-updater-e2e-006-01
```
chengfang, line 98: for this 006-apps-in-any-namespace test, the 2 namespace are different. After removing spec.namespace, image updater will search apps only in meta.namespace, which most likely does not contain the app. Can you make sure this test only uses one namespace: meta.namespace?

dkarpele: should be good now

Issue #1242, opened 2025-09-19
Image updater removes tags when registry is empty.


kinikiti:
**Describe the bug**
We are running EKS cluster with ECR pull-trough cache which caches images from Docker hub.
Image updater looks at the upstream and update tags in manifest, which later pulled by AWS to pull-trought cache.

Recently in the upstream registry all tags were removed. This caused that Image updater put `image.tag: ""` to manifests.
This is incorrect notation. It would be nice if Image Updater just log error and do not change tags to meaningless values if tags are not found.

**To Reproduce**
Setup Image update pointing to the registry.
Remove all tags from registry.

**Expected behavior**
Image updater do nothing.

**Additional context**
This was particularly caused by changes in Bitnami registries which were converted from SemVer notation to Digest and during conversion there were no tags for couple of days.

**Version**
Image Updater v0.16.0

PR #1505, opened 2026-02-19
fix: don't write empty tag when registry returns no tags

mark-liu:
Fixes https://github.com/argoproj-labs/argocd-image-updater/issues/1242

When a registry returns zero tags (happened during Bitnami's SemVer → Digest migration), `GetNewestVersionFromTags` returned the current `ImageTag` which could be a dummy with empty `TagName`. This bypassed the nil guard in `UpdateApplication` and wrote `image.tag: ""` to the manifest.

Fix is one line — return `nil` instead of `img.ImageTag` when no tags are available. The existing skip logic in `UpdateApplication` already handles nil correctly.

* Also changes behavior for semver strategy when ALL tags are non-semver (e.g. `latest`, `alpine`) — these were silently treated as "up-to-date", now correctly classified as "skipped" with a warning log. No update happened either way, but users may see new `WARN no tags found` messages for misconfigured images.

### Related

This PR fixes the "registry returns no tags" path into empty tag write-back. A separate code path exists in `marshalParamsOverride` (`update.go:~438`) where `GetTagWithDigest()` returns `""` when a pod scales to zero — reported in #1523 by @Resousse. That requires a separate guard and should be addressed in its own PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when image registries return no tags or no suitable versions are found.
  * Updated skip behavior to correctly identify and handle edge cases with empty tag responses.
  * Enhanced diagnostic logging when registries return no tags or when version constraints have no matching tags.

### DIFF
--- a/registry-scanner/pkg/image/version.go
+++ b/registry-scanner/pkg/image/version.go
@@ -100,7 +100,11 @@ func (img *ContainerImage) GetNewestVersionFromTags(ctx context.Context, vc *Ver
 
 	// It makes no sense to proceed if we have no available tags
 	if len(availableTags) == 0 {
-		logCtx.Warnf("no tags found for image %s in registry", img.GetFullNameWithoutTag())
+		if len(tagList.Tags()) > 0 {
+			logCtx.Warnf("no tags for image %s matched by the %s strategy", img.GetFullNameWithoutTag(), vc.Strategy)
+		} else {
+			logCtx.Warnf("no tags found for image %s in registry", img.GetFullNameWithoutTag())
+		}
 		return nil, nil
 	}
 
@@ -154,6 +158,7 @@ func (img *ContainerImage) GetNewestVersionFromTags(ctx context.Context, vc *Ver
 		return considerTags[len(considerTags)-1], nil
 	}
 
+	logCtx.Warnf("no tags for image %s matched constraint %q", img.GetFullNameWithoutTag(), vc.Constraint)
 	return nil, nil
 }
 


mark-liu:
Friendly ping — this fixes a data loss edge case where an empty tag gets written. Would appreciate a review when you get a chance.

Resousse:
Hello @mark-liu , I've reported a similar bug, but more linked to a pod scale-down (rather than a registry giving no tag), resulting an empty image tag, do you know by any chance if your implementation will solve this too? Thanks

mark-liu:
Hey @Resousse, thanks for linking #1523.

I suspect so.

When the pod scales to zero, ArgoCD's Status.Summary.Images either goes empty or retains the stale image ref. In both cases the path through `GetNewestVersionFromTags` is what determines whether a write-back happens:

  - Images empty, ForceUpdate=false (the default): image isn't found in the live list, skipped at the ContainsImage check. No write-back.
  - Images empty, ForceUpdate=true: a dummy ImageTag with empty TagName gets created, but `GetNewestVersionFromTags` still runs against the registry. Pre-fix, if no tags matched, it returned that dummy empty tag and the write-back proceeded with "". Post-fix, it returns nil and the image is skipped.
  - Images still has stale ref: proceeds normally with the real tag — no issue.

There's also a defence-in-depth guard in SetHelmImage that skips setting the tag parameter when GetTagWithDigest() is empty, so even if something slips through, the empty value shouldn't reach your values file.

Not sure if you can, but could you try this branch against your setup and confirm it stops the loop?

Resousse:
@mark-liu  I've tested your branch, and unfortunately, it doesn't solve my issue. As soon as, the scale down is done, an empty tag "" is write back in git causing the loop again :/

Resousse:
Ok, I found the issue, in file `update.go`, around line 440:
`err = setHelmValue(&helmNewValues, helmParamVersion, tagValue)`
tagValue when retrieved from `c.ContainerImage.GetTagWithDigest` can be an empty string, if the container is not running anymore (my scale down case)
To avoid writing an empty string, I've just added a condition on `tagValue` to not update the tag, when the target tag is empty.

```
if tagValue != "" {
		err = setHelmValue(&helmNewValues, helmParamVersion, tagValue)
		if err != nil {
			return nil, fmt.Errorf("failed to set image parameter version value: %v", err)
		}
	}
```
I've tested this based on your branch, and it works as expected

@mark-liu do you mind adding this code to your PR, as it's related to empty tag ?
Thank you very much

mark-liu:
Thanks for the investigation @Resousse. You've found a separate code path — this PR guards the "registry returns no tags" entry point (`version.go`), but your scale-down case enters through `marshalParamsOverride` (`update.go:~438`) where `GetTagWithDigest()` returns `""` on an un-aliased config image.

Your proposed guard looks good to me:

```go
if tagValue != "" {
    err = setHelmValue(&helmNewValues, helmParamVersion, tagValue)
}
```

When the file already exists, `helmNewValues` is populated from the original data (line 403-404), so skipping the overwrite preserves the previous good tag. When the file doesn't exist yet, omitting an empty tag is the right behavior.

Worth noting: the ArgoCD write-back path already has this guard at `argocd.go:766` — only the git write-back custom target path is missing it.

Since this is a different trigger and code path from #1242, it should be a separate PR linked to #1523. Would you like to open one? The fix is a one-line guard at `update.go:438`.

Resousse:
Sure, just created this : https://github.com/argoproj-labs/argocd-image-updater/pull/1531

chengfang (approved):

Issue #1523, opened 2026-02-28
Empty "" image tag git update after a pod scale down, in an infinite loop


Resousse:
**Describe the bug**
I'm using Image Updated with Knative, with a min scale of 0, which means, pod is removed after the scale down.
I'm using git write back policy to update a value file.
Each time the scale down occurs, Image tag is set to "", then Argo CD will see that as a change, and will create a new revision, that will wake up a pod for few minutes, and set the tag to a valid sha code, and then looping few minutes after, with the scale down

**To Reproduce**
Knative setup is pretty complex, but I think deleting any scale-up/down model can result the same thing

**Expected behavior**
Once a tag is assigned, no "" must be written back
Will this be fixed by : https://github.com/argoproj-labs/argocd-image-updater/pull/1505 ? 

**Additional context**
N/A

**Version**
`1.1.1`

**Regression**
Does this issue apply to (check all that apply):

- [ ] Annotation-based configuration (versions starting with `0`, e.g., `0.17.0`)
- [X] CRD-based configuration (versions starting with `1`, e.g., `1.0.0`)

**Latest working version (if applicable)**
N/A

**Logs**
There is no log about setting empty tag value, but here several logs event in few minutes setting the same value for the image tag
```
time="2026-02-28T18:11:36Z" level=info msg="Successfully updated image 'privateregistry/registry/auth@sha256:413aec3f3b106746aafc4d7997fd0bf98aac0b7ab0b10c44db60c85dedad299c' to 'privateregistry/registry/auth:sha-4bd1d59', but pending spec update (dry run=false)" application=management/micro-services controller=imageupdater controllerGroup=argocd-image-updater.argoproj.io controllerKind=ImageUpdater imageUpdater_name=microservices-image-updated imageUpdater_namespace=management image_alias=auth image_digest="sha256:413aec3f3b106746aafc4d7997fd0bf98aac0b7ab0b10c44db60c85dedad299c" image_name=privateregistry/registry/auth image_registry=privateregistry logger=reconcile
[...]

PR #1531, opened 2026-03-02
fix: don't write empty tag in helm values file

Resousse:
Closes https://github.com/argoproj-labs/argocd-image-updater/issues/1523
Do not write tag in helm chart values file, if the target tag is an empty string.
A similar logic is already implemented for Application write-back

## Summary by CodeRabbit

* **Bug Fixes**
  * Helm image parameter handling now ignores empty tag values, preventing errors and avoiding accidental overrides so existing image tags are preserved.

* **Tests**
  * Added tests verifying empty Helm tag inputs do not introduce tag keys and that existing tag values remain unchanged during parameter marshaling.

### DIFF
--- a/pkg/argocd/update_test.go
+++ b/pkg/argocd/update_test.go
@@ -3467,6 +3467,68 @@ replicas: 1
 		require.True(t, ok, "expected key %s to still exist", "image.tag")
 		assert.Equal(t, "v1.0.0", val)
 	})
+
+	t.Run("Empty target tag should not override a non existing helm tag", func(t *testing.T) {
+		app := v1alpha1.Application{
+			ObjectMeta: v1.ObjectMeta{
+				Name: "testapp",
+			},
+			Spec: v1alpha1.ApplicationSpec{
+				Source: &v1alpha1.ApplicationSource{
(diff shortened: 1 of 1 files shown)


dkarpele:
Thanks for the contribution @Resousse!
Could you please address the review from Coderabbit (`update_test.go`)?
> pkg/argocd/update.go (1)
439-445: Please add regression tests for the empty-tag branch.
Since Line 440 now skips setHelmValue, add coverage for: (1) empty tag does not create a new key, and (2) behavior when the tag key already exists in originalData is explicitly asserted.

Resousse:
> Thanks for the contribution @Resousse! Could you please address the review from Coderabbit (`update_test.go`)?
> 
> > pkg/argocd/update.go (1)
> > 439-445: Please add regression tests for the empty-tag branch.
> > Since Line 440 now skips setHelmValue, add coverage for: (1) empty tag does not create a new key, and (2) behavior when the tag key already exists in originalData is explicitly asserted.

@dkarpele , thanks! I've added two tests for this. Let me know if it requires improvement.

dkarpele (approved):

dkarpele:
LGTM, Thanks!

PR #1587, opened 2026-04-14
fix(webhook): align standalone webhook HTTP/2 hardening

doup2001:
# Summary

This change threads the `enable-http2` setting through the webhook configuration so the standalone webhook server follows the same HTTP/2 hardening policy as the controller-managed path.

When HTTP/2 is not explicitly enabled, the webhook TLS server now restricts ALPN negotiation to `http/1.1`. A standalone `webhook` command flag was also added so users can opt in to HTTP/2 intentionally.

Closes #1559

# What Changed

- added `EnableHTTP2` to `WebhookConfig`
- added `--enable-http2` to the standalone `webhook` command
- passed the controller's `enableHTTP2` setting into webhook server setup
- updated webhook TLS config building to disable HTTP/2 by default via `NextProtos = []string{"http/1.1"}`
- added regression coverage for flag defaults, config threading, and TLS config behavior

# User Impact

- standalone webhook mode now matches the project's existing HTTP/2 hardening behavior
- HTTP/2 remains disabled by default
- users can explicitly opt in with `--enable-http2`

## Summary by CodeRabbit

* **New Features**
  * Added a CLI flag --enable-http2 to explicitly control HTTP/2 for the webhook server (disabled by default). When disabled, the server will use HTTP/1.1.

* **Documentation**
  * Updated run and webhook docs to describe the --enable-http2 flag and its default.

* **Tests**
  * Added/expanded tests verifying the flag’s default and that the HTTP/2 setting is propagated when enabled.


doup2001:
To maintain a clean commit history and avoid unnecessary merge commits, I’ve rebased my changes and opened a new PR at #1588 . Closing this one now—thanks for understanding

PR #1608, opened 2026-04-28
fix: inherit global writeBackConfig for annotation-based applicationRefs

### LINKED ISSUE #1578: applicationRefs.writeBackConfig does not inherit global writeBackConfig.method/gitConfig when only gitConfig.writeBackTarget is overridden
**Describe the bug**
When using the ImageUpdater CRD with a global spec.writeBackConfig and an application-level spec.applicationRefs[].writeBackConfig that only overrides gitConfig.writeBackTarget, the application-level config does not appear to inherit the global method, repository, and branch.

In my case, this causes argocd-image-updater to update the live Application spec repeatedly, but not create a Git commit. If I repeat the full writeBackConfig block under applicationRefs[], it works correctly and commits to Git.

Based on the CRD model and the behavior implied by the config shape, I would expect applicationRefs[].writeBackConfig to behave as a partial override of the global spec.writeBackConfig, not require repeating the full block.

**To Reproduce**
I have an ApplicationSet generating Applications like this:
```
kind: ApplicationSet
metadata:
  name: sample-app
  namespace: argocd
spec:
  goTemplate: true
  goTemplateOptions:
  - missingkey=error
  generators:
  - matrix:
      generators:
      - git:
          repoURL: git@github.com:example-org/gitops-repo.git
          revision: HEAD
          files:
          - path: instances/sample-app/config.yaml
      - list:
          elementsYaml: "{{ .instances | toJson }}"
  template:
    metadata:
      name: "sample-app-{{ .namespace }}-{{ .clusterKey }}"
    spec:
      project: default
      sources:
      - repoURL: git@github.com:example-org/gitops-repo.git
        targetRevision: HEAD
        path: charts/sample-app
        helm:
          releaseName: "{{ .namespace }}"
          valueFiles:
          - $values/instances/sample-app/values/common.yaml
          - $values/instances/sample-app/values/{{ .clusterKey }}/{{ .namespace }}.yaml
      - repoURL: git@github.com:example-org/gitops-repo.git
        targetRevision: HEAD
        ref: values
      destination:
        name: "{{ .cluster }}"
        namespace: "{{ .namespace }}"
```

The chart contains:
```
# charts/sample-app/values.yaml
image:
  name: registry.example.com/example-project/example-repo/sample-app
  tag: a1c2c3e
```

The per-instance values files are empty.

**Works correctly**
```
kind: ImageUpdater
metadata:
  name: sample-app
  namespace: argocd
spec:
  writeBackConfig:
    method: git:secret:argocd/image-updater-git-creds
    gitConfig:
      repository: git@github.com:example-org/gitops-repo.git
      branch: main
  applicationRefs:
  - namePattern: "sample-app-*"
    images:
    - alias: sample-app
      imageName: registry.example.com/example-project/example-repo/sample-app
      commonUpdateSettings:
        updateStrategy: newest-build
        allowTags: "regexp:^[0-9a-f]{7}$"
      manifestTargets:
        helm:
          name: image.name
          tag: image.tag
    writeBackConfig:
      method: git:secret:argocd/image-updater-git-creds
      gitConfig:
        repository: git@github.com:example-org/gitops-repo.git
        branch: main
        writeBackTarget: helmvalues:./values.yaml
```
This version creates a Git commit successfully.

**Does not work**
```
kind: ImageUpdater
metadata:
  name: sample-app
  namespace: argocd
spec:
  writeBackConfig:
    method: git:secret:argocd/image-updater-git-creds
    gitConfig:
      repository: git@github.com:example-org/gitops-repo.git
      branch: main
  applicationRefs:
  - namePattern: "sample-app-*"
    images:
    - alias: sample-app
      imageName: registry.example.com/example-project/example-repo/sample-app
      commonUpdateSettings:
        updateStrategy: newest-build
        allowTags: "regexp:^[0-9a-f]{7}$"
      manifestTargets:
        helm:
          name: image.name
          tag: image.tag
    writeBackConfig:
      gitConfig:
        writeBackTarget: helmvalues:./values.yaml
```

In this version, I expected the application-level writeBackConfig to inherit:
	•	method: git:secret:argocd/image-updater-git-creds
	•	gitConfig.repository
	•	gitConfig.branch

from the global spec.writeBackConfig, and only override gitConfig.writeBackTarget.

Instead, there is no Git commit. The controller logs show that it updates the image and the live application spec, but the change is not committed to Git.

**Observed behavior**

Logs look like this:
```
Successfully fetched ImageUpdater resource.
Listing all applications in target namespace: argocd
Starting image update cycle, considering 1 application(s) for update
Setting new image to registry.example.com/example-project/example-repo/sample-app:b1b2d3f
Successfully updated image 'registry.example.com/example-project/example-repo/sample-app:a1c2c3e' to 'registry.example.com/example-project/example-repo/sample-app:b1b2d3f', but pending spec update (dry run=false)
Committing 1 parameter update(s) for application sample-app-dev-c1
Successfully updated application spec for sample-app-dev-c1
Successfully updated the live application spec
Processing results: applications=1 images_considered=1 images_skipped=0 images_updated=1 errors=0
```
Then on the next cycle it tries to do the same update again, which suggests the Git state did not change.

**Expected behavior**
I would expect spec.applicationRefs[].writeBackConfig to act as a partial override of the global spec.writeBackConfig, so that setting only:
```
writeBackConfig:
  gitConfig:
    writeBackTarget: helmvalues:./values.yaml
```

**Additional context**

**Version**
v1.1.1

**Regression**
Does this issue apply to (check all that apply):

- [ ] Annotation-based configuration (versions starting with `0`, e.g., `0.17.0`)

wucm667:
## Summary

When `applicationRefs[].writeBackConfig` is set via annotations, it does not inherit fields from the global `spec.writeBackConfig` (method, gitConfig repository, branch, etc.). Instead, it treats the application-level config as a separate object with its own defaults, causing `method` to fall back to `argocd` even when a global `git` method is configured.

## Root Cause

In `pkg/argocd/argocd.go`, the annotation-based path (line 301-302) calls `getWriteBackConfigFromAnnotations()` and passes the result directly to `newWBCFromSettings()` without merging against the global config. Meanwhile, the CR-based path (line 324) correctly uses `mergeWBCSettings()` to merge global and app-level configs.

## Fix

Added a single line to merge the annotation-derived config with the global `cr.Spec.WriteBackConfig`:

```go
appRefWBC := getWriteBackConfigFromAnnotations(&app)
appRefWBC = mergeWBCSettings(cr.Spec.WriteBackConfig, appRefWBC)  // NEW
appWBCSettings, err = newWBCFromSettings(appCtx, &app, kubeClient, appRefWBC)
```

This makes the annotation-based path behave consistently with the CR-based path.

## Testing

- All existing `Test_mergeWBCSettings` tests pass (11/11)
- Full `pkg/argocd` test suite passes

Fixes #1578

## Summary by CodeRabbit

* **Bug Fixes**
  * When annotation-based settings are in use, cluster/CR-level write-back defaults are now merged with annotation-derived write-back settings before resolving the final configuration. This ensures consistent precedence and predictable write-back behavior across annotation and CR configuration paths.


dkarpele:
according to doc https://argocd-image-updater.readthedocs.io/en/stable/configuration/applications/#reading-configuration-from-application-annotations
```
When useAnnotations is set to true, all CR-based configuration is ignored (images including manifestTargets within each image, commonUpdateSettings, writeBackConfig)
```

wucm667:
@dkarpele Good catch on the doc. The issue (#1578) is actually about the CR-based path (useAnnotations=false) — when you set applicationRefs[].writeBackConfig with only writeBackTarget, the global method/repository/branch aren't inherited. You mentioned there that this "sounds like a bug and should be fixed."
My fix merges global writeBackConfig into the appRef-level one so partial overrides work. The question is: should this merge only apply when useAnnotations=false? If useAnnotations=true means "ignore everything CR-based," I can add a guard to skip the merge in that case.
Let me know which behavior you'd prefer and I'll adjust.

dkarpele:
@wucm667 #1578 was resolved by #1647.
What issue do you want to solve?

wucm667:
@dkarpele #1647 covers the CR-based path. This PR covers the annotation-based path (~line 301) which still skips mergeWBCSettings. If the intent is "annotations = standalone, inherit nothing" I'll close this.

wucm667:
Opened Apr 28, well before #1647. Covers the annotation-based path (~line 301) that #1647 didn't touch. Already explained above — if not wanted, feel free to close.

dkarpele:
Annotation based path is expected to be standalone and ignore all other settings.

PR #1633, opened 2026-05-05
fix: allow webhook traffic through NetworkPolicy (#1544)

### LINKED ISSUE #1544: Exposing Webhook does not work without further adjustments
**Describe the bug**
The documentation for exposing the ArgoCD Image Updater webhook does not mention that a NetworkPolicy is created by default which restricts ingress traffic to port 8443 and only from namespaces with the label `metrics=enabled`. This causes attempts to reach the webhook on the service port (e.g., 8080) to timeout, even when the Service and Ingress are correctly configured.

**To Reproduce**

1. Install ArgoCD Image Updater in a Kubernetes cluster.
2. Expose the webhook using the default Service.
3. Attempt to reach the webhook via the ClusterIP service on port 8080 (e.g., `curl http://<service-ip>:8080/webhook`) or via Ingress.
4. Observe that the request times out.

**Expected behavior**
The documentation should mention that a NetworkPolicy exists and that it may need to be modified to allow ingress on the webhook port (8080) from the desired sources. Users should be able to access the webhook service without unexpected timeouts.

**Additional context**
The NetworkPolicy that causes this is:

```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-metrics-traffic
  namespace: argocd
spec:
  podSelector:
    matchLabels:
      control-plane: argocd-image-updater-controller
  policyTypes:
    - Ingress
  ingress:
    - from:
        - namespaceSelector:
            matchLabels:
              metrics: enabled
      ports:
        - port: 8443
          protocol: TCP
```

This policy blocks access to port 8080, which is used by the webhook.

I'm not a pro in NetworkPolicies but after adjusting the rules it startet working. Without only Port-Forwarding works. Ingress does not.

ystkfujii:
## Summary

This PR addresses #1544 by documenting the NetworkPolicy requirement for the webhook server and adding a NetworkPolicy manifest that allows webhook traffic to the argocd-image-updater controller.

The original issue could probably be fixed with documentation only, because users can create their own NetworkPolicy when exposing the webhook server. However, the default installation already includes NetworkPolicy resources for the controller, so exposing the webhook server with the provided Service and Ingress manifests does not work unless traffic to the webhook port is explicitly allowed.

For that reason, I added a webhook-specific NetworkPolicy in addition to the documentation update.

If maintainers prefer to keep this as a docs-only change, I can remove the manifest changes and keep only the documentation update.

## Notes

The webhook Service exposes port `8080` and forwards traffic to the controller webhook port `8082`, so the NetworkPolicy allows traffic to port `8082`.

Closes: #1544

## Release Notes

* **New Features**
  * Added network policy configuration for secure webhook traffic handling with built-in port configuration and namespace-level access controls.

* **Documentation**
  * Updated webhook setup documentation with comprehensive instructions, default port information, and guidance for custom webhook port configurations.

### DIFF
--- a/config/install.yaml
+++ b/config/install.yaml
@@ -1168,3 +1168,26 @@ spec:
       control-plane: argocd-image-updater-controller
   policyTypes:
   - Ingress
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  labels:
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/name: argocd-image-updater-controller
+    control-plane: argocd-image-updater-controller
+  name: allow-webhook-traffic
+spec:
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          webhooks: enabled
+    ports:
+    - port: 8082
+      protocol: TCP
+  podSelector:
+    matchLabels:
+      control-plane: argocd-image-updater-controller
+  policyTypes:
+  - Ingress
--- a/config/network-policy/allow-webhook-traffic.yaml
+++ b/config/network-policy/allow-webhook-traffic.yaml
@@ -1,0 +1,26 @@
+# This NetworkPolicy allows ingress traffic to the webhook server
+# from Pods running in namespaces labeled with 'webhooks: enabled'. Only Pods in those
+# namespaces are able to send webhook notifications to the controller.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  labels:
+    control-plane: argocd-image-updater-controller
+    app.kubernetes.io/name: argocd-image-updater-controller
+    app.kubernetes.io/managed-by: kustomize
+  name: allow-webhook-traffic
+spec:
+  podSelector:
+    matchLabels:
+      control-plane: argocd-image-updater-controller
+  policyTypes:
+    - Ingress
+  ingress:
+    # This allows ingress traffic from any namespace with the label webhooks: enabled
+    - from:
+      - namespaceSelector:
+          matchLabels:
+            webhooks: enabled  # Only from namespaces with this label
+      ports:
+        - port: 8082
+          protocol: TCP
--- a/config/network-policy/kustomization.yaml
+++ b/config/network-policy/kustomization.yaml
@@ -1,2 +1,3 @@
 resources:
 - allow-metrics-traffic.yaml
+- allow-webhook-traffic.yaml
--- a/docs/configuration/webhook.md
+++ b/docs/configuration/webhook.md
@@ -274,11 +274,28 @@ not be able to use this method.
 
 ## Exposing the Server
 
-To expose the webhook server we have provided a service and ingress to get 
-started. These manifests are not applied with `install.yaml` so you will need 
-to apply them yourself. 
+To expose the webhook server we have provided a service and ingress to get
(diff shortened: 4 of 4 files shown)


config/install.yaml
akhilnittala, line 1148: we can make this two network policies into one right
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-metrics-and-webhook-traffic
  labels:
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/name: argocd-image-updater-controller
    control-plane: argocd-image-updater-controller
spec:
  podSelector:
    matchLabels:
      control-plane: argocd-image-updater-controller
  policyTypes:
  - Ingress
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          metrics: enabled
    ports:
    - port: 8443
      protocol: TCP
  - from:
    - namespaceSelector:
        matchLabels:
          webhooks: enabled
    ports:
    - port: 8082
      protocol: TCP

config/network-policy/kustomization.yaml
```diff
@@ -1,2 +1,3 @@
 resources:
 - allow-metrics-traffic.yaml
+- allow-webhook-traffic.yaml
```
akhilnittala, line 3: if we do below change, then no need of this.

config/network-policy/allow-webhook-traffic.yaml
```diff
@@ -0,0 +1,26 @@
+# This NetworkPolicy allows ingress traffic to the webhook server
+# from Pods running in namespaces labeled with 'webhooks: enabled'. Only Pods in those
+# namespaces are able to send webhook notifications to the controller.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  labels:
```
akhilnittala, line 7: this can also be undo if we do below changes

akhilnittala (changes requested):
and also need to do make manifests generate bundle command, missing those changes in bundle directory

config/network-policy/allow-webhook-traffic.yaml
akhilnittala: @ystkfujii , @chengfang  what do you say ?

ystkfujii: Thanks for the suggestion.

I intentionally kept the webhook NetworkPolicy separate from the existing metrics NetworkPolicy because they serve different purposes. I think keeping them separate makes the intent clearer and makes future changes easier.

I also think this is more convenient for users who customize the manifests. For example, with Kustomize overlays, they can target only the webhook policy by name when they want to patch, replace, or remove it. If both rules are combined into a single NetworkPolicy, users would need to patch a specific entry inside `spec.ingress`, which feels a bit more harder to read.

That said, combining them into a single NetworkPolicy would also work from a Kubernetes NetworkPolicy perspective. If you prefer that style for this project, I can update the PR accordingly.

chengfang: If the NetworkPolicy is centrally managed by an operator, it makes sense to combine them as one unit. But for Image Updater users who directly configure it themselves, I tend to agree with @ystkfujii to keep them separate.

chengfang (approved):

akhilnittala (approved):

PR #1644, opened 2026-05-13
feat: option to compare against one registry, pull from another

### LINKED ISSUE #1176: Compare against one registry, pull from another
I need to check one registry for new image versions but then pull from another.

Imagine your source of truth is ACR. But you also use ECR as a pull-through cache. Now ECR will only check upstream once a day for newer version (mutable tags). That doesn't work well when you run, for example, `latest` and expect it to actually always have the latest image.

I thought I could be clever and point image updater at ACR with digest strategy. It would see there's a newer digest and try to pull the image from ECR (which, in turn, would pull from ACR). There's a kyverno policy in place rewriting image specs from ACR to ECR.

But it doesn't seem to work. What I see in image updater's logs is some `dummy` tag that ain't defined anywhere and for some reason an empty tag for the image.
I suppose [this](https://github.com/argoproj-labs/argocd-image-updater/blob/889af4b9de51e46fc8197c9571f9197e0b6b1ac4/registry-scanner/pkg/image/image.go#L164) could be an issue.

I could hack together a custom script doing the same thing, but since we use image updater anyway this would've been a clever use.

**Additional context**
Here are some logs:

```
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:13Z" level=debug msg="Applications listed: 20"
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:13Z" level=info msg="Starting image update cycle, considering 1 annotated application(s) for update"
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:13Z" level=debug msg="Processing application argocd/dagster"
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:13Z" level=debug msg="Considering this image for update" alias=xyz application=dagster image_name=xyz image_tag= registry=xxx.azurecr.io
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:13Z" level=debug msg="Using version constraint 'latest' when looking for a new tag" alias=xyz application=dagster image_name=xyz image_tag= registry=xxx.azurecr.io
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:14Z" level=debug msg="ignoring v2 manifest sha256:e24de3ba64eabf1ea15d09871066aae58b5e083a6393826b91078ac479964ac5. Manifest platform: linux/amd64, requested: linux/arm64" alias=xyz application=dagster image_name=xyz image_tag= registry=xxx.azurecr.io
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:14Z" level=debug msg="No metadata found for xyz:latest" alias=xyz application=dagster image_name=xyz image_tag= registry=xxx.azurecr.io
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:14Z" level=debug msg="target parameters: image-spec= image-name=dagster-user-deployments.deployments[0[].image.repository, image-tag=dagster-user-deployments.deployments[0[].image.tag" application=dagster image=xxx.azurecr.io/xyz namespace=argocd
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:14Z" level=debug msg="Image 'xxx.azurecr.io/xyz@dummy' already on latest allowed version" alias=xyz application=dagster image_name=xyz image_tag=dummy registry=xxx.azurecr.io
argocd-image-updater-6b8f7bb96d-xwwml time="2025-07-01T13:21:14Z" level=info msg="Processing results: applications=1 images_considered=1 images_skipped=0 images_updated=0 errors=0"
```

gugu:
Resolves #1176

**Problem:**

Right now argogcd-image updater does not work well with image rewrites. There are few reasons to use image rewrites in production clusters:
* In general, public container images provide no SLA and container registry disruption (or deletion) can lead to production downtime. Good example is bitnami company, which renamed hundreds of their images. Best practice here is to use some caching proxy, for example I use AWS ECR Image Proxy
* Fetching images from outside of AWS costs $0.09/GB and can lead to slow start of the container, AWS ECR Image Proxy is helpful there
* In China connections to dockerhub/ghcr impossible, so we use Alibaba Registry there

It is really easy to setup image rewrites with Kyverno (example: https://blog.oponomarov.com/posts/rewriting-docker-image-registries-with-kyverno/, https://kyverno.io/policies/other/replace-image-registry/replace-image-registry/) and it works fine with ArgoCD, but image updater does not expect this.

**What happens in this case:**

1. for example, I have an image ghcr.io/example/nginx:1.2.3 and kyverno rewrites it to 1234567890.dkr.ecr.us-east-1.amazonaws.com/example-proxy/nginx:1.2.3
2. when ghcr.io/example/nginx:1.2.4 released argocd image updater does not find any usages of ghcr.io/example/nginx:1.2.4 in production cluster and finishes the job
3. if I point image-updater to 1234567890.dkr.ecr.us-east-1.amazonaws.com/example-proxy/nginx it will never pick up a new version because the latest version there will be always 1.2.3 unless I try to fetch 1.2.4
4. I can (and it is most reliable way now) enable forceUpdate=true, but with 20-30 monitored images it adds enormous load to github

This PR allows to avoid all these issues:

1. I can set matchName = 1234567890.dkr.ecr.us-east-1.amazonaws.com/example-proxy/nginx
2. image updater will find this image during the match step
3. it will pick up new version from ghcr.io/example/nginx

**Important**: this code was written by Claude Code. I reviewed it and tried to make it minimal and covered by tests, but there is a chance some important part was not taken into account

## Summary by CodeRabbit

* **New Features**
  * Optional Kustomize matchName lets you override which image identifier is used for live-image matching and write-back (handles mutated/rewritten registries).

* **CRD**
  * Schema and install manifests expose the new matchName field for Kustomize image targets.

* **Runtime**
  * Matching and update logic prefer matchName when set, with fallback to existing identifiers; write-back uses the resolved query image.

* **Tests**
  * Added unit and E2E coverage for matchName behavior, fallback, and write-back.


dkarpele:
Thanks for the contribution. 
1. Please run 
```
make generate
make manifests
make build-installer
```

2. Please add an e2e test to `test/ginkgo/parallel`

gugu:
I'll double check the code first, then I'll reopen this PR

image

gugu:
1. done, except changing version to v99.9.9, which it generated
2. done

gugu:
I'll check why test fails on CI

gugu:
ok, it was me breaking the integration test after it was working initially, now it passes

gugu:
should I address coderabbit comments? I'm not sure they need to be addressed

dkarpele:
> should I address coderabbit comments? I'm not sure they need to be addressed

Consider coderabbit as any other reviewer. If you agree with it's comment - address them, if not - reply in the thread that you don't want to fix this.

dkarpele:
I have a concern about the design pattern. `matchName` solves a problem (Kyverno/mutating webhook rewrites) that is not Kustomize-specific — it affects Helm apps equally. So, I think `matchName` should live in `ImageConfig`.

chengfang:
Just another approach to consider, how about introduce some mapping in registries.conf between teh canonical registry url and its proxy or mirrors? Image updater can look up and replace and redirect to the proxy/mirror at various places. This way no change is needed to ImageUpdater CRD, and the above config in registries.conf will apply to all apps.

PR #1664, opened 2026-05-31
fix(git): ignore injected GIT_CONFIG_* for git subprocesses

immanuwell:
This fixes a pretty annoying env leak in `ext/git`.

If `GIT_CONFIG_PARAMETERS` is exported, git subprocesses can inherit `commit.gpgsign=true` and `user.signingkey`. Then `ext/git` tests can blow up on a normal dev machine, because the git client also forces `HOME=/dev/null`.

Repro:
1. `export GIT_CONFIG_PARAMETERS="'commit.gpgsign=true' 'user.signingkey=F7BC234BE02886EE'"`
2. `go test ./ext/git -run 'TestCommit/(SignOff|SSHSigning)' -count=1`

On current `master` this fails with `gpg: Fatal: can't create directory '/dev/null/.gnupg'`.

Fix is simple:
1. drop injected `GIT_CONFIG_*` vars before running git subprocesses
2. do the same in the test helpers
3. add a regression test for `GIT_CONFIG_PARAMETERS`

Checked with `make test`. Also re-ran the same repro on upstream `master` and on this branch. `master` fails, this branch passes. small patch, but it plugs the leak nicely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Git subprocesses now execute with a filtered set of environment variables, excluding configuration-related entries.

### DIFF
--- a/ext/git/env.go
+++ b/ext/git/env.go
@@ -1,0 +1,30 @@
+package git
+
+import "strings"
+
+// sanitizedGitEnv removes process-level Git config injection that would make
+// git subprocess behavior depend on the caller environment instead of the repo
+// or explicit command options.
+func sanitizedGitEnv(env []string) []string {
+	filtered := make([]string, 0, len(env))
+	for _, kv := range env {
+		key := kv
+		if idx := strings.IndexByte(kv, '='); idx >= 0 {
+			key = kv[:idx]
+		}
+		if shouldDropGitEnv(key) {
+			continue
+		}
+		filtered = append(filtered, kv)
+	}
+	return filtered
+}
+
+func shouldDropGitEnv(key string) bool {
+	switch key {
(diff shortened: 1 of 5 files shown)


ext/git/env.go
```diff
@@ -0,0 +1,30 @@
...
+			continue
+		}
+		filtered = append(filtered, kv)
+	}
+	return filtered
+}
+
+func shouldDropGitEnv(key string) bool {
+	switch key {
+	case "GIT_CONFIG", "GIT_CONFIG_COUNT", "GIT_CONFIG_GLOBAL", "GIT_CONFIG_NOSYSTEM", "GIT_CONFIG_PARAMETERS", "GIT_CONFIG_SYSTEM":
```
chengfang, line 25: how about not dropping GIT_CONFIG_SYSTEM and GIT_CONFIG_NOSYSTEM?

GIT_CONFIG_NOSYSTEM suppresses using the system wide config such as /etc/gitconfig, which can be a desirable protection.

GIT_CONFIG_SYSTEM: some administrators may want to configure this as a custom git config.

PR #1676, opened 2026-06-09
ci: run CI tests on push to master

AdamSaleh:
## Summary
- Adds `push` trigger for the `master` branch to the CI tests workflow
- Without this, the Codecov upload with the `unit-tests` flag only runs on PRs and never against the default branch, leaving the flag at 0% on Codecov
- This ensures coverage data is uploaded on every merge to master, keeping Codecov's flag-filtered views up to date

## Test plan
- [ ] Verify CI workflow runs on the next merge to master
- [ ] Confirm the `unit-tests` flag on Codecov shows non-zero coverage after the first push-triggered run

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Enhanced continuous integration by adding support for running automated checks on direct pushes to the main branch, expanding coverage beyond pull requests while ensuring selected jobs only run under the intended event types.

### DIFF
--- a/.github/workflows/ci-tests.yaml
+++ b/.github/workflows/ci-tests.yaml
@@ -1,5 +1,8 @@
 name: Integration tests
 on:
+  push:
+    branches:
+    - 'master'
   pull_request:
     branches:
     - 'master'
@@ -22,8 +25,8 @@ permissions:
 jobs:
   check-go:
     name: Ensure Go modules synchronicity
-    # Skip on label events — this job is already covered by the pull_request run.
-    if: github.event_name != 'pull_request_target'
+    # Skip on label events and push-to-master (already passed pre-merge).
+    if: github.event_name != 'pull_request_target' && github.event_name != 'push'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


dkarpele (approved):

.github/workflows/ci-tests.yaml
chengfang, line 1: With push trigger, every PR merge will trigger this workflow (except e2e test) to run again, since merge is a push event to master. This seems waste since before merge, all CI tests already passed.

Can we run only codecov steps for push event?

AdamSaleh: As far as I understand, this is to make sure master also gets data populated which we then use in the metrics dashboards. There might be a better/more cost effective way to do this, but I would rather have this less efficient think that works now and improve on it later :)

chengfang:
the impact is large, all ci tests except e2e tests will run again for no gain, since they already passed pre-merge. We can either completely move codecov step to another workflow file that is triggered by push event, or add conditions to restrict this push even only triggers codecov step. WDYT?

.github/workflows/ci-tests.yaml
AdamSaleh: Good point — added `&& github.event_name != 'push'` to the check-go, codegen, and lint jobs so they skip on push-to-master. Only the test and registry-scanner jobs run on push (to generate coverage for Codecov).

AdamSaleh:
@chengfang ok, added skips so that only unit-tests that generate the codecov are run. Is that  enough of a reduction, or do you need something more?

chengfang (approved):

PR #1678, opened 2026-06-09
feat(webhook): enforce mandatory webhook secret configuration

dkarpele:
- Add a startup validation check that refuses to start the webhook server (or refuses to register a handler) when its corresponding secret is empty or unset. Log a clear error message identifying which handler(s) lack secrets.

- Replace the != string comparison in docker.go:44 and quay.go:40 with crypto/subtle.ConstantTimeCompare, matching the pattern already used by GHCR, Harbor, and other handlers.

- Add a global --require-webhook-secrets flag (default true) that enforces secret presence for all registered handlersat startup.

> Migrate Docker Hub and Quay handlers from query-parameter-based secret transmission to header-based validation (e.g., HMAC signature in a custom header), or clearly document that query parameter secrets are deprecated and will be removed.

This is already documented at `docs/configuration/webhook.md`, see `### Parameter Secrets`

## Summary by CodeRabbit

* **New Features**
  * Added --webhook-require-secret flag (env WEBHOOK_REQUIRE_SECRET, default: true) to control which webhook endpoints are registered.
  * Improved secret validation to use constant-time comparisons for relevant webhook providers.

* **Bug Fixes**
  * Returns a clear error when no webhook handlers are registered.

* **Documentation**
  * Updated webhook and run command docs with flag details and security guidance.

* **Tests**
  * Added tests covering handler registration behavior and related cases.

### DIFF
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -11,6 +11,8 @@ import (
 	"github.com/go-logr/logr"
 	"go.uber.org/ratelimit"
 
+	"github.com/argoproj-labs/argocd-image-updater/registry-scanner/pkg/log"
+
 	"github.com/argoproj-labs/argocd-image-updater/internal/controller"
 	"github.com/argoproj-labs/argocd-image-updater/pkg/argocd"
 	"github.com/argoproj-labs/argocd-image-updater/pkg/common"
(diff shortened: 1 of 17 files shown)


chengfang:
about the naming of this new flag, --webhook-require-secrets and its env var: all existing secret-related flags and env vars are singluar ending with `-secret` instead of `-secrets`. I think it will be consistent and a bit mroe usable to rename it `--webhook-require-secret`

chengfang (approved):

PR #1683, opened 2026-06-15
fix(webhook-quay): Self-hosted Quay images redirect to quay.io, not self-hosted URL

chengfang:
## Summary by CodeRabbit

* **Bug Fixes**
  * Quay webhooks now derive registry URL from the docker_url field when present, falling back to quay.io if unavailable.

* **Tests**
  * Enhanced test coverage for Quay webhook registry URL parsing, including scenarios with various hostname formats and fallback cases.

### DIFF
--- a/pkg/webhook/quay.go
+++ b/pkg/webhook/quay.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
+	"strings"
 
 	"github.com/argoproj-labs/argocd-image-updater/pkg/argocd"
 )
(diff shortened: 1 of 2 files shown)


dkarpele:
Probably this is not for current PR but future follow-up.
`RegistryURL` is also hardcoded in `docker.go` and `ghcr.go`. I can expect similar changes for these handlers. Handlers like harbor or aliyn parse registry url differently.
Is it possible to make this field customazible? Otherwise we should fix this for docker and ghcr.

chengfang:
> Probably this is not for current PR but future follow-up. `RegistryURL` is also hardcoded in `docker.go` and `ghcr.go`. I can expect similar changes for these handlers. Handlers like harbor or aliyn parse registry url differently. Is it possible to make this field customazible? Otherwise we should fix this for docker and ghcr.

Yes, we should fix other handlers too, making them consistent with harbor and quay, in the future. This PR is strictly bug fix and will back port to previous release branches, so we want to keep the scope and risk low.

dkarpele (approved):

PR #1706, opened 2026-06-30
feat: implement image signature verification

dkarpele:
## Summary by CodeRabbit

* **New Features**
  * Added configurable image signature verification (cosign) with precedence across global, application reference, and per-image scopes.
  * Added support for specifying the cosign public key via a Kubernetes Secret reference.
  * Extended CRD schema and shipped updated samples demonstrating verification enablement and per-image opt-out.
* **Bug Fixes**
  * Verification failures now skip only the affected image(s) while allowing other images to continue updating.
* **Documentation**
  * Added an “Image Signature Verification” guide and updated application/CRD configuration examples.
* **Tests**
  * Expanded unit and registry verification test coverage for signature verification behavior and edge cases.


pkg/argocd/argocd.go
```diff
@@ -318,10 +327,22 @@ func FilterApplicationsForUpdate(ctx context.Context, ctrlClient *ArgoCDK8sClien
 					localAppRef.Images = appRefImages
 					localAppRef.WriteBackConfig = appRefWBC
 					mergedWBCSettings = appRefWBC
+
+					if verifyImage {
+						mergedImagesVerification = applicationRef.ImagesVerification
```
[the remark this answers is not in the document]
dkarpele: This is expected. See comment above:
```
				// When UseAnnotations is true, we ignore all CR-based configuration
				// (Images, CommonUpdateSettings, WriteBackConfig) and instead read everything from
				// the Application's legacy argocd-image-updater.argoproj.io/* annotations.
				// Only NamePattern, LabelSelectors and ImagesVerification are considered at app level.
```

docs/configuration/images.md
```diff
@@ -354,6 +354,124 @@ images:
...
+  imagesVerification:
+    method: cosign-key
+    publicKeySecret:
+      secretName: org-cosign-pubkey
+      key: cosign.pub
+  applicationRefs:
+    - namePattern: "my-app"
+      images:
+        - alias: "myapp"
+          imageName: "myorg/myapp:~1.0"
```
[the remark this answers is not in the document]
dkarpele: Hm, I'm using tilda ~ in my configs and they work fine

config/manager/manager.yaml
```diff
@@ -222,6 +222,12 @@ spec:
                 name: argocd-image-updater-config
                 key: tls.ciphers
                 optional: true
+          - name: VERIFY_IMAGES
+            valueFrom:
+              configMapKeyRef:
+                name: argocd-image-updater-config
+                key: verify.images
+                optional: true
```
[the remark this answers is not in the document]
dkarpele: flag removed

api/v1alpha1/imageupdater_types.go
```diff
@@ -277,6 +297,38 @@ type KustomizeTarget struct {
...
+	// +optional
+	// +kubebuilder:default:=true
+	Enable *bool `json:"enable,omitempty"`
+
+	// PublicKeySecret references a Kubernetes Secret in the same namespace as the
+	// ImageUpdater CR that holds the PEM-encoded ECDSA public key used to verify
+	// cosign signatures.
+	// Required when method is "cosign-key" and enable is true.
+	// +optional
+	PublicKeySecret *SecretRef `json:"publicKeySecret,omitempty"`
```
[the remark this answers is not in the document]
dkarpele: fixed

pkg/argocd/argocd.go
```diff
@@ -141,7 +141,7 @@ func nameMatchesLabels(ctx context.Context, appLabels map[string]string, selecto
 
 // processApplicationForUpdate checks if an application is of a supported type,
 // and if so, creates an ApplicationImages struct and adds it to the update map.
-func processApplicationForUpdate(ctx context.Context, app *argocdapi.Application, appRef iuapi.ApplicationRef, appCommonUpdateSettings *iuapi.CommonUpdateSettings, appWBCSettings *WriteBackConfig, appNSName string, appsForUpdate map[string]ApplicationImages, webhookEvent *WebhookEvent) {
+func processApplicationForUpdate(kubeClient *kube.ImageUpdaterKubernetesClient, ctx context.Context, app *argocdapi.Application, appRef iuapi.ApplicationRef, appCommonUpdateSettings *iuapi.CommonUpdateSettings, appImagesVerification *iuapi.ImagesVerification, appWBCSettings *WriteBackConfig, appNSName string, appsForUpdate map[string]ApplicationImages, webhookEvent *WebhookEvent) {
```
chengfang: can we keep ctx as the first parameter, just to be consistent with existing convention?

docs/configuration/images.md
```diff
@@ -582,6 +703,28 @@ update strategies and set options for images.
...
+    verification. Set `enable: false` to explicitly opt out of verification for a specific image.
+
+#### SecretRef fields
+
+| Field        | Type   | Required | Description                                                                          |
+|--------------|--------|----------|--------------------------------------------------------------------------------------|
+| `secretName` | string | Yes      | Name of the Kubernetes Secret in the same namespace as the ImageUpdater CR           |
+| `key`        | string | Yes      | Key within the Secret's `data` map whose value is the credential (e.g. `cosign.pub`) |
+
 #### WriteBackConfig fields
```
chengfang, line 725: can we also document the case when `useAnnotation` is true, there is no global level or image level verification configuration?

api/v1alpha1/imageupdater_types.go
```diff
@@ -277,6 +297,42 @@ type KustomizeTarget struct {
...
+	// +kubebuilder:validation:Enum=cosign-key
+	// +optional
+	Method *string `json:"method,omitempty"`
+
+	// Enable controls whether signature verification is active at this scope.
+	// Defaults to true when the ImagesVerification block is present.
+	// Set to false to explicitly opt out of verification for this image or group.
+	// +optional
+	// +kubebuilder:default:=true
+	Enable *bool `json:"enable,omitempty"`
```
chengfang: does `enabled` sound better? I think in ArgoCD CR we use `enabled` to enable ImageUpdater.

docs/configuration/images.md
dkarpele: Yep, I remembered about this already at night. I updated `applications.md`

docs/configuration/applications.md
```diff
@@ -203,14 +203,20 @@ spec:
       labelSelectors:
         matchLabels:
           image-updater: my-image-updater
+      imagesVerification:
+        method: cosign-key
+        publicKeySecret:
+          secretName: org-cosign-pubkey
+          key: cosign.pub
```
chengfang, line 209: how about removing `method` and add a `cosignKey` element to contain all data needed for cosign-key?

In the future when we implement keyless, we can then add another `cosignKeyless` element to contain its own set of necessary fields?

Since for different methods, their required data fields will be different, so it's better to keep them separate.

something like:
```yaml
 imagesVerification:
   enabled: true
   cosignKey:                       # mutually exclusive with cosignKeyless
     secretName: my-cosign-keys
     key: cosign.pub
   cosignKeyless:
     xxx:
     yyy:
   anotherSigningMethod:     
```

dkarpele: sounds reasonable. It simplifies API though requires some rework. WIll have a look

dkarpele: done

chengfang (approved):

PR #1707, opened 2026-07-01
feat(argocd): support Plugin source type with git write-back

### LINKED ISSUE #168: Image Update support for "Plugin" ArgoCD Applications.
ArgoCD supports a concept of Plugins, such as the [kustomize/helm](https://github.com/argoproj/argocd-example-apps/tree/master/plugins/kustomized-helm) integration, and also used for extending ArgoCD for other use cases.

It appears that the argocd-image-updater only functions with the app.Status.SourceType is set to `Kustomize` or `Helm` (via auto-detect), and not when it is set to `Plugin`.

When the type Plugin is used, we could still implement Helm or Kustomize like-behavior for git-based workflows, whereas the image override file can be written back to git as if it were one of these two types.

Mainly it would be dropping the use of the plugin, but it has proven to be handy at times for pre-Sync instrumentation of service discovery.

**Additional context**
Relevant segment using the kustomized helm example:
```
resources:
  - health:
      status: Healthy
    kind: Service
    name: kustomize-release-name-helm-guestbook
    namespace: default
    status: Synced
    version: v1
  - group: apps
    health:
      status: Healthy
    kind: Deployment
    name: kustomize-release-name-helm-guestbook
    namespace: default
    status: Synced
    version: v1
  sourceType: Plugin
  summary:
    images:
    - gcr.io/heptio-images/ks-guestbook-demo:0.2
```

I'm not 100% sure if this is an ArgoCD thing, argocd-image-updater, or sort of both for a feature such as this.  I understand there's a lot of "if" there especially if it couldn't reconcile using the imperative updated to the Application resource itself.

flobz:
Closes #168

## Summary

ArgoCD Image Updater currently skips any Application whose source type is `Plugin` with:

```
skipping app '...' of type 'Plugin' because it's not of supported source type
```

This makes it incompatible with helmfile, cdk8s, argocd-vault-plugin, and any other custom plugin source — regardless of write-back method.

## Root cause

`getApplicationType()` in `pkg/argocd/argocd.go` only recognises `Helm` and `Kustomize`. Plugin falls through to `ApplicationTypeUnsupported`, which causes `IsValidApplicationType()` to return false and skip the app entirely.

## Fix

When source type is `Plugin` **and** the write-back method is `git`, return `ApplicationTypeHelm`. For git write-back, the write-back target format (helmvalues file, kustomization file, or default Helm parameter overrides) determines how the image tag is serialised — the source type is irrelevant to that operation.

The `argocd` write-back method for Plugin apps remains unsupported: ArgoCD has no mechanism to override plugin parameters directly from the Application spec.

**Note:** The `kustomization` write-back target is already handled transparently. When `writeBackTarget: "kustomization"` is set, `parseKustomizeBase()` populates `wbc.KustomizeBase`, which causes `getApplicationSourceType()` to return `ApplicationSourceTypeKustomize` before `getApplicationType()` ever reaches the Plugin branch. So Plugin + kustomization write-back target was already working; this patch covers Plugin + helmvalues and Plugin + default Helm overrides.

## Changes

| File | Change |
|------|--------|
| `pkg/argocd/argocd.go` | Add Plugin branch in `getApplicationType()` (5 lines) |
| `pkg/argocd/argocd_test.go` | Three new sub-tests in `Test_GetApplicationType` (Plugin + git → Helm, Plugin + kustomization target → Kustomize, Plugin + argocd → Unsupported) |
| `pkg/argocd/update_test.go` | New sub-test in `Test_MarshalParamsOverride` — Plugin app + git write-back → Helm YAML output |
| `docs/configuration/applications.md` | New "Plugin source type" section under write-back documentation |

## Testing

Tests follow TDD: failing tests committed before the fix, fix committed after.

```
go test ./pkg/argocd/... -count=1
ok   github.com/argoproj-labs/argocd-image-updater/pkg/argocd
```

## Motivation

The driving use case is a helmfile-based ArgoCD application: helmfile is configured as a Config Management Plugin, and the image tag is passed through a `values.yaml` file tracked in git. When a new image is pushed to the registry, Image Updater should write the new tag back to that values file so the next ArgoCD sync picks it up automatically.

This pattern — helmfile CMP + git write-back to a chart values file — is currently blocked entirely because Image Updater rejects any Plugin source app before even attempting a write-back. This patch removes that restriction for git write-back, where the target file format (not the source type) is what determines how the tag is serialised.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for plugin-based apps, including the required Git write-back behavior, supported `writeBackTarget` outputs, and updated examples.
* **Bug Fixes**
  * Corrected plugin app handling so Git write-back classifies plugin sources appropriately and generates the expected Helm-formatted output.
* **Tests**
  * Expanded coverage for plugin scenarios to verify application type selection and the generated parameter override YAML format.

### DIFF
--- a/docs/configuration/applications.md
+++ b/docs/configuration/applications.md
@@ -351,6 +351,46 @@ Where `<method>` must be one of `argocd` (imperative) or `git` (declarative).
 
 The default used by Argo CD Image Updater is `argocd`.
 
+### Plugin source type
+
+Applications using a custom plugin source type are supported with the **git**
+write-back method, provided the plugin consumes Helm-style or Kustomize-style
+values files from git (e.g. helmfile). The `argocd` write-back method is
+**not** supported for Plugin apps because ArgoCD has no mechanism to override
+plugin environment variables from the Application spec.
+
+Configure the write-back target explicitly to control the file format:
+
+| Target | Format written to Git |
+|--------|-----------------------|
+| `kustomization` or `kustomization:<path>` | `kustomization.yaml` (Kustomize image override) |
+| `helmvalues:<path>` | Helm values YAML |
+| *(none / default)* | `.argocd-source-<app>.yaml` (ArgoCD Helm parameter overrides — Plugin apps are classified as Helm) |
+
+**Helmfile example** using `kustomization` write-back:
+
+```yaml
+spec:
+  writeBackConfig:
+    method: "git"
+    gitConfig:
+      repository: "https://github.com/myorg/infra.git"
+      branch: "main"
+  applicationRefs:
+    - namePattern: "my-helmfile-app.*"
+      writeBackConfig:
+        gitConfig:
+          writeBackTarget: "kustomization"
+      images:
+        - alias: img
+          imageName: "myregistry/myapp"
+          commonUpdateSettings:
+            updateStrategy: semver
+```
+
+The helmfile then reads the image tag from the generated `kustomization.yaml`
+via `readFile "kustomization.yaml" | fromYaml`.
+
 ## <a name="complete-example"></a>Complete example
 
 Here's a complete example that demonstrates various configuration options:
--- a/pkg/argocd/argocd.go
+++ b/pkg/argocd/argocd.go
@@ -1103,6 +1103,11 @@ func getApplicationType(app *argocdapi.Application, wbc *WriteBackConfig) Applic
 		return ApplicationTypeKustomize
 	} else if sourceType == argocdapi.ApplicationSourceTypeHelm {
 		return ApplicationTypeHelm
+	} else if sourceType == argocdapi.ApplicationSourceTypePlugin &&
+		wbc != nil && wbc.Method == WriteBackGit {
+		// Plugin apps: the write-back target format determines serialization, not the source type.
+		// KustomizeBase being non-empty is already caught above via getApplicationSourceType.
+		return ApplicationTypeHelm
 	} else {
 		return ApplicationTypeUnsupported
 	}
--- a/pkg/argocd/argocd_test.go
+++ b/pkg/argocd/argocd_test.go
@@ -359,6 +359,7 @@ func Test_GetApplicationType(t *testing.T) {
 	})
 
 	t.Run("Plugin app with argocd write-back remains unsupported", func(t *testing.T) {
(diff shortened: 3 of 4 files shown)


docs/configuration/applications.md
```diff
@@ -346,6 +346,45 @@ Where `<method>` must be one of `argocd` (imperative) or `git` (declarative).
...
+is **not** supported for Plugin apps because ArgoCD has no mechanism to
+override plugin environment variables from the Application spec.
+
+Configure the write-back target explicitly to control the file format:
+
+| Target | Format written to Git |
+|--------|-----------------------|
+| `kustomization` or `kustomization:<path>` | `kustomization.yaml` (Kustomize image override) |
+| `helmvalues:<path>` | Helm values YAML |
+| *(none / default)* | `.argocd-source-<app>.yaml` (ArgoCD Helm parameter overrides) |
```
chengfang: the default write-back target file works for both helm and kustomize overrides.

docs/configuration/applications.md
```diff
@@ -346,6 +346,45 @@ Where `<method>` must be one of `argocd` (imperative) or `git` (declarative).
 
 The default used by Argo CD Image Updater is `argocd`.
 
+### Plugin source type
+
+Applications using a custom plugin source type (e.g. helmfile, cdk8s) are
+supported with the **git** write-back method. The `argocd` write-back method
+is **not** supported for Plugin apps because ArgoCD has no mechanism to
+override plugin environment variables from the Application spec.
```
chengfang: The above gives the impression that both helmfile and cdk8s are supported, but I'm not so sure about cdk8s. Can we qualify it a bit to say something like plugins that consume Helm-style or Kustomize-style values files from git?

chengfang:
Can you also fix the DCO check failure?

docs/configuration/applications.md
flobz: I updated the docs tell me if it's ok.

docs/configuration/applications.md
flobz: I updated the docs tell me if it's ok.

chengfang (approved):
Thanks for the contribution. This will enable many to use helm- or kustomized-based plugins. We also have plan to further extend it to support general plugin in the near future.

PR #1733, opened 2026-07-17
fix: reject cross-namespace secret references in commonUpdateSettings.pullSecret

dkarpele:
I don't introduce simplified format (`pullsecret:<secret>`) - without namespace, because it's required complicated changes in many functions, including registry-scanner. And the value of this enhancement is almost 0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Image updates now reject image pull credential references that point to secrets in a different namespace.
  * Affected images are skipped and reported as errors instead of being updated incorrectly.

* **Documentation**
  * Added guidance clarifying that image pull secrets must be located in the same namespace as the associated image updater resource.

### DIFF
--- a/docs/configuration/images.md
+++ b/docs/configuration/images.md
@@ -281,6 +281,13 @@ Valid values for `secret_ref` are:
   secret, that is, it must have a valid Docker config in JSON format in the
   field `.dockerconfigjson`.
 
+!!!warning "Cross-namespace secret references are rejected"
+    For `secret:` and `pullsecret:` references, the `namespace` in the
+    reference **must** match the namespace of the `ImageUpdater` CR. Referencing
+    a secret from a different namespace (e.g. specifying `other-team/my-secret`
+    while the CR runs in `my-team`) is rejected with an error and the image is
+    skipped. This prevents tenants from reading secrets they do not own.
+
 * `env:<variable_name>` - Use credentials supplied by the environment variable
   named `variable_name`. This can be a variable that is i.e. bound from a
   secret within your pod spec.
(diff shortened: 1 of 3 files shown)


chengfang (approved):

YoanSimco:
Would it be possible to add an option to allow access to cross-namespace pullSecrets (like before https://github.com/argoproj-labs/argocd-image-updater/releases/tag/v1.3.0)?
Or rollback this PR?

My use case:

- ArgoCD applications located in `argocd` namespace
- So ArgoCD Image Updated CRs located in `argocd` namespace as well (requirement from image updater)
- Deployment and imagePullSecret located in `home-lab` namespace
- I don't want to duplicate the pull secret in both `home-lab` and `argocd` namespaces

Moreover, I don't get the benefit from this PR as this is a subject of RBAC here, the verification should not be done by the image updater controller itself, the security is already done by Kubernetes RBAC.
By default, the `argocd-image-updater-controller` service account is already least privilege and **do not have permissions for cross-account namespaces secrets**: https://github.com/argoproj-labs/argocd-image-updater/blob/v1.3.0/config/rbac/kustomization.yaml.

YoanSimco:
> Would it be possible to add an option to allow access to cross-namespace pullSecrets (like before https://github.com/argoproj-labs/argocd-image-updater/releases/tag/v1.3.0)? Or rollback this PR?
> 
> My use case:
> 
>     * ArgoCD applications located in `argocd` namespace
> 
>     * So ArgoCD Image Updated CRs located in `argocd` namespace as well (requirement from image updater)
> 
>     * Deployment and imagePullSecret located in `home-lab` namespace
> 
>     * I don't want to duplicate the pull secret in both `home-lab` and `argocd` namespaces
> 
> 
> Moreover, I don't get the benefit from this PR as this is a subject of RBAC here, the verification should not be done by the image updater controller itself, the security is already done by Kubernetes RBAC. By default, the `argocd-image-updater-controller` service account is already least privilege and **do not have permissions for cross-account namespaces secrets**: https://github.com/argoproj-labs/argocd-image-updater/blob/v1.3.0/config/rbac/kustomization.yaml.

Hi @dkarpele @chengfang,

Any news about my previous message?

PR #1743, opened 2026-07-31
feat: support custom CA certificates for registries

### LINKED ISSUE #1739: Support custom CA certificates for registry connections
## Summary

Image Updater does not currently support configuring custom CA certificates for registry connections. This makes it impossible to use Image Updater with registries that use certificates signed by an internal or private CA.

## Problem

When connecting to container registries, the `registry-scanner` creates a fresh `tls.Config{}` without loading the system certificate pool (`x509.SystemCertPool()`). This means:

1. Certificates from the system trust store (e.g., `/etc/pki/ca-trust/`) are not used for registry connections.
2. There is no configuration option to specify a CA bundle on a per-registry basis.
3. The only current workaround is marking the registry as `insecure`, which disables TLS verification entirely.

**Relevant code:** `registry-scanner/pkg/registry/endpoints.go` — `GetTransport()` initializes `tls.Config{}` with no root CAs.

## Proposed Solution

Two complementary changes:

1. **Use `x509.SystemCertPool()` as the default root CA pool** in `GetTransport()`, so that any CA certificates added to the container's system trust store are automatically respected for registry connections.

2. **Add a `caFile` (or `caBundle`) field to the registry configuration**, allowing users to specify a path to a PEM-encoded CA certificate file for a specific registry endpoint.

## Use Case

Organizations running private container registries with certificates signed by an internal CA (common in air-gapped or enterprise environments) cannot use Image Updater without disabling TLS verification. This feature would allow secure connections to such registries.

w3lld1:
## Summary

- I initialize registry transports with the system certificate pool.
- I added an optional per-registry `ca_file` setting that appends PEM certificates to the system trust store.
- I validate CA files while loading registry configuration and documented the new setting.
- I added coverage for configuration parsing, invalid bundles, system roots, and a successful TLS connection using a custom CA.

## Validation

- `cd registry-scanner && GOMAXPROCS=2 go test -p 2 ./pkg/registry -count=1`
- `cd registry-scanner && GOMAXPROCS=2 go vet ./pkg/registry`
- `gofmt -l` on the changed Go files
- `git diff --check origin/master...HEAD`

I did not run the full repository suite locally because it is disproportionate for this focused registry-scanner change; CI can cover the broader modules.

Fixes #1739

## Summary by CodeRabbit

* **New Features**
  * Added support for custom CA certificate files and inline certificates when connecting to registries.
  * Secure connections now use configured certificates, automatically discovered Argo CD certificates, or system trust stores.
  * Insecure registries skip CA certificate processing.

* **Bug Fixes**
  * Improved validation and error reporting for invalid or unavailable certificates.
  * Isolated connections with different certificate settings.

* **Documentation**
  * Added configuration guidance and updated the Red Hat Quay example to use a CA certificate file.

### DIFF
--- a/registry-scanner/pkg/registry/config.go
+++ b/registry-scanner/pkg/registry/config.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"context"
+	"crypto/x509"
 	"fmt"
 	"os"
 	"time"
@@ -22,6 +23,7 @@ type RegistryConfiguration struct {
 	TagSortMode string        `yaml:"tagsortmode,omitempty"`
 	Prefix      string        `yaml:"prefix,omitempty"`
 	Insecure    bool          `yaml:"insecure,omitempty"`
+	CAFile      string        `yaml:"ca_file,omitempty"`
 	DefaultNS   string        `yaml:"defaultns,omitempty"`
 	Limit       int           `yaml:"limit,omitempty"`
 	IsDefault   bool          `yaml:"default,omitempty"`
@@ -30,6 +32,36 @@ type RegistryConfiguration struct {
 // RegistryList contains multiple RegistryConfiguration items
 type RegistryList struct {
 	Items []RegistryConfiguration `yaml:"registries"`
+}
+
+func newRegistryEndpointFromConfig(config RegistryConfiguration) (*RegistryEndpoint, error) {
+	endpoint := NewRegistryEndpoint(config.Prefix, config.Name, config.ApiURL, config.Credentials, config.DefaultNS, config.Insecure, TagListSortFromString(config.TagSortMode), config.Limit, config.CredsExpire)
+	rootCAs, err := loadRootCAs(config.CAFile)
+	if err != nil {
+		return nil, fmt.Errorf("could not configure CA certificates for registry %s: %w", config.Name, err)
+	}
+	endpoint.CAFile = config.CAFile
+	endpoint.rootCAs = rootCAs
+	return endpoint, nil
+}
+
+func loadRootCAs(caFile string) (*x509.CertPool, error) {
+	rootCAs, err := x509.SystemCertPool()
+	if err != nil {
+		return nil, fmt.Errorf("could not load system certificate pool: %w", err)
+	}
+	if caFile == "" {
+		return rootCAs, nil
+	}
+
+	certificates, err := os.ReadFile(caFile)
+	if err != nil {
+		return nil, fmt.Errorf("could not read CA file %s: %w", caFile, err)
+	}
+	if !rootCAs.AppendCertsFromPEM(certificates) {
+		return nil, fmt.Errorf("CA file %s does not contain a valid PEM certificate", caFile)
+	}
+	return rootCAs, nil
 }
 
 func clearRegistries() {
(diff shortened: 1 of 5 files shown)


chengfang:
Thanks for the contribution. Can we add some auto-discovery capability like checking /app/config/tls/<hostname> from argocd-tls-certs-cm? argocd argocd-tls-certs-cm configmap (https://argo-cd.readthedocs.io/en/stable/operator-manual/argocd-tls-certs-cm-yaml/) is a typical place users keep ca data, and so we want to reuse any ca data already managed by argocd.

In OpenShift GitOps, argocd-tls-certs-cm is mounted to `/app/config/tls`, and the ca data for `host1` will be availabel at `/app/config/tls/host1`, without any user configuration, besides adding ca data for host1 to argocd-tls-certs-cm.

Note the above auto-discovery will not work in ImageUpdater standalone out of the box, since the path /app/config/tls is already taken by webhook tls cert and key. Users will need to mount argocd-tls-certs-cm to different path and, with your impl, configure ca_file accordingly.

I think we can also introduce another field like `ca_data` to embed PEM ca data directly in registries.conf field. Users then don't have to configure volume and mount, if they do not want mix their registry ca with other argocd ca.

chengfang:
when insecure is set to true in registries.conf, we should skip these tls-related processing. https://argocd-image-updater.readthedocs.io/en/release-1.2/configuration/registries/

registry-scanner/pkg/registry/endpoints.go
```diff
@@ -349,7 +357,15 @@ func (ep *RegistryEndpoint) GetTransport(ctx context.Context) *http.Transport {
...
 	// Create a new transport with optimized connection pool settings
-	tlsC := &tls.Config{}
+	rootCAs := ep.rootCAs
+	if rootCAs == nil {
+		var err error
+		rootCAs, err = x509.SystemCertPool()
+		if err != nil {
+			logCtx.Warnf("Could not load system certificate pool for %s: %v", ep.RegistryAPI, err)
+		}
+	}
```
chengfang, line 372: when rootCAs is nil, we should just use &tls.Config{} to let the go default kick in.

registry-scanner/pkg/registry/config.go
```diff
@@ -32,6 +34,36 @@ type RegistryList struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("could not configure CA certificates for registry %s: %w", config.Name, err)
+	}
+	endpoint.CAFile = config.CAFile
+	endpoint.rootCAs = rootCAs
+	return endpoint, nil
+}
+
+func loadRootCAs(caFile string) (*x509.CertPool, error) {
+	rootCAs, err := x509.SystemCertPool()
```
chengfang: when no custom ca is configured, I think we should just return rootCAs as nil to rely on the underlying go lib defaults.

w3lld1:
Thanks for the review — addressed the TLS follow-ups in `4111232f`:

- registries now auto-discover `/app/config/tls/<hostname>` when no `ca_file` is configured;
- added inline `ca_data` support;
- insecure registries skip CA loading and discovery;
- the default path leaves `RootCAs` nil so Go uses its normal trust defaults;
- transport cache keys now include endpoint identity and TLS settings, with regression coverage for shared API URLs using different CA configuration.

Fresh validation on the pushed commit: `go test -count=1 -p=2 ./pkg/registry`, `gofmt`, and `git diff --check`.

registry-scanner/pkg/registry/config.go
```diff
@@ -32,6 +37,66 @@ type RegistryList struct {
...
+
+func loadRootCAs(caFile, caData string) (*x509.CertPool, error) {
+	if caFile == "" && caData == "" {
+		return nil, nil
+	}
+
+	rootCAs, err := x509.SystemCertPool()
+	if err != nil {
+		return nil, fmt.Errorf("could not load system certificate pool: %w", err)
+	}
```
chengfang, line 87: when failing to load system cert pool, how about just log the warning and continue, without returning error, since there is still custom certs configured? Something like:
      log.Warnf("Could not load system cert pool, using empty pool: %v", err)
      rootCAs = x509.NewCertPool()

w3lld1: Updated in `4bc3642`: if the system certificate pool cannot be loaded, I now log a warning, initialize an empty pool, and continue loading the configured custom CA. I also added focused coverage for that fallback. After correcting the initial staticcheck finding, the focused registry tests, package golangci-lint, and `git diff --check` pass; the commit now includes the required DCO sign-off.

chengfang (approved):

PR #1757, opened 2026-08-02
fix(digest): pick the right live image when multiple containers share the same image

### LINKED ISSUE #1547: Cross-Alias Tag Confusion When Two Containers Share the Same Image Name
**Component:** argocd-image-updater
**Version:** v1.1.1
**Configuration method:** CR-based (`ImageUpdater` custom resource)
**Update strategy:** `digest`
**Write-back method:** git

## Summary

When a single ArgoCD application has two containers that use the **same image name (registry + repository) but different tags**, and both containers are tracked as separate aliases in an `ImageUpdater` CR, the updater incorrectly reads the current tag/digest of alias `A` from the container belonging to alias `B`. This causes:

1. The wrong tag name to be recorded as the "current" tag for the misidentified alias.
2. A wrong tag prefix (e.g. `latest`) to be written into the git write-back override file for the other container.
3. An **infinite commit loop** — the updater commits a "change" every reconcile cycle even though the desired state is already in place, because it perpetually reads a stale/wrong digest for one of the aliases.

## Environment

- ArgoCD image-updater: v1.1.1
- ArgoCD: (standard deployment, `argocd` namespace)
- Write-back method: git
- Helm-based application
- Two containers in one `Deployment`, both pulling from the same registry+repository path, under different named tags (`tag-a` and `tag-b`)

### 1. ArgoCD application values (helm)

Two containers in the same deployment, same image repository, different tags:

```yaml
# values.yaml (simplified)
containers:
  app:
    image:
      name: registry.example.com/myrepo/myimage
      tag: prod
  app-nginx:
    image:
      name: registry.example.com/myrepo/myimage
      tag: nginx-variant
```

### 2. ImageUpdater CR

```yaml
kind: ImageUpdater
metadata:
  name: argocd-image-updater
spec:
  writeBackConfig:
    method: git
  applicationRefs:
    - namePattern: "my-app-prod"
      images:
        - alias: app
          imageName: registry.example.com/myrepo/myimage:prod
          commonUpdateSettings:
            updateStrategy: digest
            allowTags: "regexp:^prod$"
          manifestTargets:
            helm:
              name: containers.app.image.name
              tag: containers.app.image.tag
        - alias: app-nginx
          imageName: registry.example.com/myrepo/myimage:nginx-variant
          commonUpdateSettings:
            updateStrategy: digest
            allowTags: "regexp:^nginx-variant$"
          manifestTargets:
            helm:
              name: containers.app-nginx.image.name
              tag: containers.app-nginx.image.tag
```

### 3. Trigger

Add the `app-nginx` alias to the CR for an application where `app` is already being tracked. On the next reconcile, the bug manifests immediately.

### Phase 1 — First reconcile: wrong `image_tag` read for `app` alias

The `app` alias (which tracks tag `prod`) logs `image_tag=nginx-variant`, indicating it read the current tag from the **`app-nginx` container** rather than its own container. Despite this, both `manifestTargets` write correctly in the first commit.

**Relevant logs (first reconcile, warmup-cache pass):**

```
time="...T05:44:06Z" level=info
  msg="Setting new image to registry.example.com/myrepo/myimage:prod@sha256:<DIGEST-PROD>"
  image_alias=app
  image_name=registry.example.com/myrepo/myimage
  image_tag=nginx-variant        ← WRONG: should be "prod"
  logger=warmup-cache

time="...T05:44:06Z" level=info
  msg="Successfully updated image 'registry.example.com/myrepo/myimage:nginx-variant' to
       'registry.example.com/myrepo/myimage:prod@sha256:<DIGEST-PROD>', but pending spec update"
  image_alias=app
  image_tag=nginx-variant        ← WRONG source tag reported
  logger=warmup-cache
```

The commit message generated for this first commit is also wrong — it reports both aliases as updating **from** `nginx-variant`:

```
build: automatic update of my-app-prod

updates image myrepo/myimage tag 'nginx-variant' to 'sha256:<DIGEST-PROD>'
updates image myrepo/myimage tag 'nginx-variant' to 'sha256:<DIGEST-NGINX>'
```

Both lines reference `nginx-variant` as the source tag. The first line should have referenced `prod` as the source tag for the `app` alias.

### Phase 2 — Subsequent reconciles: infinite commit loop (2 parameter updates per cycle)

Every ~2 minutes, image-updater commits 2 parameter updates for the application, even though the desired state is already written correctly. The `app` alias continues reading `image_tag=nginx-variant`:

```
time="...T05:46:15Z" level=info
  msg="Setting new image to registry.example.com/myrepo/myimage:prod@sha256:<DIGEST-PROD>"
  image_alias=app
  image_tag=nginx-variant        ← still wrong every cycle
  logger=reconcile

time="...T05:46:16Z" level=info
  msg="Setting new image to registry.example.com/myrepo/myimage:nginx-variant@sha256:<DIGEST-NGINX>"
  image_alias=app-nginx
  image_tag=nginx-variant        ← correct
  logger=reconcile

msg="Committing 2 parameter update(s) for application my-app-prod"
```

This repeats at 05:44, 05:46, 05:48 — every reconcile cycle.

### Phase 3 — After ArgoCD sync: tag prefix corruption and 1-parameter loop

After ArgoCD syncs the application and the containers start running with their pinned digests, the running image in the ArgoCD application spec is stored as a bare digest (no tag name prefix). At this point the `app` alias shifts to reading by digest:

```
time="...T05:50:27Z" level=info
  msg="Setting new image to registry.example.com/myrepo/myimage:prod@sha256:<DIGEST-PROD>"
  image_alias=app
  image_digest="sha256:<DIGEST-NGINX>"    ← reading nginx-variant's digest as app's current digest
  logger=reconcile

time="...T05:50:27Z" level=info
  msg="Successfully updated image 'registry.example.com/myrepo/myimage@sha256:<DIGEST-NGINX>'
       to 'registry.example.com/myrepo/myimage:prod@sha256:<DIGEST-PROD>'"
  image_alias=app
  image_digest="sha256:<DIGEST-NGINX>"
```

The updater now commits **1 parameter update** per cycle (down from 2). The single parameter changed is `containers.app-nginx.image.tag` — it is corrupted from:

```
nginx-variant@sha256:<DIGEST-NGINX>
```

to:

```
latest@sha256:<DIGEST-NGINX>
```

The tag prefix `nginx-variant` is replaced with `latest`. The digest itself does not change. This is data corruption: `latest` may point to a completely different image layer in the registry, causing the wrong binary to be deployed if the override is ever re-applied without a running container pinning the digest.

The infinite loop then continues: every reconcile, `containers.app-nginx.image.tag` is rewritten from `nginx-variant@sha256:<DIGEST-NGINX>` to `latest@sha256:<DIGEST-NGINX>`, generating 21+ spurious git commits over the observed log window.

## Git Write-Back: Before and After

**Override file before corruption** (correct state):

```yaml
helm:
  parameters:
  - name: containers.app-nginx.image.name
    value: registry.example.com/myrepo/myimage
    forcestring: true
  - name: containers.app-nginx.image.tag
    value: nginx-variant@sha256:<DIGEST-NGINX>    # correct
    forcestring: true
  - name: containers.app.image.name
    value: registry.example.com/myrepo/myimage
    forcestring: true
  - name: containers.app.image.tag
    value: prod@sha256:<DIGEST-PROD>              # correct
    forcestring: true
```

**Override file after corruption** (commit introduced by image-updater):

```yaml
  - name: containers.app-nginx.image.tag
    value: latest@sha256:<DIGEST-NGINX>           # WRONG prefix: 'nginx-variant' → 'latest'
```

## Root Cause (Hypothesis)

During the **read phase**, when image-updater resolves the current image for each alias from the ArgoCD application spec, it appears to match containers by **image name only** (registry + repository path), without considering the tag. When two containers share the same image name, both are matched by both aliases — the first (or last) matched container's tag is then attributed to whichever alias happens to process it, causing the cross-alias confusion.

The **write phase** (using `manifestTargets`) is respected and writes to the correct helm parameters, but because the read phase fed the wrong current digest into the alias's state, the updater perpetually believes an update is needed and also corrupts the tag prefix of the sibling alias's override parameter.

## Expected Behaviour

When two aliases in the same application reference the same image name but different tags, and each alias has explicit `manifestTargets` pointing to distinct helm parameters, the updater should:

1. Read the current image state for each alias from its **own** `manifestTargets` helm parameter (rather than scanning all containers by image name).
2. Correctly attribute `image_tag=prod` to the `app` alias and `image_tag=nginx-variant` to the `app-nginx` alias.
3. Stop committing after the write-back override already reflects the desired state.
4. Never write a tag prefix (`latest`) that was not configured in `allowTags` for that alias.

## Workaround

Remove one of the conflicting aliases from the `ImageUpdater` CR so that only one alias tracks the shared image name. The removed alias must be managed with manual digest pins in the git override file.

## Additional Notes

- The `allowTags` filter is set correctly for both aliases (`regexp:^prod$` and `regexp:^nginx-variant$`). The bug occurs before `allowTags` filtering — in the current-state read phase.
- The bug is reproducible deterministically: it triggers immediately on the first reconcile after adding a second alias whose image name matches an existing alias.
- The problem does not occur when the two containers use images from **different** repositories.

C4tWithShell:
Fixes #1547 

## The problem
When several containers in an Application reference the same image name (registry + repository) with `updateStrategy: digest`, a container can be evaluated against a sibling's live image instead of its own - either missing an update or making a spurious one.

**Root cause**. The current live image for an alias is resolved via `ContainerImageList.ContainsImage`, which matches on `image name + registry only` and returns the first match. The live list comes from Argo CD's `Status.Summary.Images`, which is alias-less and de-duplicated by full reference. So when two containers share an image name, the lookup for one alias can return the other's live image. Two observed cases:

- **Same tag (e.g. two deployments both on `:develop`)**: once a sibling is bumped, the tracked container matches the sibling's already-updated digest, is considered up to date, and is never updated (`images_updated=0`, permanently stuck).
- **Different tags (e.g. `tag-a / tag-b`, #1547)**: the tracked alias reads the wrong container's digest, leading to wrong writes / churn.

## Fix
Add `ContainerImageList.MatchForDigestUpdate(img, latestDigest)`, used for digest strategy, which re-selects the live image that actually belongs to the alias:

1. Prefer a live image whose tag name matches the alias's  tag - so aliases that share an image name but track different tags are matched to the right container.
2. Among those, prefer one whose digest differs from the latest - so when two containers share both name and tag, the one still on the old digest is the one that gets updated.
3. Fall back to a stale name+registry match when the summary has no matching tag (previous behavior).

Tests
`registry-scanner/pkg/image: Test_MatchForDigestUpdate ` - same-tag stale selection, different-tag selection (both aliases), fallback, and nil cases.
`pkg/argocd: Test_UpdateApplication regression` -  a tracked container is updated even when a sibling shares the image name and tag; fails without the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved digest-based image updates when multiple containers share an image name.
  * Ensured updates select the correct image across differing tags and digests.
  * Added fallback handling for digest-only image references.
  * Preserved configured image tags when the live image is already up to date.

* **Tests**
  * Added coverage for shared image names, tag variations, stale digests, digest-only references, and unmatched images.

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -199,6 +199,20 @@ func UpdateApplication(ctx context.Context, updateConf *UpdateConfiguration, sta
 			continue
 		}
 
+		// When several containers in the application reference the same image
+		// name, the live image list (Argo CD's alias-less, de-duplicated
+		// Status.Summary.Images) can cause the earlier ContainsImage lookup to
+		// resolve to a sibling container, hiding this image's pending update or
+		// triggering a spurious one. For digest strategy, re-select the live
+		// image that actually belongs to this alias (by tag, then by stale
+		// digest) so it is evaluated against its own container.
+		if vc.Strategy == image.StrategyDigest {
+			if m := applicationImages.MatchForDigestUpdate(applicationImage.ContainerImage, latest.TagDigest); m != nil && m != updateableImage {
+				imgCtx.Debugf("Multiple containers reference image '%s'; re-selected live image (tag '%s', digest '%s') for update evaluation", applicationImage.ImageName, m.ImageTag.TagName, m.ImageTag.TagDigest)
+				updateableImage = m
+			}
+		}
+
 		if needsUpdate(updateableImage, applicationImage.ContainerImage, latest, vc.Strategy) {
 			appImageWithTag := applicationImage.WithTag(latest)
 			appImageFullNameWithTag := appImageWithTag.GetFullNameWithTag()
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -266,7 +266,22 @@ func UpdateApplication(ctx context.Context, updateConf *UpdateConfiguration, sta
 			// We need to explicitly set the up-to-date images in the spec too, so
 			// that we correctly marshal out the parameter overrides to include all
 			// images, regardless of those were updated or not.
-			err = setAppImage(imageOpCtx, &updateConf.UpdateApp.Application, applicationImage.WithTag(updateableImage.ImageTag), updateConf.UpdateApp.WriteBackConfig, applicationImage)
+			currentTag := updateableImage.ImageTag
+			// After a sync, Argo CD can record the running image in
+			// Status.Summary.Images as a bare digest with no tag name. For digest
+			// strategy, re-writing the image from that tag-less value drops the
+			// tracked tag name and degrades it to "latest@<digest>", corrupting the
+			// write-back and producing a commit every reconcile cycle. Preserve the
+			// alias's tracked tag name in that case.
+			if vc.Strategy == image.StrategyDigest && vc.Constraint != "" &&
+				(currentTag == nil || currentTag.TagName == "") {
+				digest := ""
+				if currentTag != nil {
+					digest = currentTag.TagDigest
+				}
+				currentTag = tag.NewImageTag(vc.Constraint, time.Unix(0, 0), digest)
+			}
+			err = setAppImage(imageOpCtx, &updateConf.UpdateApp.Application, applicationImage.WithTag(currentTag), updateConf.UpdateApp.WriteBackConfig, applicationImage)
 			if err != nil {
 				imgCtx.Errorf("Error while trying to update image: %v", err)
 				result.NumErrors += 1
--- a/pkg/argocd/update_test.go
+++ b/pkg/argocd/update_test.go
@@ -2230,6 +2230,106 @@ registries:
 		assert.Equal(t, 0, res.NumErrors)
 		assert.Equal(t, 1, res.NumImagesUpdated)
 	})
+
+	// Regression test for #1547: when two containers share the same image name
+	// and tag but their digests have diverged, the live image list
+	// (Status.Summary.Images) is alias-less and de-duplicated by full reference.
+	// A plain ContainsImage lookup returns the first name match — a sibling
+	// container that was already updated — hiding the tracked container's pending
+	// update. Without the fix this reports images_updated=0; with it, the tracked
+	// container is updated to the new digest.
+	t.Run("Test digest update when a sibling container shares the image name and tag (#1547)", func(t *testing.T) {
+		newDigest := "sha256:1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef"
+		newDigestBytes := [32]byte{
+			0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef,
+			0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef,
+			0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef,
+			0x12, 0x34, 0x56, 0x78, 0x90, 0xab, 0xcd, 0xef,
+		}
+		// A different, older digest that our tracked container is still pinned to.
+		oldDigest := "sha256:0000000000000000000000000000000000000000000000000000000000000000"
+
+		mockClientFn := func(endpoint *registry.RegistryEndpoint, username, password string) (registry.RegistryClient, error) {
+			regMock := regmock.RegistryClient{}
+			regMock.On("NewRepository", mock.Anything, mock.Anything).Return(nil)
+			regMock.On("Tags", mock.Anything).Return([]string{"latest"}, nil)
+			meta := &schema2.DeserializedManifest{}
+			regMock.On("ManifestForTag", mock.Anything, "latest").Return(meta, nil)
+			// The mutable "latest" tag now resolves to the NEW digest.
+			regMock.On("TagMetadata", mock.Anything, mock.Anything, mock.Anything).Return(&tag.TagInfo{
+				CreatedAt: time.Now(),
+				Digest:    newDigestBytes,
+			}, nil)
+			return &regMock, nil
+		}
+
+		argoClient := argomock.ArgoCD{}
+		argoClient.On("UpdateSpec", mock.Anything, mock.Anything).Return(nil, nil)
+
+		kubeClient := kube.ImageUpdaterKubernetesClient{
+			KubeClient: &registryKube.KubernetesClient{
+				Clientset: fake.NewFakeKubeClient(),
+			},
+		}
+
+		// We track one alias ("web"), still pinned to the old digest.
+		containerImg := image.NewFromIdentifier("web=docker.io/library/nginx:latest")
+		iuImg := NewImage(containerImg)
+		iuImg.UpdateStrategy = image.StrategyDigest
+
+		appImages := &ApplicationImages{
+			Application: v1alpha1.Application{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      "guestbook",
+					Namespace: "guestbook",
+				},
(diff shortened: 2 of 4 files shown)


chengfang:
looking at the related issue https://github.com/argoproj-labs/argocd-image-updater/issues/1547 , this pr fixes the phase 1 and 2 problems, but not phase 3. During phase 3, after argcd sync, image updater will not be able to distinguish between multiple same-named, bare-tag images, and the data corruption will continue in every cycle. So I think it will be better to aim for a complete fix to all symptoms in #1547

C4tWithShell:
> looking at the related issue #1547 , this pr fixes the phase 1 and 2 problems, but not phase 3. During phase 3, after argcd sync, image updater will not be able to distinguish between multiple same-named, bare-tag images, and the data corruption will continue in every cycle. So I think it will be better to aim for a complete fix to all symptoms in #1547

That's right - this PR intentionally covers Phases 1 and 2, which are the symptoms I actually hit: two containers in one app built from the same image and tag (…/app:develop) under the digest strategy, where one container gets stuck reading the other's already-updated digest. The tag-aware selection in `MatchForDigestUpdate` fixes that case. I referenced #1547 rather than opening a new issue to avoid a duplicate, since the root cause is the same.

Phase 3 is a real problem,  and I believe it is better handled in a separate PR

Its trigger is different - after sync, the live image list (Status.Summary.Images) records bare digests with no tag, so there's nothing left to match on. The tag-aware selection here has no signal to work with once that happens.
It also involves a write-side bug (the tag prefix being lost and rewritten to latest@…), which is outside the read/selection path this PR touches

Possible fix for Phase 3, I think, is to stop deriving each alias's current image from the alias-less, tag-strippable summary altogether, and instead read it from the alias's own manifest target (the written-back Helm param / kustomize entry), which keeps the tag - it would also remove the latest corruption. But that is a deeper change to the write-back read path and deserves its own PR

chengfang:
>Possible fix for Phase 3, I think, is to stop deriving each alias's current image from the alias-less, tag-strippable >summary altogether, and instead read it from the alias's own manifest target (the written-back Helm param / kustomize >entry), which keeps the tag - it would also remove the latest corruption. But that is a deeper change to the write-back >read path and deserves its own PR

The above fix for phase 3 may likely also fix all problems 3 phases. So if in teh future we implement it, we will probalby undo this pr.

C4tWithShell:
> > Possible fix for Phase 3, I think, is to stop deriving each alias's current image from the alias-less, tag-strippable >summary altogether, and instead read it from the alias's own manifest target (the written-back Helm param / kustomize >entry), which keeps the tag - it would also remove the latest corruption. But that is a deeper change to the write-back >read path and deserves its own PR
> 
> The above fix for phase 3 may likely also fix all problems 3 phases. So if in teh future we implement it, we will probalby undo this pr.

I think I found simpler solution - I can recover the tag from config we already have, not by reading the write-back target.

It should preserve the alias's tracked tag name (the digest-strategy constraint) when the running image carries no tag, so the write-back keeps `<tag>@<digest>` and stops churning. I also added an UpdateApplication regression test with a bare-digest summary

This removes the corruption and the commit loop. BUT in a fully tag-less summary the read still can't distinguish two bare-digest siblings, but that case now routes through the normal update branch, which writes the correct `<tag>@<latest-digest>` - so there's no data corruption and no loop

chengfang (approved):

PR #1767, opened 2026-08-10
fix(argocd): retain yaml document start if exists

stphnlwlsh:
If a yaml write-back target has an explicit document-start, the current update.go disregards it. This fix retains the explicit document-start but does not add one if it doesn't.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved existing YAML document-start markers (`---`) when updating Kustomize, Helm, plugin, and Helm values files.
  * Prevented markers from being added when they were not originally present.
  * Preserved leading blank lines, comments, and directives around markers.
  * Improved image digest updates for duplicate references and retained configured tag names when only digest data is available.

* **Tests**
  * Added regression coverage for YAML formatting preservation and image digest update scenarios.

### DIFF
--- a/pkg/argocd/update.go
+++ b/pkg/argocd/update.go
@@ -414,7 +414,11 @@ func setAppImage(ctx context.Context, app *v1alpha1.Application, img *image.Cont
 	}
 }
 
-func marshalWithIndent(in any, indent int) (out []byte, err error) {
+// marshalWithIndent marshals in to YAML with the given indent, optionally
+// prepending an explicit "---" document-start marker. See
+// hasExplicitDocumentStart for why callers track and pass that bit through
+// explicitly rather than relying on the encoder to preserve it.
+func marshalWithIndent(in interface{}, indent int, explicitStart bool) (out []byte, err error) {
 	var b bytes.Buffer
 	encoder := yaml.NewEncoder(&b)
 	defer encoder.Close()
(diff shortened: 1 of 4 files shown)


pkg/argocd/update.go
```diff
@@ -427,7 +431,31 @@ func marshalWithIndent(in any, indent int) (out []byte, err error) {
...
+// document-start marker ("---"). Directives are only ever valid directly
+// before a document-start marker, so skipping them is required for
+// correctness, not just leniency. Write-back re-marshals the target file
+// through an Encoder that has no memory of the source's original formatting,
+// so callers use this to carry that one bit of information through the
+// round-trip explicitly, rather than always adding or always dropping the
+// marker regardless of what the file looked like before.
+func hasExplicitDocumentStart(raw []byte) bool {
+	for _, line := range strings.Split(string(raw), "\n") {
+		trimmed := strings.TrimSpace(line)
```
chengfang: Both `hasExplicitDocumentStart` (here) and `hasDocumentStartAt` (git.go:660) use `TrimSpace` before checking for `"---"`, which means an indented line like `   ---` is falsely detected as a document-start marker. Per the YAML spec, the document-start indicator must begin at column 0 — indented `---` is content, not a marker.

Low probability for `.argocd-source-*` and kustomization files in practice, but the fix is small — check the untrimmed line instead:

```go
// hasExplicitDocumentStart (update.go)
return trimmed == "---" || ...
// →
return line == "---" || strings.HasPrefix(line, "--- ") || strings.HasPrefix(line, "---\t")
```

```go
// hasDocumentStartAt (git.go)
trimmed := bytes.TrimSpace(line)
// →  check `line` directly (after trimming only trailing whitespace / \r\n)
line = bytes.TrimRight(line, " \t\r\n")
```

pkg/argocd/update_test.go
```diff
@@ -7272,6 +7272,11 @@ plugin:
 	})
 }
 
+// Helper function to create string pointers for testing
+func stringPtr(s string) *string {
+	return &s
+}
+
```
chengfang: Also pls run `go fix ./...` from the project root to replace stringPtr with built-in new, and interface{} with any, etc. Looks like your changes overwrote some lines from previously merged PR during rebase.

Then, this unused func should be removed after the first step.

chengfang (approved):

PR #1782, opened 2026-08-14
fix(crd): remove commonUpdateSettings defaults at the CRD level

MrFreezeex:
Defaults for these fields are already handled within `newImageFromCommonUpdateSettings` and having them set by default by Kubernetes prevents `mergeCommonUpdateSettings` to properly handle inheritance from `spec.commonUpdateSettings` to
`spec.applicationRefs.*.images.*.commonUpdateSettings`.

Not setting any default is something already correctly tested but in practice the following `ImageUpdater` resource would have incorrectly defaulted `updateStrategy` to `semver` and `forceUpdate` to `false`:

```yaml
kind: ImageUpdater
metadata:
  name: broken-default
  namespace: argocd
spec:
  commonUpdateSettings:
    updateStrategy: digest
    forceUpdate: true
  applicationRefs:
    - namePattern: my-app
      images:
        - alias: app
          imageName: something
          commonUpdateSettings: {}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed automatic default values for update strategy and force-update settings.
  * These options now remain unset unless explicitly configured, while existing field types and validation remain unchanged.
  * This prevents unintended settings from being applied across global, application, and image configurations.

* **Documentation**
  * Added an upgrade warning about potential inheritance changes in existing resources.
  * Documented steps for removing unintended persisted values or recreating affected resources.

### DIFF
--- a/api/v1alpha1/imageupdater_types.go
+++ b/api/v1alpha1/imageupdater_types.go
@@ -177,13 +177,11 @@ type CommonUpdateSettings struct {
 	// Examples: "semver", "latest", "digest", "name".
 	// This acts as the default if not overridden at a more specific level.
 	// +optional
-	// +kubebuilder:default:="semver"
 	UpdateStrategy *string `json:"updateStrategy,omitempty"`
 
 	// ForceUpdate specifies whether updates should be forced.
 	// This acts as the default if not overridden.
 	// +optional
-	// +kubebuilder:default:=false
 	ForceUpdate *bool `json:"forceUpdate,omitempty"`
 
 	// AllowTags is a regex pattern for tags to allow.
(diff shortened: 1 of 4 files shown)


chengfang:
this change will affect existing ImageUpdater CRs that use or rely on the CRD-level defaults. These apps will see different effective updateStrategy values when deployed to existing versions of ImageUpdater vs next version that include this PR.

Alsor for apps upgrading to the next version of ImageUpdater, their existing ImageUpdater CRs already persisted with prevoius defaults, ans will see different inheritance behaviors. 

We shoudl document these behavior changes with mitigation help.

MrFreezeex:
Thanks @chengfang, I added some documentation for that let me know your though on that (including if this is in the correct place :see_no_evil:).

chengfang (approved):

PR #1801, opened 2026-09-10
fix(registry): controller sends unauthenticated registry requests — registries.conf credentials AND per-image pullSecret never reach the wire

### LINKED ISSUE #1797: v1.3.0: controller sends unauthenticated registry requests — registries.conf credentials AND per-image pullSecret never reach the wire
## Description

Since upgrading to the v1.3.0 controller (CRD ImageUpdater architecture), the controller sends **anonymous** requests to a private registry that requires Basic auth — no Authorization header is ever attached, regardless of how credentials are configured. The registry returns 401 "authentication required" on every cycle for every application, so no update PRs are ever opened.

This looks related to #1046 (reported against 0.15.2 for registries.conf credentials) — but it extends to the per-image `commonUpdateSettings.pullSecret` path in v1.3.0, which the docs describe as overriding registry-level credentials.

## Environment
- argocd-image-updater v1.3.0 (chart 1.3.0, controller kind ImageUpdater)
- Registry: zot (Basic auth, anonymous denied on manifest/tag reads)
- Secret: opaque, key `creds` = `<user>:<password>` (58 chars, no trailing newline), same namespace (argocd) as the ImageUpdater CRs

## Config used

registries.conf (confirmed inside the pod at /app/config/registries.conf):
```yaml
registries:
  - api_url: http://10.10.8.50:5000
    credentials: secret:argocd/zot-registry-creds#creds
    name: zot-shrestech
    ping: true
    prefix: registry.shrestech.com
```

Per-image (added later, same result):
```yaml
images:
  - alias: web
    imageName: registry.shrestech.com/shrestech/printready-web:stg-latest
    commonUpdateSettings:
      updateStrategy: digest
      pullSecret: secret:argocd/zot-registry-creds#creds
```

## Evidence the credentials themselves are fine
From a pod in the same namespace with the same secret mounted:
```
GET /v2/shrestech/printready-web/manifests/stg-latest  (with creds) -> 200 in 0.011s
GET /v2/shrestech/printready-web/manifests/stg-latest  (anonymous)  -> 401 UNAUTHORIZED "authentication required"
```

## Wire-level proof the controller sends nothing
tcpdump on the controller's node (host <registry> and port 5000), plain HTTP, 150s capture while cycles ran:
```
15 x GET /v2/
15 x GET /v2/shrestech/printready-*/tags/list
Authorization request headers found: NONE
```
The only Authorization string in the capture is the registry's CORS response header (Access-Control-Allow-Headers). The controller logs no credential-resolution warnings (no "Could not fetch credentials", no cross-namespace rejection) — the 401 surfaces only as `Could not get tags from registry: unauthorized: authentication required`.

## What we checked
- RBAC: controller Role has get,list,watch on secrets/configmaps in the CR namespace ✓
- Cross-namespace guard: secret ns == CR ns == app ns (argocd) ✓
- Secret format: user:password, single colon, no trailing newline ✓
- Controller restart (clears any credential cache) — still anonymous ✓
- ping: true pings succeed (200) because zot allows anonymous /v2/; the tag/manifest reads fail

## Workaround
Replaced the AIU pin role with a scheduled GitHub Action that resolves the mutable tags and opens the pin PRs itself. Config left in git for when this is fixed.

chengfang:
Fixes #1797 

## Summary by CodeRabbit

* **Bug Fixes**
  * Registry access now handles repositories that require authentication even when the registry root allows anonymous access.
  * Supports Basic and Bearer authentication challenges, including requests with replayable bodies and mixed authentication responses.
  * Prevents credentials from being sent to unrelated redirect targets or public repositories.
  * Preserves appropriate authentication errors when credentials cannot satisfy a challenge.
  * Warns when Basic authentication is used over unencrypted HTTP.

### DIFF
--- a/registry-scanner/pkg/registry/client.go
+++ b/registry-scanner/pkg/registry/client.go
@@ -6,6 +6,7 @@ import (
 	stdjson "encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"strings"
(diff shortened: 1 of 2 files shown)


registry-scanner/pkg/registry/client.go
```diff
@@ -123,6 +124,147 @@ func (rlt *rateLimitTransport) RoundTrip(r *http.Request) (*http.Response, error
...
+
+// canSatisfy reports whether any offered scheme can actually be answered with
+// what this client holds. Recording a Basic challenge with no credentials to
+// answer it would make endpointAuthorizer fail the replay before it reaches
+// the wire, replacing the registry's own "unauthorized" with an opaque
+// "no basic auth credentials". Bearer is always worth attempting: token
+// services routinely issue anonymous pull tokens.
+func (crt *challengeRetryTransport) canSatisfy(challenges []challenge.Challenge) bool {
+	for _, c := range challenges {
+		if c.Scheme != "basic" {
```
chansuke: canSatisfy and AddResponse disagree here: with a Bearer+Basic 401 and no creds, canSatisfy passes (Bearer works anonymously) but AddResponse records the Basic challenge too, and the replay dies with [ErrNoBasicAuthCredentials](https://github.com/argoproj-labs/argocd-image-updater/pull/1801/...). Filter the challenges before recording?

chengfang: @chansuke I've updated the pr. Can you review again? Thanks.

chengfang: @chansuke can you please review again?

chansuke (approved):
lgtm

